PR #58, opened 2026-04-27
windows: functional port — transport, install, CI, tests, polish

h5kk:
## Summary

End-to-end Windows functionality + the bugs that surfaced while getting there. Tested on Windows 11 (Windows Terminal + Rust 1.93). Linux/macOS paths untouched (verified with `cargo clippy --all-targets -- -D warnings` on both `--no-default-features --features pdf` and default features).

Twelve focused commits, ~1200 LOC net add. Squash-merge is fine; the per-commit messages are written so the history reads cleanly either way.

### Windows port — transport / build / CI

- **`Listener::accept` is now uniform across platforms** (`24414662`). Held the Windows `NamedPipeServer` behind `tokio::sync::Mutex` so `accept(&self)` matches `tokio::net::UnixListener::accept` on Unix. Stops the regression cycle in `src/server/runtime.rs` where `mut listener: Listener` kept getting re-added (e.g. `e5d91ab2`, `d398bafa`) and stripped again by Linux clippy. Also drops the `#[cfg(windows)] let mut listener` shadowing in `src/tui/app/state_ui.rs` and the `let mut server` in `pair()`.
- **cfg-mismatch warnings cleaned up so `cargo clippy --all-targets -- -D warnings` is green on Windows** (`40b8bac5`). 30 files, mostly cfg-gating imports/items behind the cfgs that actually use them (`BufReader`/`Write`/`Duration`/`Instant` in `build.rs`, `OpenOptions`/`StdCommand` in `tool/bash.rs`, `OsString` in `tool/open.rs`, `Listener` in `registry.rs`/`socket_tests.rs`, `Command/Stdio` in `stdin_detect.rs`, etc.). Also: collapsed nested `if`s in `cli/dispatch.rs`/`cli/tui_launch.rs`, dropped needless `return`s, switched `eprint!("\n")` → `eprintln!()`, switched `state.launch_count % 3 != 0` → `.is_multiple_of(3)`, panic-converted `assert!(false, ...)` in `server/reload.rs`, swapped `#[expect(clippy::result_large_err)]` → `#[allow]` in `gateway.rs`. **Real bug fix tucked in here**: `background.rs::cancel_with_grace` now actually observes the grace period on Windows instead of dropping the timeout (was both an `unused_variable` warning *and* a silent skip of the SIGTERM-equivalent grace period).
- **CI gate**: `windows-build-test` job in `ci.yml` now installs the clippy component and runs `cargo clippy --locked --target x86_64-pc-windows-msvc --all-targets --no-default-features --features pdf -- -D warnings` before the test compile (`ae587530`). This catches the kind of regression the existing Linux `quality` job missed (clippy on Linux doesn't see Windows code paths). `docs/WINDOWS.md` rewritten to reflect actual state and re-scoped Remaining Work.

### Self-update + first-run install path

- **`scripts/install_release.ps1`** (`016e5f72`). Windows companion to `install_release.sh`: builds, copies into `%LOCALAPPDATA%\jcode\builds\versions\<hash>\jcode.exe`, refreshes stable + current channels, points the launcher at current. Uses copy (Windows symlinks need Developer Mode), reuses the rename-aside trick from the new `replace_executable_atomic`.
- **`replace_executable_atomic`** in `src/platform.rs`. Windows can't unlink a running `.exe`, so we rename the running image to a hidden `.<name>.<pid>-<nanos>.old` sibling (Windows allows rename even while the file is loaded) before copying the new image into place. `src/build.rs::install_binary_at_version` and `src/platform.rs::atomic_symlink_swap` route through it. Two unit tests cover the idle case and the locked-running-image case (mimicked with `FILE_SHARE_READ | FILE_SHARE_DELETE` on a held handle — the share mask Windows uses for loaded images).
- **Startup sweeps** the `.old` sidecars in `stable`/`current`/`canary`/`shared-server` channels and every `versions/<v>/` on every launch (`src/cli/startup.rs`).

### Auto-spawn + `replace_process`

- **Auto-spawn timeout & detached handoff** (`106cf104`). Two bugs in the `cfg(not(unix))` branch of `start_server_inner`:
  1. The 500 ms wait was way too short. A cold-start release server with default features (embeddings + pdf) takes 2–3 s to bind the named pipe; the wait gave up and the client fell into a reconnect loop that never recovered ("starting server…" forever in the TUI).
  2. `cmd.stderr(Stdio::piped())` plus immediately dropping the `Child` closed the pipe's read end before the server had written anything — the server's first stderr write hit `ERROR_BROKEN_PIPE`.

  Fix: 15 s wait, 100 ms polls, stderr → NUL, spawn via `platform::spawn_detached` (`CREATE_NEW_PROCESS_GROUP | DETACHED_PROCESS`) so the daemon survives the client losing its terminal.
- **`replace_process` Windows path** also passes `CREATE_NEW_PROCESS_GROUP` so Ctrl+C in the parent terminal doesn't kill the new child during the brief overlap window, and logs spawn failures (was previously silent).

### Test stabilization on Windows

`c5a16915` and `1e245200`:

- **`vscdb` cursor tests soft-skip when `sqlite3` is not on PATH** (the default for Windows GitHub runners). Linux/macOS preinstalls sqlite3, so they keep running there.
- **`provider/openrouter_tests` redirects autodetection probes to `$JCODE_HOME/config/jcode/`**. The previous `APPDATA` redirect didn't work on Windows because `dirs::config_dir()` uses `SHGetKnownFolderPath` and ignores the `APPDATA` env var; only `JCODE_HOME` is honored by `storage::app_config_dir()`. Without this, the first poisoned `ENV_LOCK` cascaded into 6 `PoisonError` failures.
- **`tool::agentgrep_tests` and `tool::communicate_tests::end_to_end::communicate_spawn_…`** — Unix-shaped tests (path-separator literals, agentgrep crate internals, comm-channel handshake) gated to `#[cfg(unix)]` until the cross-platform paths are sorted.
- **`tool::todo` and `tool::mcp` test refresh** — stale assertions (todos schema also exposes the universal `intent` property; mcp description is the short form `"Manage MCP servers."`).
- **JSON path-separator escape in `src/tui/session_picker/loading_tests.rs`** — fixtures were `format!()`-ing `Path::display()` into JSON string literals; on Windows that produces unescaped backslashes (`"fullPath":"C:\Users\..."`) which makes the JSON parser reject the file. Use `serde_json::to_string` for proper escaping.

### Provider / UX (general, found while porting)

- **`AnthropicProvider` is created when `ANTHROPIC_API_KEY` is set**, not just when Claude OAuth credentials exist (`75c5ede0`). `MultiProvider::new_with_auth_status` previously gated the provider on `auth::claude::load_credentials().is_ok()`, so a fresh box with only `ANTHROPIC_API_KEY` exported ended up with `claude=false, anthropic=false` in availability. Result: the auto-detected default fell through to the next provider, and `[provider] default_provider = "claude"` logged "Preferred provider 'claude' is not configured, using auto-detected default" and was silently ignored. Meanwhile the status bar's `anthropic(key)` indicator (driven by `AuthStatus::check`) *did* notice the env var, making the discrepancy extra confusing. `AnthropicProvider::get_access_token` already preferred `ANTHROPIC_API_KEY` over OAuth — it just never got called because the provider wasn't instantiated.
- **Better Copilot 404 diagnostic** (`4e292aca`). When a GitHub OAuth token authenticates fine but the account has no active Copilot subscription, GitHub deliberately returns HTTP 404 (not 403) from `copilot_internal/v2/token` to hide the endpoint's existence. The previous error dumped the raw `{"message":"Not Found",...}` body. Now jcode emits a diagnostic message naming the actual cause (no Copilot subscription on this account) and the remediation paths (subscribe, switch accounts via `gh auth login`, set `COPILOT_GITHUB_TOKEN`, or pick another provider). Failover classifier still routes the new message correctly (test added).

### Optional new feature — `/help-cmds`

- `61fcc23d` adds a `/help-cmds` slash command that pushes a single-message cheat sheet covering features, configuration knobs, multi-window workflow, and the status-bar legend. Easy to drop if you'd rather not add a new slash command — it's the last commit on the branch.

### Optional new doc — `CLAUDE.md`

- `CLAUDE.md` at the repo root, parallel to the existing `AGENTS.md` but specifically the repo guide consumed by Claude Code agents (workflow commands + architecture + quality guardrails). Easy to drop if you don't want a separate file.

## Test plan

- [x] `cargo fmt --all -- --check` passes
- [x] `cargo clippy --all-targets --no-default-features --features pdf -- -D warnings` passes on Windows (new gate)
- [x] Targeted Windows tests pass: the named set in `windows-smoke.yml` plus `replace_executable_atomic_*` plus `copilot_exchange_*` plus `should_failover_*` plus `test_help_cmds_command_is_registered_and_runs` (18/18 locally)
- [x] e2e: `provider_behavior::test_socket_model_cycle_supported_models` and `provider_behavior::test_model_switch_resets_provider_session` pass on Windows
- [x] Local Windows TUI launch from a fresh terminal hits the chat in 2–3 s (auto-spawn fix)
- [x] `ANTHROPIC_API_KEY`-only setup boots with `default_provider = "claude"` and resolves to `Claude` (anthropic-key fix)

## Things I deliberately did *not* do

- Self-dev binary swap on Windows is partial. The rename-aside path works for the launcher in `install_release.ps1` and for `install_binary_at_version`, but the running-server `/reload` path may still have edge cases on Windows; not exercised in this PR.
- The full `cargo test --lib --bins` is not yet verified on Windows. Roughly 40 ratatui-snapshot tests in `tui::ui::*`, `tui::app::tests`, `tui::markdown`, `tui::mermaid` fail with what look like font-metric / snapshot-timing issues, not product behavior. `server::` long socket-handshake tests appear to deadlock when run together. Documented in `docs/WINDOWS.md`.
- `windows-smoke.yml` is still `workflow_dispatch`-only. The new `windows-build-test` clippy gate in `ci.yml` covers the regressions that historically slipped through; promoting `windows-smoke` to required-on-PR is a separate decision.

## Reviewer's split (if a single PR is too big)

The natural split is:

1. **Core Windows port** — `24414662 40b8bac5 016e5f72 c5a16915 ae587530 1e245200 dd23cafd 5f5463f0 106cf104` (transport, clippy, install, tests, CI, JSON path escape, doc, listener-shadow cleanup, auto-spawn fix).
2. **Provider/UX improvements found while porting** — `4e292aca 75c5ede0 61fcc23d` (Copilot 404 diagnostic, AnthropicProvider env-key fix, `/help-cmds`).

Both halves stand alone; #2 doesn't depend on #1.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
<a href="https://app.blacksmith.sh/1jehuang/codesmith/jcode/pr/58"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


h5kk:
Pushed a follow-up commit ([`c24727f6`](https://github.com/h5kk/jcode/commit/c24727f6)): **windows: suppress new console window when bash tool spawns cmd.exe**.

When the jcode server runs detached (`CREATE_NEW_PROCESS_GROUP | DETACHED_PROCESS` via `platform::spawn_detached`), it has no inherited console. `tool::bash::build_shell_command` invokes `cmd.exe /C <command>` for every Bash tool call; Windows then auto-allocates a fresh console window for the child cmd.exe (it's a console-subsystem program), which flashes up as a stray PowerShell/cmd window. Adding `CREATE_NO_WINDOW` (`0x08000000`) to creation flags suppresses the window — stdio is already piped, so no functional impact on tool I/O.

Same flicker pattern likely exists for other server-side spawns of console programs (`tool::ambient` git probe, `tool::selfdev` git/cargo invocations); not covered in this commit so the fix stays scoped to the reported repro. Open to extracting a shared `platform::suppress_child_console` helper if you'd prefer the broader sweep land here.

h5kk:
Follow-up commit ([`ad139a9b`](https://github.com/h5kk/jcode/commit/ad139a9b)): **windows: extract suppress_child_console helper, sweep server-side spawns**.

Generalizes the previous fix into a reusable `platform::suppress_child_console{,_async}` helper (no-op on non-Windows) and applies it to every server-side spawn of a console-subsystem program with already-piped stdio:
- `src/tool/bash.rs` (refactor of c24727f6 to call the helper)
- `src/server/client_actions.rs` — input shell command (`cmd.exe /C ...`)
- `src/tool/ambient.rs` — `git rev-parse` probe
- `src/tool/selfdev/status.rs` — `git status --porcelain`
- `src/tool/selfdev/build_queue.rs` — build streaming command (cargo etc.)
- `src/agent/utils.rs` — `git_output` used by image side-panel hooks

Same root cause throughout: detached server has no inherited console, so Windows allocates a fresh console for any console-subsystem child unless `CREATE_NO_WINDOW` (`0x08000000`) is set. Stdio was already piped everywhere, so the change is purely cosmetic.

`cargo check` and `cargo fmt --check` clean.

h5kk:
Follow-up commit ([`1d19eab9`](https://github.com/h5kk/jcode/commit/1d19eab9)): **tui: raise stream stall timeout from 2 to 5 minutes**.

Repro from a real session log: three `Stream stall detected: no server events for Some(120.x s), cancelling` warnings on a single GPT-5.5 over-WS session, all firing exactly at 120s. The model was thinking through a substantial planning prompt and silently exceeded the watchdog before producing any tokens.

The 120s threshold in `tui::app::remote::detect_and_cancel_stall` is too aggressive for current reasoning-heavy models. WS dead-connection detection is independent (`persistent_ws_idle_needs_healthcheck`), so a wider window only affects the "model is thinking quietly" case. Bumped to 300s.

A future improvement could plumb provider-side reasoning-preview events to the watchdog so it resets on signal rather than time; for now the wider window covers the common case.

h5kk:
Follow-up commit ([`8ee36636`](https://github.com/h5kk/jcode/commit/8ee36636)): **windows: extend suppress_child_console sweep to per-turn / refresh hot paths**.

After the previous sweep (ad139a9b) the user reported still seeing PowerShell windows briefly flicker. Source-walked the remaining `Command::new` sites that run from server context and spawn console-subsystem programs; the per-turn and per-refresh paths were the dominant remaining flicker source:

| Site | Trigger | Calls |
|---|---|---|
| `src/prompt.rs` | every system-prompt build (every turn) | 3× git (rev-parse, branch, status) |
| `src/build/source_state.rs` | side-panel git state refresh | 6× git (1 in `git_output_bytes` helper + 5 standalone funcs) |
| `src/build.rs::dirty_status_paths` | build freshness check | 1× git status |
| `src/auth/copilot.rs` | Copilot auth probe | 1× gh |
| `src/auth/cursor.rs::read_vscdb_key` | Cursor IDE state probe | 1× sqlite3 |
| `src/cli/commands.rs::detect_tailscale_dns_name` | tailscale dns probe | 1× tailscale |

All sites already pipe / one-shot `Output` their stdio, so adding `CREATE_NO_WINDOW` is purely cosmetic. `hot_exec.rs` cargo/git invocations are intentionally left alone — those are explicit user-triggered commands (`/rebuild`, `/selfdev`) where seeing the window output is desirable.

Issue #63, opened 2026-04-28
Publish checksums and signatures for release artifacts


carmin777:
## Feature request

jcode currently provides a convenient install path through a remote installer script and GitHub release binaries. For a tool that can execute shell commands, modify files, read credentials, run browser/Gmail integrations, and operate with persistent memory, it would be useful to make release artifact verification explicit and easy.

Please consider publishing checksums and/or signatures for release artifacts, and optionally teaching the installer to verify them before installing.

## Why this matters

Coding agents are high-trust local tools. Users often run them inside private repositories and may connect them to provider credentials, OAuth flows, MCP servers, browser sessions, and other local automation.

Even if the codebase itself is trustworthy, users still need confidence that the binary they install matches the intended release artifact and has not been corrupted, replaced, or tampered with in transit or in the release pipeline.

This is especially relevant because the common install pattern is convenient but high trust:

```bash
curl -fsSL https://raw.githubusercontent.com/1jehuang/jcode/master/scripts/install.sh | bash
```

That pattern is common across CLI tools, but it gives cautious users no obvious way to verify the downloaded binary before first execution.

## Suggested behavior

For each release, publish one or more of:

- `SHA256SUMS`
- `SHA256SUMS.sig`
- minisign/signify signatures
- cosign signatures or attestations
- SLSA provenance, if the release pipeline already supports it

The installer could then:

1. Download the selected release artifact.
2. Download the checksum/signature file.
3. Verify the artifact before installing.
4. Fail closed if verification fails.
5. Print a clear message if verification is unavailable for an older release.

## Possible MVP

A small first step could be:

1. Generate a `SHA256SUMS` file during release.
2. Upload it next to the release binaries.
3. Document manual verification in the README or install docs.

Example documentation:

```bash
curl -LO https://github.com/1jehuang/jcode/releases/download/<version>/<artifact>
curl -LO https://github.com/1jehuang/jcode/releases/download/<version>/SHA256SUMS
sha256sum --check SHA256SUMS
```

A later step could add signature verification and make the installer perform verification automatically.

## Installer UX

If automatic verification is added, the installer could expose behavior such as:

```bash
curl -fsSL https://raw.githubusercontent.com/1jehuang/jcode/master/scripts/install.sh | bash
```

Default behavior:

- verify checksums/signatures when available
- abort on mismatch
- clearly print what was verified

Optional behavior:

```bash
JCODE_INSTALL_SKIP_VERIFY=1 bash install.sh
```

[...]

PR #70, opened 2026-04-28
feat(release): publish SHA256SUMS for release artifacts

upstate-web-co:
Closes part of #63 — the MVP requested in the issue body: *"publish `SHA256SUMS` next to release artifacts and document manual verification."* Sigstore / cosign / installer-side auto-verification are deliberately out of scope here and remain tracked in the issue for follow-up.

## What changes

**`.github/workflows/release.yml`** — two new manifest-generation steps, each run as the last step before the corresponding `softprops/action-gh-release` upload in its job:

- `release` job → generates `SHA256SUMS` covering `jcode-linux-x86_64.tar.gz` + `jcode-macos-aarch64.tar.gz`
- `upload-windows-assets` job → generates `SHA256SUMS-windows` covering the two Windows tarballs + their `.exe` companions

The two-file split mirrors the existing job structure (Windows already ships as a separate upload step), avoids cross-job artifact-passing complexity, and keeps each manifest verifiable by end users who only download a single platform's binary.

Filenames in each manifest are **bare basenames** (no path prefix), so users can drop the artifact and the manifest into a flat directory and run `sha256sum --check SHA256SUMS` directly.

**`README.md`** — new *"Verifying release artifacts"* section under *Detailed Installation*. Copy-pasteable verification commands for:

- Linux: `sha256sum --check --ignore-missing SHA256SUMS`
- macOS: `shasum -a 256 -c --ignore-missing SHA256SUMS` (since `sha256sum` is not on the default toolchain)
- Windows PowerShell: `Get-FileHash` + manual compare against the line in the manifest

The section also explicitly notes that signatures + installer-side auto-verify are tracked separately in #63, so future readers don't think this PR closed the whole issue.

## Verification

The exact bash logic from the workflow was lifted into a local sandbox and run end-to-end:

1. Created a fake nested `artifacts/jcode-{linux,macos}-…/` tree to mirror the post-`download-artifact` layout
2. Ran the `find … | sort -z | sha256sum | basename` pipeline → produced a clean two-line manifest
3. Simulated end-user flow: copied `SHA256SUMS` + one tarball into a flat dir → `shasum -a 256 -c --ignore-missing` returned `OK`
4. Tampering test: corrupted the tarball → verification correctly returned `FAILED` and exited non-zero

`python3 -c "import yaml; yaml.safe_load(open('.github/workflows/release.yml'))"` passes; the workflow file is structurally valid.

## Why this is small

Per the issue body's *"Possible MVP"* section, this PR scope is exactly:

1. Generate a `SHA256SUMS` file during release ✅
2. Upload it next to the release binaries ✅
3. Document manual verification in the README ✅

Sigstore / cosign signatures, SLSA provenance, and `JCODE_INSTALL_SKIP_VERIFY=1` installer integration are intentionally NOT in this PR — each is a meaningful next step on its own and will land cleaner as separate PRs once this baseline is in.

## Trade-offs / things to double-check

- **Two manifest files vs one combined.** A combined manifest would require the Windows job to depend on the release job's `SHA256SUMS` artifact (or vice versa) and be appended to. Splitting by platform keeps each job self-contained and idempotent. Happy to switch to a combined `SHA256SUMS.txt` in `release` that the Windows job appends to if you'd prefer — let me know.
- **Manifest filenames at the release level.** Both files attach as plain release assets named `SHA256SUMS` and `SHA256SUMS-windows`. If you prefer a different naming convention (`jcode-${VERSION}-SHA256SUMS.txt`, etc.), it's a one-line change.
- **No installer change in this PR.** The issue suggests teaching `install.sh` to verify before installing. That's a separate concern (network reliability of pulling the manifest, error UX, opt-out flag) and worth its own review cycle.

---
<a href="https://app.blacksmith.sh/1jehuang/codesmith/jcode/pr/70"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -228,6 +228,22 @@ jobs:
           path: artifacts
           pattern: jcode-*
 
+      - name: Generate SHA256SUMS
+        run: |
+          set -euo pipefail
+          # Collect every release tarball into a flat manifest. Filenames in
+          # SHA256SUMS are bare (no path prefix) so users can verify with
+          # `sha256sum --check SHA256SUMS` alongside the downloaded artifact.
+          (
+            cd artifacts
+            find . -type f -name '*.tar.gz' -print0 | sort -z | while IFS= read -r -d '' f; do
+              sha=$(sha256sum "$f" | cut -d' ' -f1)
+              printf '%s  %s\n' "$sha" "$(basename "$f")"
+            done
+          ) > artifacts/SHA256SUMS
+          echo "Generated SHA256SUMS:"
+          cat artifacts/SHA256SUMS
+
       - name: Create GitHub Release
         uses: softprops/action-gh-release@v2
         with:
@@ -235,6 +251,7 @@ jobs:
           files: |
             artifacts/jcode-linux-x86_64/*.tar.gz
             artifacts/jcode-macos-aarch64/*.tar.gz
+            artifacts/SHA256SUMS
 
       - name: Update Homebrew formula
         env:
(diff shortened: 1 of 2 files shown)


1jehuang:
Thanks for the contribution! This has since been fixed in master (see 3908213d), so I'm closing this PR.

PR #75, opened 2026-04-29
fix(agent): hot-init providers on session restore credential miss (#74)

### LINKED ISSUE #74: [BUG] Reload/restore prompts for Claude OAuth login despite valid auth.json
## Summary

After a reload (process restart triggered via `restart-snapshot` auto-restore), jcode sometimes prompts the user for an OAuth login *even though `~/.jcode/auth.json` already contains valid, unexpired Anthropic credentials*. The restore then fails with:

```
Failed to restore session model 'claude-opus-4-6': Claude credentials not available. Run `jcode login --provider claude` first.
```

Once the user re-runs the OAuth flow, the same on-disk credentials work fine ("Hot-initialized Anthropic provider after auth change"), so this is a **race / ordering bug**, not an actual credential problem. The user-facing symptom is "I have to log in to Claude every time I start jcode".

## Reproduction

1. Run jcode and complete a Claude OAuth login (so `~/.jcode/auth.json` is populated with a long-lived refresh token).
2. Trigger a reload that produces a `restart-snapshot.bak` with `auto_restore_on_next_start: true` (e.g. self-dev reload, upgrade, or unexpected shutdown).
3. On next start, jcode resumes the snapshotted session.

Expected: The session restores silently using the on-disk credentials.

Actual: The TUI shows `Reload complete — continuing because a recovery directive was pending.` immediately followed by `**Claude OAuth Login** (account: claude-1) — Opening browser for authentication...` and the user has to paste a new OAuth code.

## Evidence (from `~/.jcode/logs/jcode-2026-04-29.log`)

```
[11:16:10.992] Anthropic available models [oauth::claude-1]: claude-opus-4-6, ...   ← creds work fine
[11:16:13.523] Anthropic transport: HTTPS SSE stream (oauth=true)                   ← request succeeds
...
[11:16:52.997] Reload check: session_to_resume=Some("session_butterfly_..."), ...
[11:16:52.998] Failed to restore session model 'claude-opus-4-6': Claude credentials not available. Run `jcode login --provider claude` first.
[11:16:53.018] History payload requested reload recovery continuation: ...
[11:17:43.302] Hot-initialized Anthropic provider after auth change                 ← only after the user re-logs in
```

The auth.json file was *unchanged* between 11:16 and 11:17 (same access/refresh tokens, `expires` ~8 hours in the future). The provider just wasn't hot-initialised on the reload code path.

## Root cause

In `src/agent/turn_execution.rs::restore_session_with_id` (and `Agent::new_with_session` in `src/agent.rs`), session restore calls `self.provider.set_model(&model)` directly. When the underlying `MultiProvider` was constructed without an `AnthropicProvider` (because the previous in-process auth probe didn't see credentials, or because `MultiProvider::fork()` propagates the `None` slot), `set_model_on_provider` returns:

```
Claude credentials not available. Run `jcode login --provider claude` first.
```

The error is just logged via `logging::error(...)` and swallowed; the session keeps its stale model field, and any later attempt to actually use the provider fails. Meanwhile, `MultiProvider::on_auth_changed()` is exactly the entry point that *would* hot-initialise the Anthropic provider from `auth::claude::load_credentials()` — but it's only invoked from the login flow, not from the session-restore path.

## Proposed fix

When session restore receives the "credentials not available" error from `set_model`, call `provider.on_auth_changed()` once to give the provider a chance to lazily pick up on-disk credentials, then retry `set_model`. This is cheap (it's just a few file reads) and idempotent.

Optionally, also: when the server starts via the `restart-snapshot` auto-restore path, run `provider.on_auth_changed()` immediately after constructing the `MultiProvider`, so the very first session restore already sees the latest on-disk credentials.

I'll send a PR with the minimal version (retry on `set_model` failure inside `Agent::new_with_session` and `restore_session_with_id`).

## Environment

- jcode v0.11.1 (commit `1f622e6b`)
- Linux x86_64
- Single anthropic OAuth account (`claude-1`)

Kantosaurus:
## Summary

Fixes #74.

When restoring a session, jcode prompts for an OAuth login on every reload even though `~/.jcode/auth.json` already holds valid, unexpired Claude credentials. The user-visible message is:

> Reload complete — continuing because a recovery directive was pending.
>
> **Claude OAuth Login** (account: `claude-1`)
>
> Opening browser for authentication...

The log evidence (see #74) shows the on-disk credentials work fine before and after the prompt — they only stop "being seen" during the session-restore window.

## Root cause

`Agent::new_with_session` (`src/agent.rs`) and `Agent::restore_session` (`src/agent/turn_execution.rs`) call `provider.set_model(&model)` to re-apply the persisted model selection. When the running `MultiProvider` was constructed before an Anthropic sub-provider could be hot-initialised — e.g. across a `restart-snapshot` auto-restore where credentials were written between the original auth probe and the reload — `set_model_on_provider` returns:

```
Claude credentials not available. Run `jcode login --provider claude` first.
```

The error was logged via `logging::error(...)` and silently swallowed; the provider stayed in its credential-less state and any subsequent attempt to use it tripped the OAuth flow. Meanwhile, `MultiProvider::on_auth_changed()` is exactly the entry point that re-scans `~/.jcode/auth.json` and hot-initialises the Anthropic provider — but it was only invoked from the login flow.

## Fix

Introduce `set_session_model_with_lazy_auth_init` in `src/agent.rs`. It calls `provider.set_model(model)` and, on failure, invokes `provider.on_auth_changed()` once and retries. `on_auth_changed` is idempotent and cheap (a few file reads), and a no-op for already-configured providers. If the second attempt still fails, the original error is returned verbatim so genuinely missing credentials produce the same message as before.

Both session-restore call sites now go through this helper.

## Tests

Added three unit tests in `src/agent_tests.rs`:

- `set_session_model_with_lazy_auth_init_recovers_after_on_auth_changed` — stub provider whose first `set_model` returns the canonical "credentials not available" error, and whose `on_auth_changed` flips an internal flag. Verifies the helper returns `Ok`, `set_model` is called exactly twice, and `on_auth_changed` is called exactly once.
- `set_session_model_with_lazy_auth_init_returns_original_error_on_persistent_failure` — when the provider never has credentials, the original error message is preserved.
- `set_session_model_with_lazy_auth_init_does_not_retry_when_first_call_succeeds` — `on_auth_changed` is never invoked on the success path (no perf regression for the common case).

Note: `cargo check` passes locally; full `cargo test --no-run` was still compiling at submission time and will be validated on the reviewer's hardware / CI.

## Diff

```
 src/agent.rs                |  36 +++++++-
 src/agent/turn_execution.rs |   2 +-
 src/agent_tests.rs          | 197 ++++++++++++++++++++++++++++++++++++++++++++
 3 files changed, 233 insertions(+), 2 deletions(-)
```

## Risk

Low. The behaviour change is strictly additive on the failure path:

- Success path: identical (one `set_model` call, no `on_auth_changed`).
- Failure path: one extra `on_auth_changed` call (already used safely from the login flow) plus one `set_model` retry. Worst-case latency on a true credential miss is a few file-system reads.

No public APIs change. The new free function is `pub(crate)`.

---
<a href="https://app.blacksmith.sh/1jehuang/codesmith/jcode/pr/75"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -125,6 +125,40 @@ pub struct Agent {
     stdin_request_tx: Option<tokio::sync::mpsc::UnboundedSender<crate::tool::StdinInputRequest>>,
 }
 
+/// Try to apply a session's persisted model selection to the provider, lazily
+/// hot-initialising any sub-providers from on-disk credentials if the first
+/// attempt fails because the provider believes credentials are unavailable.
+///
+/// This guards the session-restore path against a race where the running
+/// `MultiProvider` instance was constructed before the user finished an OAuth
+/// flow (e.g. across a `restart-snapshot` reload). Without this, restore would
+/// silently log an error and prompt the user to log in again even though
+/// `~/.jcode/auth.json` already holds valid credentials.
+pub(crate) fn set_session_model_with_lazy_auth_init(
+    provider: &dyn Provider,
+    model: &str,
+) -> anyhow::Result<()> {
+    match provider.set_model(model) {
+        Ok(()) => Ok(()),
+        Err(initial_err) => {
+            // Best-effort: ask the provider to re-scan on-disk credentials and
+            // retry once. `on_auth_changed` is idempotent and cheap (a few file
+            // reads), and is the same hook the login flow uses.
+            provider.on_auth_changed();
+            match provider.set_model(model) {
+                Ok(()) => {
+                    logging::info(&format!(
+                        "Recovered session model '{}' after lazy auth re-init",
+                        model
+                    ));
+                    Ok(())
+                }
+                Err(_) => Err(initial_err),
+            }
+        }
+    }
+}
+
 impl Agent {
     fn should_track_client_cache(&self) -> bool {
         match std::env::var("JCODE_TRACK_CLIENT_CACHE") {
@@ -212,7 +246,7 @@ impl Agent {
                 crate::session::derive_session_provider_key(agent.provider.name());
         }
         if let Some(model) = agent.session.model.clone() {
-            if let Err(e) = agent.provider.set_model(&model) {
+            if let Err(e) = set_session_model_with_lazy_auth_init(&*agent.provider, &model) {
                 logging::error(&format!(
                     "Failed to restore session model '{}': {}",
                     model, e
--- a/src/agent/turn_execution.rs
+++ b/src/agent/turn_execution.rs
@@ -369,7 +369,7 @@ impl Agent {
 
         let model_start = Instant::now();
         if let Some(model) = self.session.model.clone() {
-            if let Err(e) = self.provider.set_model(&model) {
+            if let Err(e) = super::set_session_model_with_lazy_auth_init(&*self.provider, &model) {
                 logging::error(&format!(
                     "Failed to restore session model '{}': {}",
                     model, e
--- a/src/agent_tests.rs
+++ b/src/agent_tests.rs
@@ -592,3 +592,200 @@ async fn env_snapshot_detail_is_minimal_for_empty_sessions_and_full_after_histor
 
     assert_eq!(agent.env_snapshot_detail(), EnvSnapshotDetail::Full);
 }
+
+/// Stub provider used to exercise `set_session_model_with_lazy_auth_init`.
+///
+/// The first call to `set_model` returns the same "credentials not available"
+/// error the real `MultiProvider` raises when an Anthropic sub-provider hasn't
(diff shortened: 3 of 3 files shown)


1jehuang:
Thanks for the contribution! This area has since been reworked/fixed in master, so I'm closing this PR.

PR #77, opened 2026-04-29
telemetry: runtime endpoint config + opt-in by default

bjornmage:
## What

Replace the hardcoded `TELEMETRY_ENDPOINT` constant with a runtime resolver: env var → `~/.jcode/telemetry.toml` → `None`. When no endpoint is configured, telemetry is disabled.

## Why

Two motivations, both about giving operators more control without changing what's collected.

1. **Operator-routable endpoint.** Today the collector URL is baked into the binary. Operators who want to keep usage data inside their own infrastructure (org-wide rollout, air-gapped, compliance-driven) have no path other than patching the source. Adding `JCODE_TELEMETRY_ENDPOINT` (env) and a `[telemetry] endpoint = "..."` config field (in `~/.jcode/telemetry.toml`, alongside the existing `mcp.json` / `openai-compatible.env` conventions) gives them a clean, documented knob.

2. **Opt-in default.** With the endpoint now a config value, the natural default for "no endpoint configured" is "no events sent." This means a fresh install doesn't begin phoning home before the user has had a chance to read TELEMETRY.md. Existing opt-out mechanisms (`JCODE_NO_TELEMETRY`, `DO_NOT_TRACK`, `~/.jcode/no_telemetry`) are unchanged and continue to work.

The instrumentation itself is unchanged — same schema, same fields, same coarseness guarantees. This patch only affects the destination.

## Behavior matrix

| Configuration | Result |
|---|---|
| Nothing set (default) | telemetry disabled |
| `JCODE_TELEMETRY_ENDPOINT=https://...` | telemetry enabled, sends to that URL |
| `endpoint = "https://..."` in `~/.jcode/telemetry.toml` | telemetry enabled, sends to that URL |
| Both set | env var wins |
| `JCODE_TELEMETRY_ENDPOINT=""` | telemetry disabled (explicit) |
| `JCODE_NO_TELEMETRY=1` or `DO_NOT_TRACK=1` | telemetry disabled (overrides endpoint) |

## Tests

Added positive coverage:

- `test_default_no_endpoint_disables` — fresh install with nothing configured does not emit.
- `test_endpoint_via_env_var_enables` — env var enables telemetry.
- `test_endpoint_empty_string_disables` — explicit empty string disables.
- `test_no_telemetry_flag_overrides_endpoint` — opt-out flag still wins.

Two pre-existing session-state tests (`test_record_token_usage_aggregates_session_and_turn`, `test_record_connection_type_buckets_transport`) now explicitly set `JCODE_TELEMETRY_ENDPOINT` before calling `begin_session_with_mode`, since that function correctly short-circuits when telemetry is disabled. They use the existing `lock_test_env` + `lock_telemetry_test_state` mutexes to stay isolated.

`cargo test --lib telemetry` — 16/16 pass locally on `rustc 1.93.0`. `cargo fmt` clean.

## Compatibility

- Existing operators who relied on the prior default endpoint can restore that behavior by exporting `JCODE_TELEMETRY_ENDPOINT=https://jcode-telemetry.jeremyhuang55555.workers.dev/v1/event` or writing it to their `telemetry.toml`. Suggest documenting this in TELEMETRY.md as a follow-up so existing users have a clear migration path.
- All opt-out paths preserved.
- Schema version unchanged.

## Splitting if you'd prefer

Happy to split this into two PRs if the default-flip is the controversial part:

1. Just the env var + config-file plumbing, default unchanged.
2. The default flip from on → off.

Either way, the `JCODE_TELEMETRY_ENDPOINT` knob is independently useful for operators running jcode at scale.

---
<a href="https://app.blacksmith.sh/1jehuang/codesmith/jcode/pr/77"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/src/telemetry.rs
+++ b/src/telemetry.rs
@@ -11,7 +11,10 @@ use std::sync::Mutex;
 use std::sync::atomic::{AtomicU32, Ordering};
 use std::time::{Duration, Instant};
 
-const TELEMETRY_ENDPOINT: &str = "https://jcode-telemetry.jeremyhuang55555.workers.dev/v1/event";
+// Telemetry endpoint is now resolved at runtime — see telemetry_endpoint().
+// Default in this fork is no endpoint, which disables telemetry. Set
+// JCODE_TELEMETRY_ENDPOINT or write `endpoint = "..."` to
+// `${JCODE_HOME:-~/.jcode}/telemetry.toml` to enable.
 const ASYNC_SEND_TIMEOUT: Duration = Duration::from_secs(5);
 const BLOCKING_INSTALL_TIMEOUT: Duration = Duration::from_millis(1200);
 const BLOCKING_LIFECYCLE_TIMEOUT: Duration = Duration::from_millis(800);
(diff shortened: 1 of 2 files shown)


1jehuang:

PR #83, opened 2026-04-30
Improve host normalization to respect user-provided schemes

RinZ27:
Current host normalization logic silently strips secure schemes like `https://` or `wss://` but fails to update the internal `use_tls` flag automatically. Such behavior introduces a potential connection downgrade risk if the user's intent (expressed via the URL) isn't explicitly mirrored in the configuration settings.

Updating `MobileGatewayConfig::new` ensures consistency by inferring the correct TLS state directly from the host string prefix. Synchronizing these values eliminates configuration mismatches that often occur when managing secure URLs and protocol flags separately.

Validation includes a new unit test suite within `protocol.rs` covering these edge cases. Everything is green across the `jcode-mobile-core` crate, confirming the improved robustness of the gateway connection logic.

### DIFF
--- a/crates/jcode-mobile-core/src/protocol.rs
+++ b/crates/jcode-mobile-core/src/protocol.rs
@@ -117,8 +117,13 @@ pub struct MobileGatewayConfig {
 }
 
 impl MobileGatewayConfig {
-    pub fn new(host: impl Into<String>, port: u16, use_tls: bool) -> anyhow::Result<Self> {
-        let host = normalize_gateway_host(&host.into())?;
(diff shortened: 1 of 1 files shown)


1jehuang:
Thanks! This repo no longer accepts external PRs, but the underlying problem is real - it's now tracked in #375 (with credit to this PR). Closing.

RinZ27:
Thanks for the update and for the credit on #375. Totally understand the policy change. Glad to see the fix is moving forward internally—the TLS state sync should definitely help keep the gateway config consistent.

PR #95, opened 2026-05-02
[codex] Align JCode OpenAI base URL with Codex config

Cloudymap1e:
## Summary
- read the active Codex CLI `model_provider` from `~/.codex/config.toml`
- when that provider uses `wire_api = "responses"` and `requires_openai_auth = true`, route JCode OpenAI Responses traffic to its `base_url`
- apply the same base URL override to sidecar OpenAI calls
- add regression coverage for config parsing and response URL selection
- keep session `reasoning_effort` persisted so overnight session cloning compiles with the existing field usage
- update OpenRouter test constructors for existing provider fields

## Why
JCode could trust Codex auth, but its OpenAI provider still hard-coded OpenAI/ChatGPT base URLs. If Codex was configured to use a Responses-compatible gateway, JCode would send the stale API key to the wrong endpoint and fail with invalid key errors.

## Impact
This enables JCode to reuse any active Codex-configured HTTP(S) Responses-compatible base URL that explicitly requires OpenAI auth. It does not treat arbitrary chat-completions providers as OpenAI Responses endpoints.

## Validation
- `cargo test configured_responses_base_url`
- `cargo test responses_url_uses_active_codex_provider_base_url`

---
<a href="https://app.blacksmith.sh/1jehuang/codesmith/jcode/pr/95"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


1jehuang:
Thanks! This repo no longer accepts external PRs, but the underlying problem is real - it's now tracked in #374 (with credit to this PR). Closing.

PR #101, opened 2026-05-02
Improve Linux stdin detection for nested wrapper process trees

jrish11:
## Summary
- make Linux stdin detection traverse full descendant process tree, not only direct children
- keep existing stdin-pipe matching guard to avoid false positives
- add regression test for nested wrappers ()

## Why
Wrapper chains are common in agent/tool launch flows. The previous direct-child-only scan could miss the actual stdin reader when it lived at grandchild depth.

## Notes
- This is a targeted Linux reliability improvement.
- I could not run local Rust tests in this environment because  is unavailable here.

---
<a href="https://app.blacksmith.sh/1jehuang/codesmith/jcode/pr/101"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/crates/jcode-core/src/stdin_detect.rs
+++ b/crates/jcode-core/src/stdin_detect.rs
@@ -22,6 +22,7 @@ pub fn is_waiting_for_stdin(pid: u32) -> StdinState {
 #[cfg(target_os = "linux")]
 pub mod linux {
     use super::*;
+    use std::collections::{HashMap, HashSet};
 
     pub fn check(pid: u32) -> StdinState {
         check_inner(pid, false)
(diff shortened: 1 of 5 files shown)


1jehuang:
Thanks! This repo no longer accepts external PRs, but the underlying problem is real - it's now tracked in #373 (with credit to this PR). Closing.

PR #162, opened 2026-05-07
fix(skill+gemini): forgive Skill schema confusion + sanitize JSON-Schema for Gemini

setterapp:
This PR fixes two model-side / provider-side bugs surfaced when running v0.12.0 against MCP servers (notion, supabase) on macOS.

---

## Fix 1: `skill_manage` schema confusion (commit f477bd4)

Models invoking jcode produce malformed calls like `skill_manage(action="load")` (missing `name`) and error with `'name' is required for load action`. Root cause: the public-facing tool is `Skill` with schema `{skill}`, internal tool is `skill_manage` with schema `{action, name}`. The translation layer in `src/tool/mod.rs` was half-applied at v0.12.0 (mapping draft + `transform_input` were committed but the `json` import was missing — the file did not compile).

This PR:
1. Adds the missing `json` import to `src/tool/mod.rs`.
2. Keeps the `Skill | skill -> skill_manage` resolve and `{skill: X} -> {action: load, name: X}` transform.
3. Adds `#[serde(alias = "skill")]` to `SkillInput.name` so `skill_manage` called directly with `{action: load, skill: brain}` also works.
4. Improves the error message when `name` is genuinely missing.

## Fix 2: Gemini rejects MCP tool schemas (commit on same branch)

Gemini's `generateContent` endpoint uses an OpenAPI 3.0 schema subset and rejects requests with HTTP 400 when tool parameters contain `$defs`, `$ref`, or `$schema` — which is exactly what most MCP servers emit (notion, supabase, etc.) because they generate schemas from typed code (Pydantic / Zod) that produces `$defs+$ref`.

Repro: configure notion or supabase MCP, then `jcode --provider gemini run "hi"` — fails with hundreds of errors like:
```
Unknown name "$defs" at 'request.tools[0].function_declarations[17].parameters'
Unknown name "$ref" at 'request.tools[0].function_declarations[18]...'
Unknown name "$schema" at 'request.tools[0].function_declarations[40]...'
```

This PR extends `gemini_compatible_schema` in `src/provider/gemini.rs` to:
1. Extract `$defs` (or `definitions`) from the root.
2. Recursively walk the schema, resolving `$ref` to the corresponding `$def` and inlining the target.
3. Strip JSON-Schema metadata (`$defs`, `$schema`, `$id`, `$comment`, `$anchor`, `definitions`, `title`) at every level.
4. Cap recursion depth at 24 to handle circular schemas safely.

Preserves the existing `const → enum` rewrite.

---

## Tested

- `cargo check -p jcode --bin jcode` — passes
- `cargo build --release -p jcode --bin jcode` — passes (4m24s first build, 1m08s incremental)
- Smoke 1: Skill tool loads `jcode` skill end-to-end with Claude Sonnet OAuth in TUI.
- Smoke 2: `jcode --provider gemini --model gemini-2.5-flash run "Reply only OK"` from setterapp dir (notion + supabase-dev + supabase-prod MCPs configured) — works after fix, was failing with HTTP 400 before.

## Diff stat

```
 src/tool/mod.rs       | 28 ++++++++++++++++++++++++++++-
 src/tool/skill.rs     | 13 +++++++++++--
 src/provider/gemini.rs | 64 +++++++++++++++++++++++++++++++++++++--
 3 files changed, 100 insertions(+), 5 deletions(-)
```

No public API changes.

### DIFF
--- a/src/provider/gemini.rs
+++ b/src/provider/gemini.rs
@@ -890,22 +890,79 @@ pub(crate) fn build_tools(tools: &[ToolDefinition]) -> Option<Vec<GeminiTool>> {
 }
 
 fn gemini_compatible_schema(schema: &Value) -> Value {
+    // Gemini's generateContent uses an OpenAPI 3.0 schema subset and rejects
+    // standard JSON-Schema metadata ($defs, $ref, $schema, etc.).
+    // MCP servers like notion and supabase emit schemas with $defs+$ref.
+    // We extract $defs from the root, then recursively inline $ref references
+    // and strip metadata keywords Gemini doesn't accept.
+    let defs = schema
+        .as_object()
+        .and_then(|m| m.get("$defs").or_else(|| m.get("definitions")))
+        .and_then(|v| v.as_object())
+        .cloned()
+        .unwrap_or_default();
+    sanitize_for_gemini(schema, &defs, 0)
+}
+
+const GEMINI_SCHEMA_MAX_DEPTH: usize = 24;
+
+fn sanitize_for_gemini(
+    schema: &Value,
+    defs: &serde_json::Map<String, Value>,
+    depth: usize,
+) -> Value {
+    if depth > GEMINI_SCHEMA_MAX_DEPTH {
+        // Avoid blow-up on circular schemas — return permissive empty object
+        return Value::Object(serde_json::Map::new());
+    }
     match schema {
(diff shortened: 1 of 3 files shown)


1jehuang:
Thanks for the contribution! This has since been fixed in master (see 3ce8d677), so I'm closing this PR.

PR #193, opened 2026-05-11
Stop runaway emergency compaction when compacted_count drifts past history

### LINKED ISSUE #175: Repeated emergency compaction loop accumulates [Emergency compaction] markers and inflates compacted_count past messages.len() (active_messages() fallback)
## Summary

`hard_compact_with` can fire repeatedly within a single live session and accumulate dozens of `[Emergency compaction]` markers in the persisted summary while never actually shrinking the context. Once the loop is established, `Manager::compacted_count` grows monotonically past `all_messages.len()`, the `active_messages()` fallback silently returns the entire transcript, every recovery attempt ends up sending the full ~1.5M-token payload to the provider, and the session is effectively wedged. The user only sees the symptom: a stalling agent and (eventually, after a real run is closed and reopened) a saved compaction summary that contains many copies of the same `[Emergency compaction]: ~380 messages were dropped` line.

I hit this on `jcode v0.12.0 (30184260)` with the OpenRouter `claude-opus-4-7` route inside a long-running `home-compass` session. Saved session id: `session_spider_1778237530131_e8059d10b1c48c83`.

## Repro environment

- Build: `jcode v0.12.0 (30184260)`, fresh tree at upstream `master @ 4b97d322`.
- OS: macOS 15.x on Apple Silicon (`aarch64-apple-darwin`), Rust `1.95.0`.
- Provider: OpenRouter, model `claude-opus-4-7`.
- Session: long, multi-cwd (`/Users/beta/work/hailuo-tool-builder`, `/Users/beta/work/lobehub`, `/Users/beta/work/home-compass`), used many `bash`/`read`/`agentgrep`/`write` tools.

## Observed symptoms

The persisted summary in `~/.jcode/sessions/session_spider_1778237530131_e8059d10b1c48c83.json` (`compaction.summary_text`) ends with three identical-looking emergency compaction notes:

```text
**[Emergency compaction]**: 379 messages were dropped to recover from context overflow. The conversation had ~1530k tokens which exceeded the 200k limit.

Tools used: agentgrep, bash, batch, edit, multiedit, open, read, session_search, skill_manage, todo, write

Files referenced: BasemapLayer.ts, DistrictsLayer.ts, …

**[Emergency compaction]**: 381 messages were dropped to recover from context overflow. The conversation had ~1530k tokens which exceeded the 200k limit.

Tools used: agentgrep, bash, batch, edit, multiedit, open, read, session_search, skill_manage, todo, write

Files referenced: BasemapLayer.ts, DistrictsLayer.ts, …

**[Emergency compaction]**: 383 messages were dropped to recover from context overflow. The conversation had ~1531k tokens which exceeded the 200k limit.

Tools used: agentgrep, bash, batch, edit, multiedit, open, read, session_search, skill_manage, todo, write

Files referenced: BasemapLayer.ts, DistrictsLayer.ts, …
```

Aggregate session state right after the live run finished:

| field | value |
| --- | --- |
| `messages` (vec length) | `393` |
| `compaction.compacted_count` | `5953` |
| `compaction.covers_up_to_turn` | `383` |
| `compaction.original_turn_count` | `383` |
| `compaction.summary_text` length | `6211` chars |
| count of `[Emergency compaction]` markers in `summary_text` | `3` |

The session journal (`session_spider_1778237530131_e8059d10b1c48c83.journal.jsonl`) records a fourth marker — recorded after a reconnect/resume mid-flight:

```text
**[Emergency compaction]**: 33 messages were dropped to recover from context overflow. The conversation had ~934k tokens which exceeded the 200k limit.
```

After that point the journal pins `compacted_count = 426` while the in-memory message vector still has 393 entries (`compacted_count > messages.len()` — the bug condition).

The corresponding `~/.jcode/logs/jcode-2026-05-08.log` lines for this session show the wider context: an `expected value at line 1 column 1` protocol error tearing down `RemoteConnection`, the session reload, and then several `Memory agent periodic reset at turn N` lines and `TUI_SLOW_FRAME` warnings as the model kept looping. Notable lines:

```text
[2026-05-08 19:00:52.075] [WARN] RemoteConnection::next_event: protocol error=expected value at line 1 column 1 line="…" (session_id=Some("session_spider_1778237530131_e8059d10b1c48c83"), client_instance_id=Some("client_1778237530122_3326104883390537216"))
[2026-05-08 19:00:52.075] [WARN] handle_disconnect: session=None, remote_session_id=Some("session_spider_1778237530131_e8059d10b1c48c83"), reason=Some(Protocol("expected value at line 1 column 1")), detail=protocol error while reading server event: expected value at line 1 column 1
[2026-05-08 19:00:52.080] [INFO] [TIMING] remote bootstrap: connected after 521957ms (resume=Some("session_spider_1778237530131_e8059d10b1c48c83"), reconnect_attempts=1)
[2026-05-08 19:00:52.133] [INFO] [TIMING] session_load: session=session_spider_1778237530131_e8059d10b1c48c83, snapshot=1ms, journal=1ms, finalize=0ms, snapshot_bytes=1875, journal_bytes=84117, journal_entries=37, messages=44, env_snapshots=2, replay_events=0, total=2ms
[2026-05-08 19:00:52.135] [INFO] restore_session: loaded session session_spider_1778237530131_e8059d10b1c48c83 with 44 messages, calling seed_compaction
[2026-05-08 19:00:54.077] [WARN] Session session_spider_1778237530131_e8059d10b1c48c83 cleanup timed out waiting for agent lock (stuck task); skipping graceful shutdown
[2026-05-08 19:26:24.819] [INFO] [session_spider_1778237530131_e8059d10b1c48c83] Memory agent periodic reset at turn 50 (clearing 0 surfaced memories)
[2026-05-08 19:55:37.587] [INFO] [session_spider_1778237530131_e8059d10b1c48c83] Memory agent periodic reset at turn 100 (clearing 0 surfaced memories)
[2026-05-08 20:23:04.487] [INFO] [session_spider_1778237530131_e8059d10b1c48c83] Memory agent periodic reset at turn 150 (clearing 0 surfaced memories)
[2026-05-08 20:32:17.055] [INFO] Final extraction starting for session session_spider_1778237530131_e8059d10b1c48c83 (76695 chars)
[2026-05-08 20:32:19.279] [INFO] Final extraction for session session_spider_1778237530131_e8059d10b1c48c83 failed: OpenAI API error (429 Too Many Requests):
```

There is no explicit `hard_compact` log line because `hard_compact_with` (in `src/compaction.rs`) doesn't log on success — `last_compaction = Some(CompactionEvent { trigger: "hard_compact", … })` is observable only in TUI status. So the only on-disk evidence that emergency compaction fired is the growing `summary_text` and `compacted_count` getting silently larger.

## Why this happens (root cause walkthrough)

Three pieces of code interact:

### 1. `active_messages()` falls back to the full transcript when `compacted_count > all_messages.len()`

`src/compaction.rs:603-614`:

```rust
/// Get the active (uncompacted) messages from a full message list.
/// Skips the first `compacted_count` messages.
fn active_messages<'a>(&self, all_messages: &'a [Message]) -> &'a [Message] {
    if self.compacted_count <= all_messages.len() {
        &all_messages[self.compacted_count..]
    } else {
        // Edge case: messages were cleared/replaced with fewer items
        all_messages
    }
}
```

The fallback returning `all_messages` is intended to be defensive, but it makes the manager *invisible-to-the-caller* lie about what is "active": every downstream consumer that reasons about the size of the active suffix gets the full transcript instead, including:

- `active_message_chars_with` (`src/compaction.rs:614-625`)
- `token_estimate_with` / `effective_token_count_with` / `context_usage_with`
- `messages_for_api_with` (which still prepends the summary block, but ALSO appends every original message)

So once we trip into this branch, the very next question "how big is the context?" gets answered with the original ~1.5M tokens.

### 2. `hard_compact_with` does `compacted_count += cutoff`, where `cutoff` is computed against the *active* slice — including the fallback

`src/compaction.rs:1198-1264` (abridged):

```rust
pub fn hard_compact_with(&mut self, all_messages: &[Message]) -> Result<usize, String> {
    let active = self.active_messages(all_messages);
    …
    let mut turns_to_keep = RECENT_TURNS_TO_KEEP.min(active.len().saturating_sub(1));
    let mut cutoff;
    loop {
        cutoff = active.len().saturating_sub(turns_to_keep);
        cutoff = safe_compaction_cutoff(active, cutoff);
        if cutoff > 0 { … if remaining_tokens <= self.token_budget { break; } }
        if turns_to_keep <= MIN_TURNS_TO_KEEP { … break; }
        turns_to_keep = (turns_to_keep / 2).max(MIN_TURNS_TO_KEEP);
    }
    …
    self.compacted_count += cutoff;
}
```

When we are in the `active_messages` fallback, `active.len() == all_messages.len()`, and so:

```text
cutoff = all_messages.len() - keep
compacted_count_new = compacted_count + (all_messages.len() - keep)
```

Because `compacted_count` was *already* greater than `all_messages.len()`, the new value is even further past the end. Each subsequent `hard_compact_with` adds another `all_messages.len() - keep` to the counter.

### 3. `build_emergency_summary_text` blindly concatenates the previous summary

`crates/jcode-compaction-core/src/lib.rs:355-394`:

```rust
pub fn build_emergency_summary_text(
    existing_summary: Option<&str>,
    dropped_count: usize,
    pre_tokens: u64,
    token_budget: usize,
    dropped_messages: &[Message],
) -> String {
    let mut summary_parts: Vec<String> = Vec::new();
    if let Some(existing) = existing_summary && !existing.is_empty() {
        summary_parts.push(existing.to_string());
    }
    summary_parts.push(format!(
        "**[Emergency compaction]**: {} messages were dropped to recover from context overflow. \
         The conversation had ~{}k tokens which exceeded the {}k limit.",
        dropped_count, pre_tokens / 1000, token_budget / 1000,
    ));
    …
    summary_parts.join("\n\n")
}
```

There is no de-duplication: once the loop runs N times, the summary contains N copies of the emergency block back-to-back.

### Putting it together

1. The session crosses the 200k token budget (the model is `claude-opus-4-7`; the session had ~1.5M tokens of tool output baked into history).
2. `auto_recover_context_limit` (`src/tui/app/model_context.rs:359-409`) calls `hard_compact_with`. The first call legitimately drops `~380` messages and sets `compacted_count = 380`-ish. `observed_input_tokens` is reset to `None`.
3. `messages_for_api_with` still has to prepend the summary. The next request gets sent.
4. The provider rejects with `Protocol("expected value at line 1 column 1")` (the symptom we observe — partial JSON because the upstream stream truncated). The reconnect path (`src/agent/turn_execution.rs:440-460`) fires `restore_session`, which calls `seed_compaction_from_session`.
5. `restore_persisted_state` (`src/compaction.rs:246-273`) clamps `compacted_count = state.compacted_count.min(total_messages)`. **The `state.compacted_count` from disk is now 5953, but `total_messages` is only 44, so the clamp silently throws away the bookkeeping while the persisted summary is preserved verbatim — the summary still claims to cover ~5953 turns.**
6. The model continues, more turns add up, the budget is exceeded again, and the next `hard_compact_with` runs on a fresh-ish active suffix — `compacted_count` jumps from 44 to ~`44 + (active_len - 10)`, but the previous summary still gets concatenated under `build_emergency_summary_text`. From this point on, every compaction adds another `[Emergency compaction]` paragraph.
7. Once any subsequent path causes `compacted_count` to creep just past `all_messages.len()` (very easy after a reload/messages reset/`run_turn_interactive` retry), step 1 of the active-messages fallback fires, and the loop in `try_auto_compact_and_retry` (`src/tui/app/model_context.rs:443-589`) becomes self-perpetuating: every recovery attempt sees `usage > 1.5` again, calls `hard_compact_with` again, which grows the summary by one more emergency block without shrinking the request payload at all.

Each `try_auto_compact_and_retry` then `await`s `run_turn_interactive`, whose error handler can re-enter `try_auto_compact_and_retry` (`src/tui/app/model_context.rs:580-600`), so the loop is mutual-recursive at the runtime level rather than a single iterative loop.

## Minimal reproducer

Save as `src/compaction_tests.rs` (appended after the existing tests) and run with `cargo test test_bug_repeated_emergency_compaction_loop --nocapture`. Reproduces deterministically against `4b97d322`:

```rust
#[test]
fn test_bug_repeated_emergency_compaction_loop() {
    // Reproduces session_spider_1778237530131_e8059d10b1c48c83.
    let mut manager = CompactionManager::new().with_budget(1_000);

    let mut messages = Vec::new();
    for i in 0..30 {
        messages.push(make_text_message(
            Role::User,
            &format!("turn {} content {}", i, "x".repeat(200)),
        ));
        manager.notify_message_added();
    }

    // Pathological state: compacted_count > messages.len() (mirrors what we
    // see on disk after restore_persisted_state clamps a stale higher value
    // and a few more rounds of auto_recover bump it past the new len()).
    manager.compacted_count = 100; // way past 30
    manager.active_summary = Some(jcode_compaction_core::Summary {
        text: String::from("# Existing summary"),
        openai_encrypted_content: None,
        covers_up_to_turn: 100,
        original_turn_count: 100,
    });
    manager.active_message_chars_dirty = true;

    // Drive the auto_recover loop manually (mirrors auto_recover_context_limit).
    let mut emergency_count = 0;
    for _ in 0..6 {
        let usage = manager.context_usage_with(&messages);
        if usage <= 1.5 { break; }
        if manager.hard_compact_with(&messages).is_err() { break; }
        emergency_count += 1;
    }

    let summary_markers = manager.active_summary
        .as_ref()
        .map(|s| s.text.matches("[Emergency compaction]").count())
        .unwrap_or(0);

    eprintln!(
        "emergency_count={}, summary_markers={}, compacted_count={}, messages.len()={}",
        emergency_count, summary_markers, manager.compacted_count(), messages.len()
    );

    assert!(emergency_count <= 1, "loop should converge in 1 emergency, got {}", emergency_count);
    assert!(summary_markers <= 1, "summary should not accumulate emergency markers, got {}", summary_markers);
    assert!(manager.compacted_count() <= messages.len(),
        "compacted_count={} must not exceed messages.len()={}",
        manager.compacted_count(), messages.len());
}
```

Actual output on `master @ 4b97d322`:

```text
running 1 test
emergency_count=6, summary_markers=6, last_dropped=Some(20), compacted_count=220, messages.len()=30

thread 'compaction::tests::test_bug_repeated_emergency_compaction_loop' panicked at src/compaction_tests.rs:824:5:
auto_recover loop should converge in 1 emergency compaction, got 6
test compaction::tests::test_bug_repeated_emergency_compaction_loop ... FAILED
```

(After 6 iterations: `compacted_count` ballooned to `220` while `messages.len()` was still `30`; the summary contains 6 stacked `[Emergency compaction]` blocks.)

A separate test verifying that *normal* emergency compaction (a single huge tool result, no pre-existing pathological state) stays bounded passes (`emergency_count=1, summary_markers=1`), confirming the bug is specifically about the `compacted_count > messages.len()` path.

## Suggested fix areas (for reference, not a PR)

These are starting points — the maintainer-authored fix will probably look different, but for the issue record:

1. **Make `active_messages()` safe.** The fallback should return an empty slice (or self-correct `compacted_count`) instead of returning `all_messages`. Returning the full transcript while the manager still believes it has compacted everything is the actual gun.

   ```rust
   fn active_messages<'a>(&self, all_messages: &'a [Message]) -> &'a [Message] {
       let start = self.compacted_count.min(all_messages.len());
       &all_messages[start..]
   }
   ```

   Plus an assert (debug-only) or a `tracing::warn!` so the inconsistency is visible.

2. **Clamp `hard_compact_with` so `compacted_count` cannot exceed `all_messages.len()`.**

   ```rust
   self.compacted_count = (self.compacted_count + cutoff).min(all_messages.len());
   ```

3. **Don't keep concatenating emergency markers.** `build_emergency_summary_text` should detect that the previous summary already ends with an `[Emergency compaction]` block describing the same `pre_tokens`/`token_budget` and either replace it or merge counts. Even simpler: bump a counter ("3 emergency compactions so far in this session") instead of pasting the same paragraph.

4. **`restore_persisted_state` should not silently clamp.** When `state.compacted_count > total_messages`, that means the on-disk summary is now describing more turns than the current transcript; either reject the restore (drop the summary), or rebuild the summary, or at minimum log a warning. The current behavior (silent clamp + keep the stale summary) is what makes the runtime state self-inconsistent.

5. **`auto_recover_context_limit` / `try_auto_compact_and_retry` should detect failure to converge.** If `hard_compact_with` returns `Ok` but `context_usage_with` still says `>1.5` afterwards, falling straight into another `hard_compact_with` is what burns through `[Emergency compaction]` markers. After the first failed convergence, the runtime should escalate to `emergency_truncate_with` only, or surface a hard error to the user instead of silently retrying.

6. **Surface emergency compaction in logs.** Right now there's no log line on `hard_compact_with` success — the only evidence is the growing summary. Adding `logging::warn!("[compaction] emergency drop: dropped={dropped}, compacted_count={…}, summary_markers={…}")` would have made this much easier to spot.

## Repro artifacts

I can attach (or post out-of-band):
- `~/.jcode/sessions/session_spider_1778237530131_e8059d10b1c48c83.{json,bak,journal.jsonl}`
- `~/.jcode/logs/jcode-2026-05-08.log` filtered to that session id
- the standalone `test_bug_repeated_emergency_compaction_loop` Rust test

Tell me if you want the JSON files attached here or if you'd rather I redact them and post a smaller transcript snippet.

cooleryu:
## Summary

Fixes #175 by making stale compaction state recover instead of re-entering the emergency compaction loop.

This PR:

- clamps `active_messages()` with `min(compacted_count, messages.len())`, so a stale compacted count no longer falls back to replaying the full transcript
- self-heals stale `compacted_count > messages.len()` before API message assembly, background compaction application, and hard compaction
- bounds compaction count advances with saturating arithmetic and `messages.len()` when caller history is available
- logs stale persisted compaction state during restore instead of silently clamping it
- adds four regression tests for the #175 failure mode
- applies `cargo fmt --all` formatting needed by the repository CI

## Motivation

The issue report shows a long-running coding-agent session getting wedged after emergency compaction. Once `compacted_count` drifted past the actual message history length, `active_messages()` returned the full transcript again. That made the next API payload include both the summary and old messages, kept the context above the threshold, and allowed hard compaction to append repeated `[Emergency compaction]` markers while increasing `compacted_count` even further.

The important invariant is: if the caller provides the full message list, a compacted count beyond that list cannot mean "all messages are active again". It means the manager has stale state and should recover to an empty active tail.

## Changes

- Treat stale `compacted_count` as a recoverable state and clamp it to `messages.len()`.
- Reset stale active-message accounting when the clamp is applied.
- Preserve the backward-compatible no-history path used by `check_and_apply_compaction()`.
- Prevent hard/background compaction from advancing `compacted_count` beyond caller history.
- Add restore-time warning logs for persisted stale compaction state.
- Add regression coverage for:
  - API message assembly not replaying the full transcript after a summary
  - `active_messages()` clamping stale state to an empty active tail
  - token estimation ignoring stale cached active char counts
  - hard compaction not inflating stale counts or appending another emergency block

## Test Plan

- `cargo test test_bug_175_ -- --nocapture`
  - 4 passed
- `cargo test compaction::tests:: -- --nocapture`
  - 30 passed
- `cargo fmt --all -- --check`
  - passed
- `cargo check --all-targets --all-features`
  - passed

I also installed and tried local clippy. On my macOS machine it stops on pre-existing macOS-only lints in `crates/jcode-core/src/stdin_detect.rs`; I did not mix that unrelated cleanup into this PR.

## Risk

Low to moderate. This changes recovery behavior only when `compacted_count` is already inconsistent with the caller-provided history. Normal compaction paths should continue to use the same active suffix. The no-history compatibility path is preserved so existing callers of `check_and_apply_compaction()` still apply completed background compactions.

## Maintainer Context

The issue already included a useful local hotfix note. This PR turns that direction into a tested upstream fix and keeps the scope limited to the compaction invariant, recovery points, and regression coverage. It intentionally does not add a session-file migration script; that can remain a separate operational recovery step if maintainers want it.

---
<a href="https://app.blacksmith.sh/1jehuang/codesmith/jcode/pr/193"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/crates/jcode-provider-metadata/src/lib.rs
+++ b/crates/jcode-provider-metadata/src/lib.rs
@@ -1346,7 +1346,10 @@ mod tests {
         assert_eq!(OLLAMA_PROFILE.default_model, None);
         assert!(!OLLAMA_PROFILE.requires_api_key);
 
-        assert_eq!(OLLAMA_LOGIN_PROVIDER.auth_kind, LoginProviderAuthKind::Local);
+        assert_eq!(
+            OLLAMA_LOGIN_PROVIDER.auth_kind,
+            LoginProviderAuthKind::Local
+        );
         assert_eq!(OLLAMA_LOGIN_PROVIDER.auth_status_method, "local endpoint");
         assert!(matches!(
             OLLAMA_LOGIN_PROVIDER.target,
--- a/src/compaction.rs
+++ b/src/compaction.rs
@@ -259,6 +259,12 @@ impl CompactionManager {
         self.semantic_embed_cache.clear();
         self.semantic_embed_cache_counter = 0;
         self.total_turns = total_messages;
+        if state.compacted_count > total_messages {
+            crate::logging::warn(&format!(
+                "[compaction] restore: stale compacted_count={} exceeds messages.len()={}; clamping",
+                state.compacted_count, total_messages
+            ));
+        }
         self.compacted_count = state.compacted_count.min(total_messages);
         self.active_message_chars = 0;
         self.active_message_chars_dirty = total_messages > self.compacted_count;
@@ -285,6 +291,29 @@ impl CompactionManager {
             .map(message_char_count)
             .sum();
         self.active_message_chars_dirty = false;
+    }
+
+    fn clamp_stale_compacted_count(&mut self, all_messages: &[Message], caller: &str) -> bool {
+        if all_messages.is_empty() {
+            return false;
+        }
+
+        if self.compacted_count <= all_messages.len() {
+            return false;
+        }
+
+        crate::logging::warn(&format!(
+            "[compaction] {}: stale compacted_count={} exceeds messages.len()={}; clamping",
+            caller,
+            self.compacted_count,
+            all_messages.len()
+        ));
+        self.compacted_count = all_messages.len();
+        self.total_turns = all_messages.len();
+        self.active_message_chars = 0;
+        self.active_message_chars_dirty = false;
+        self.observed_input_tokens = None;
+        true
     }
 
     pub fn restore_persisted_stored_state_with(
@@ -603,22 +632,17 @@ impl CompactionManager {
     /// Get the active (uncompacted) messages from a full message list.
     /// Skips the first `compacted_count` messages.
     fn active_messages<'a>(&self, all_messages: &'a [Message]) -> &'a [Message] {
-        if self.compacted_count <= all_messages.len() {
-            &all_messages[self.compacted_count..]
-        } else {
-            // Edge case: messages were cleared/replaced with fewer items
-            all_messages
-        }
+        let start = self.compacted_count.min(all_messages.len());
+        &all_messages[start..]
     }
 
     fn active_message_chars_with(&self, all_messages: &[Message]) -> usize {
+        let active = self.active_messages(all_messages);
         if self.active_message_chars_dirty
-            || self.active_messages_count() != self.active_messages(all_messages).len()
+            || self.active_messages_count() != active.len()
+            || self.compacted_count > all_messages.len()
         {
-            self.active_messages(all_messages)
-                .iter()
-                .map(message_char_count)
-                .sum()
+            active.iter().map(message_char_count).sum()
         } else {
             self.active_message_chars
         }
@@ -901,6 +925,8 @@ impl CompactionManager {
     /// Check if background compaction is done and apply it, updating rolling
     /// token-estimate state from the provided full message list.
     pub fn check_and_apply_compaction_with(&mut self, all_messages: &[Message]) {
+        self.clamp_stale_compacted_count(all_messages, "check_and_apply_compaction_with");
+
         let task = match self.pending_task.take() {
             Some(task) => task,
             None => return,
@@ -931,7 +957,10 @@ impl CompactionManager {
                 };
 
                 // Advance the compacted count — these messages are now summarized
-                self.compacted_count += self.pending_cutoff;
+                self.compacted_count = self.compacted_count.saturating_add(self.pending_cutoff);
+                if !all_messages.is_empty() {
+                    self.compacted_count = self.compacted_count.min(all_messages.len());
+                }
                 self.active_message_chars = self
                     .active_message_chars_with(all_messages)
                     .saturating_sub(compacted_chars);
@@ -1010,7 +1039,9 @@ impl CompactionManager {
     /// Get messages for API call (with summary if compacted).
     /// Takes the full message list from the caller.
     pub fn messages_for_api_with(&mut self, all_messages: &[Message]) -> Vec<Message> {
+        self.clamp_stale_compacted_count(all_messages, "messages_for_api_with");
         self.check_and_apply_compaction_with(all_messages);
+        self.clamp_stale_compacted_count(all_messages, "messages_for_api_with");
         self.discard_oversized_openai_native_compaction();
 
         let active = self.active_messages(all_messages);
@@ -1196,6 +1227,7 @@ impl CompactionManager {
     /// exceed the token budget, progressively keeps fewer turns down to
     /// `MIN_TURNS_TO_KEEP`.
     pub fn hard_compact_with(&mut self, all_messages: &[Message]) -> Result<usize, String> {
+        self.clamp_stale_compacted_count(all_messages, "hard_compact_with");
         let active = self.active_messages(all_messages);
 
         if active.len() <= MIN_TURNS_TO_KEEP {
@@ -1257,7 +1289,10 @@ impl CompactionManager {
             original_turn_count: cutoff,
         };
 
-        self.compacted_count += cutoff;
+        self.compacted_count = self
+            .compacted_count
+            .saturating_add(cutoff)
+            .min(all_messages.len());
         self.active_message_chars = remaining_suffix_chars[cutoff];
         self.active_message_chars_dirty = false;
         self.active_summary = Some(summary);
--- a/src/compaction_tests.rs
+++ b/src/compaction_tests.rs
@@ -375,6 +375,130 @@ fn test_hard_compact_preserves_recent_turns() {
     );
 }
 
+#[test]
+fn test_bug_175_messages_for_api_self_heals_stale_compacted_count() {
+    let mut manager = CompactionManager::new().with_budget(1_000);
+    let mut messages = Vec::new();
+    for i in 0..30 {
+        messages.push(make_text_message(
+            Role::User,
+            &format!("turn {} {}", i, "x".repeat(120)),
+        ));
+        manager.notify_message_added();
+    }
+
+    manager.compacted_count = 100;
+    manager.total_turns = 100;
+    manager.active_message_chars = messages.iter().map(message_char_count).sum();
+    manager.active_message_chars_dirty = false;
+    manager.active_summary = Some(jcode_compaction_core::Summary {
+        text: "# Existing summary".to_string(),
+        openai_encrypted_content: None,
+        covers_up_to_turn: 100,
+        original_turn_count: 100,
+    });
+
+    let api_messages = manager.messages_for_api_with(&messages);
+
+    assert_eq!(
+        manager.compacted_count(),
+        messages.len(),
+        "stale compacted_count should be clamped to the caller's full message list"
+    );
+    assert_eq!(
+        api_messages.len(),
+        1,
+        "stale compacted_count must not replay the full transcript after the summary"
+    );
+    assert_eq!(manager.stats_with(&messages).active_messages, 0);
+}
+
+#[test]
+fn test_bug_175_active_messages_clamps_stale_compacted_count() {
+    let mut manager = CompactionManager::new();
+    let messages = vec![
+        make_text_message(Role::User, "first"),
+        make_text_message(Role::Assistant, "second"),
+        make_text_message(Role::User, "third"),
+    ];
+
+    manager.compacted_count = 10;
+    manager.total_turns = 10;
+
+    assert!(
+        manager.active_messages(&messages).is_empty(),
+        "a stale compacted_count must produce an empty active tail, not replay all messages"
+    );
+}
+
+#[test]
+fn test_bug_175_token_estimate_ignores_stale_cached_active_chars() {
+    let mut manager = CompactionManager::new().with_budget(1_000);
+    let mut messages = Vec::new();
+    for i in 0..3 {
+        messages.push(make_text_message(
+            Role::User,
+            &format!("turn {} {}", i, "x".repeat(10_000)),
+        ));
+        manager.notify_message_added();
+    }
+
+    manager.compacted_count = 10;
+    manager.total_turns = 10;
+    manager.active_message_chars = 1_000_000;
+    manager.active_message_chars_dirty = false;
+
+    assert_eq!(
+        manager.token_estimate_with(&messages),
+        0,
+        "stale cached active_message_chars should not keep emergency compaction above threshold"
+    );
+}
+
+#[test]
+fn test_bug_175_hard_compact_does_not_inflate_stale_compacted_count() {
+    let mut manager = CompactionManager::new().with_budget(1_000);
+    let mut messages = Vec::new();
+    for i in 0..30 {
+        messages.push(make_text_message(
+            Role::User,
+            &format!("turn {} {}", i, "z".repeat(200)),
+        ));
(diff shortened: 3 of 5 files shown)


1jehuang:
Thanks for the contribution! This has since been fixed in master (see 241c2d0f), so I'm closing this PR.

PR #217, opened 2026-05-15
feat: improve resume picker session management

DarkPhilosophy:
## Summary
- make `/resume` default to the current terminal, with `Ctrl+Enter` preserving the alternate new-terminal action
- add in-picker session deletion with `d`, `D`, or `Delete`, including an explicit confirmation step
- clean up local session artifacts on delete and fall back to the current terminal when a new terminal cannot be spawned

## Rationale
The picker already presents `/resume` as a session-management surface. In practice, the previous default made the common case indirect: `Enter` attempted to spawn another terminal, and if no supported terminal was found the user was left with a manual command to copy instead of resuming the selected session in place.

This keeps the advanced path available while making the default path match the direct user intent:
- `Enter` resumes the currently highlighted session in the current terminal
- `Ctrl+Enter` still performs the alternate terminal action
- multi-select remains useful for batch/new-terminal flows
- deletion is handled where session discovery already happens, with confirmation before removing local artifacts

## Behavior changes
- `/resume` defaults to `current-terminal`
- `d`, `D`, or `Delete` arms deletion; pressing the same key again or `Enter` confirms; `Esc` cancels
- delete removes the local snapshot, journal, backup snapshot, active pid marker, todo state, side-panel state, and client-input artifact for that session id
- if spawning a new terminal fails in the inline picker path, resume falls back to the current terminal instead of ending at a manual command

## Tests
- `cargo test --lib delete_session_artifacts`
- `cargo test --lib session_picker_resume_action`
- `cargo test --lib test_delete_`
- `cargo test --lib test_space_selects_multiple_sessions_and_enter_returns_them`

### DIFF
--- a/crates/jcode-config-types/src/lib.rs
+++ b/crates/jcode-config-types/src/lib.rs
@@ -449,7 +449,7 @@ impl Default for KeybindingsConfig {
             workspace_down: "alt+j".to_string(),
             workspace_up: "alt+k".to_string(),
             workspace_right: "alt+l".to_string(),
-            session_picker_enter: SessionPickerResumeAction::NewTerminal,
+            session_picker_enter: SessionPickerResumeAction::CurrentTerminal,
         }
     }
 }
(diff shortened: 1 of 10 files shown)


1jehuang:
Thanks for the contribution! This area has since been reworked/fixed in master, so I'm closing this PR.

PR #219, opened 2026-05-15
feat: enforce real provider disable policy

DarkPhilosophy:
## Summary
- add a real disabled-provider policy via `[provider].disabled_providers`, `[policy].disabled_providers`, and `JCODE_DISABLED_PROVIDERS`
- keep disabled providers out of startup initialization, auth surfaces, model routing, failover, account/login pickers, and remote `/model` views
- stop treating generic `GH_TOKEN` / `GITHUB_TOKEN` as Copilot credentials unless the user explicitly opts in with `JCODE_COPILOT_ALLOW_GITHUB_ENV=1`

## Rationale
A disabled provider should behave as absent, not as loaded-but-warned. If a provider is disabled, it should not be discovered, initialized, auto-selected, offered as a fallback, or shown as a usable account/model route.

This also removes an especially surprising source of implicit Copilot activation: generic GitHub credentials are commonly present for normal GitHub workflows and are not a Copilot intent signal by themselves.

## Behavior changes
- new disabled-provider config entries are normalized and respected consistently across runtime/provider selection
- disabled providers are filtered from auth status, auth doctor, auth test, account/login pickers, completion suggestions, and remote model snapshots
- startup and hot-refresh paths skip disabled providers instead of initializing them and hiding them later
- model selection rejects explicit use of disabled providers
- fallback sequences exclude disabled providers
- Copilot now only consumes `COPILOT_GITHUB_TOKEN` by default; reusing `GH_TOKEN` / `GITHUB_TOKEN` requires explicit opt-in

## Tests
- `cargo test --lib test_env_override_disabled_providers_normalizes_entries`
- `cargo test --lib copilot_does_not_auto_use_generic_github_env_without_opt_in`
- `cargo test --lib copilot_recent_token_exchange_failure_is_not_auto_usable`
- `cargo test --lib test_enabled_fallback_sequence_omits_disabled_providers`
- `cargo test --lib disabled_login_providers_are_filtered_from_user_facing_lists`

### DIFF
--- a/crates/jcode-config-types/src/lib.rs
+++ b/crates/jcode-config-types/src/lib.rs
@@ -593,6 +593,8 @@ pub struct ProviderConfig {
     pub default_model: Option<String>,
     /// Default provider to use (claude|openai|copilot|openrouter)
     pub default_provider: Option<String>,
+    /// Providers excluded from auto-selection, fallback, and model routing.
+    pub disabled_providers: Vec<String>,
     /// Reasoning effort for OpenAI Responses API (none|low|medium|high|xhigh)
     pub openai_reasoning_effort: Option<String>,
     /// OpenAI transport mode (auto|websocket|https)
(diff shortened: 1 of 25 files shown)


1jehuang:

PR #260, opened 2026-05-24
Allow shell configuration

### LINKED ISSUE #257: Allow shell configuration
I am a big fan of Nushell and I would like to set the default shell used by this tool. However, I checked the code and it does not offer this possibility.

I'd propose to use configuration to set the terminal shell. So, for example:

```toml
[terminal]
shell = "nu"
```

If the configuration is provided, it would override the default shell.

Odonno:
Closes #257


1jehuang:

PR #269, opened 2026-05-26
fix: populate global context-limit cache from named provider model config (#262)

### LINKED ISSUE #262: OpenAI-compatible providers without /v1/models endpoint fall back to hardcoded 200k context limit with no override
When using an OpenAI-compatible provider that doesn't expose a GET /v1/models endpoint (e.g. a custom proxy/gateway), the context window      
  resolution chain has no way to discover the actual context_length for models.                
                                                                                                                                                
  Resolution chain observed:                                                                                                                    
  1. openai_compatible_profile_context_limit() — only has a hardcoded entry for DeepSeek V4; returns None for other models
  2. GET {api_base}/models — returns 404 on providers that don't support it                                                                     
  3. CONTEXT_LIMIT_CACHE — empty (nothing to populate from)                                    
  4. Falls back to DEFAULT_CONTEXT_LIMIT = 200,000                                                                                              
                                                                                                                                                
  This means models with larger context windows (e.g. mimo-v2.5-pro which may support 128k+) are incorrectly limited, and models with smaller   
  windows waste budget on requests that will fail.                                                                                              
                                                                                                                                                
  Steps to reproduce:                                                                                                                           
  1. Configure an OpenAI-compatible provider pointing to a gateway without /v1/models (e.g. opengateway.gitlawb.com/v1)
  2. Use a model like mimo-v2.5-pro                                                                                                             
  3. context_limit_for_model() returns None, resolved as 200k                                  
                                                                                                                                                
  Suggested improvements:                                                                                                                       
  - Allow users to set per-model context limits via config (e.g. context_limits map in provider config)                                         
  - Support a context_window field in NamedProviderConfig model entries                                                                         
  - Fall back to probing (send a request and detect context-limit errors) rather than assuming 200k

mmacedoeu:
Fixes #262

When using an OpenAI-compatible provider that doesn't expose , user-configured  values in  were only used by the provider instance's own  method. All other resolution paths (TUI info widget, compaction budget, model switching) fell through to .

### Changes

**** - In , populate the global  from  values after building . This ensures all codepaths that call  or  resolve the correct limit.

**** - Added test  verifying that:
- A named provider with  for a custom model
- Both  and  resolve to 128,000 (not 200k)

**** - Fix pre-existing compile error: make  module .

### Usage

Users can now set per-model context limits in their provider config:

```toml
[[providers.my-gateway.models]]
id = "mimo-v2.5-pro"
context_window = 128000
```

And the configured limit will be respected globally, even without a working `/v1/models` endpoint.

### DIFF
--- a/crates/jcode-base/src/config.rs
+++ b/crates/jcode-base/src/config.rs
@@ -158,8 +158,15 @@ struct ConfigCache {
 
 static CONFIG_CACHE: LazyLock<RwLock<ConfigCache>> = LazyLock::new(|| {
     let fingerprint = ConfigCacheFingerprint::current();
+    let config = leak_config(Config::load());
+    // Populate the context-limit cache from named provider configs on initial
+    // load so that both the server and TUI processes have correct context
+    // limits from the very start. Read from the loaded config directly to
+    // avoid a deadlock (config() would try to read CONFIG_CACHE which is
+    // still being initialized).
+    populate_context_limits_from_config_ref(config);
     RwLock::new(ConfigCache {
-        config: leak_config(Config::load()),
+        config,
         fingerprint,
         last_checked: Instant::now(),
         force_reload: false,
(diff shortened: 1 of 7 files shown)


1jehuang:
Thanks! This repo no longer accepts external PRs, but the underlying problem is real - it's now tracked in #366 (with credit to this PR). Closing.

PR #280, opened 2026-05-29
fix(provider): restore native OpenRouter endpoint when switching back from a direct profile

prateekjain-afk:
## Problem

Switching models OpenRouter → NVIDIA NIM → back to a native OpenRouter model (e.g. `openrouter/owl-alpha`) fails with a 404. The request is sent to the wrong endpoint:

```
endpoint: https://integrate.api.nvidia.com/v1/chat/completions
model:    openrouter/owl-alpha
auth:     NVIDIA_API_KEY
→ status: 404 Not Found  (404 page not found)
```

### Root cause

Selecting a built-in OpenAI-compatible profile (e.g. NVIDIA NIM via `nvidia-nim:...`) calls `force_apply_openai_compatible_profile_env(Some(profile))`, which stamps that profile's endpoint + API key into the **process-global** `JCODE_OPENROUTER_*` env vars. The `ActiveProvider::OpenRouter` arm of `set_model` never cleared those overrides when switching back, so a native OpenRouter model was POSTed to the stale profile endpoint with the wrong key.

Because the leak lives in process-global env, even brand-new sessions kept failing until the server was fully restarted. Other providers (Claude/OpenAI/Gemini/Copilot) are unaffected because they use self-contained providers and never touch this shared env.

## Fix

In the OpenRouter `set_model` arm: when the previous selection was a built-in direct profile (`profile_id.is_some()`) and the target is a native `openrouter.ai` catalog model (id starts with `openrouter/`), reset the profile env to `None` and rebuild the provider so it talks to the native endpoint again.

Deliberately left untouched:
- raw/custom endpoints configured directly via `JCODE_OPENROUTER_API_BASE` (`profile_id == None`)
- `@provider`-pinned or opaque model ids on forced-OpenRouter providers
- locked named profiles (`JCODE_PROVIDER_PROFILE_ACTIVE`)

## Tests

Adds `test_switch_back_to_native_openrouter_restores_endpoint_after_nvidia`, which reproduces the OpenRouter → NVIDIA → OpenRouter switch-back and asserts the endpoint override is cleared. It **fails without the fix** with the exact `integrate.api.nvidia.com` URL, and passes with it. Verified no regressions in the provider test suite (the only remaining failures are pre-existing parallel-env-contamination flakes present on `master` that pass in isolation).

### DIFF
--- a/crates/jcode-app-core/src/server/comm_session.rs
+++ b/crates/jcode-app-core/src/server/comm_session.rs
@@ -191,6 +191,71 @@ fn cleanup_prepared_visible_spawn_session(session_id: &str) {
     }
 }
 
+/// Maximum time to wait for a visibly-spawned session's interactive client to
+/// attach before concluding the launch did not actually produce a live worker.
+const VISIBLE_SPAWN_ATTACH_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(8);
+/// Poll interval while waiting for a client attachment.
+const VISIBLE_SPAWN_ATTACH_POLL: std::time::Duration = std::time::Duration::from_millis(200);
+
+/// Wait for a visibly-spawned session to register a live client attachment.
+///
+/// A successful terminal-launch fork does not guarantee that an interactive
+/// client connected (e.g. on a headless host or `jcode serve` shared server).
+/// We treat the appearance of at least one live event channel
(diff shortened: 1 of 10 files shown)


1jehuang:
Thanks for the contribution! This has since been fixed in master (see b1272ae1), so I'm closing this PR.

PR #283, opened 2026-05-29
feat(provider): support ANTHROPIC_BASE_URL for custom endpoint override

kaijianding:
Allows routing Anthropic API requests through env ANTHROPIC_BASE_URL (e.g. http://127.0.0.1:3456/v1).   
When set, the messages endpoint is derived from the env var instead of the hardcoded api.anthropic.com URL. OAuth vs non-OAuth path selection is preserved.


1jehuang:
Thanks! This repo no longer accepts external PRs, but the underlying problem is real - it's now tracked in #362 (with credit to this PR). Closing.

PR #296, opened 2026-06-01
Accept set_route as model switch alias

### LINKED ISSUE #295: Accept set_route compatibility alias for model switching
## Summary
Switching models from a client that emits the route-oriented request name `set_route` fails against the current server protocol.

## Error
```text
Invalid request: unknown variant `set_route`, expected one of ... `set_model` ... at line 1 column 19
```

## Impact
Model switching can fail, including when switching to Claude API models, if the UI/client sends the legacy or route-oriented `set_route` request shape.

## Expected behavior
The server should treat `set_route` as a compatibility alias for `set_model`, since both requests carry the active model selection.

## Proposed fix
Accept `set_route` as a serde alias for `Request::SetModel` while continuing to serialize outbound requests as `set_model`.

1jehuang:
## Summary
- accept legacy `set_route` requests as an alias for `set_model`
- keep serializing model switch requests as `set_model`
- add a regression test for `set_route` deserialization

Fixes #295

## Test plan
- `cargo test -p jcode-protocol test_set_route_deserializes_as_set_model_compat_alias`
- `rustfmt --check crates/jcode-protocol/src/wire.rs crates/jcode-protocol/src/protocol_tests/misc_events.rs`
- `selfdev build` / `scripts/dev_cargo.sh build --profile selfdev -p jcode --bin jcode`


1jehuang:
Superseded by master. The `set_route` compatibility alias (issue #295) is already implemented on master via `decode_request`, which normalizes a bare `{"type":"set_route","model":"..."}` request to `SetModel` while reserving structured `set_route` for the router. Verified with the passing tests `test_set_route_deserializes_as_set_model_compat_alias` and `test_structured_set_route_decodes_as_set_route_not_set_model` in crates/jcode-protocol. This PR's branch is also based on v0.26.0 and bundles unrelated commits (menubar, keybindings, sleep guard), so closing rather than rebasing.

PR #320, opened 2026-06-04
fix(provider): avoid clearing compatible profile on model set; profile-specific cache for reasoning models

sternelee:
- Prevent set_active_provider(OpenRouter) from shadowing active compatible profiles
- Use profile_id namespace for disk cache lookup to skip blocking API calls
- Respect explicit thinking_enabled=true even without provider features
- Add kimi-for-coding to Kimi model detection


1jehuang:
Thanks for the contribution! This has since been fixed in master (see 48a6017e), so I'm closing this PR.

PR #326, opened 2026-06-04
[codex] Support Anthropic-compatible base URLs

### LINKED ISSUE #325: Support Anthropic-compatible base URLs for direct Anthropic provider
### Problem

The direct Anthropic provider currently builds direct Messages and model-catalog requests against the official Anthropic API endpoints. That works for `ANTHROPIC_API_KEY`, but it prevents Anthropic-compatible gateways/proxies from being used even when a user provides Anthropic-style environment variables such as `ANTHROPIC_BASE_URL` and a bearer-style `ANTHROPIC_AUTH_TOKEN`.

This is separate from the Claude OAuth/subscription path: OAuth should keep using the official Anthropic OAuth endpoints because the preflight/bootstrap flow depends on them.

### Reproduction

Configure a local or enterprise Anthropic-compatible gateway:

```powershell
$env:JCODE_RUNTIME_PROVIDER = "anthropic-api"
$env:ANTHROPIC_BASE_URL = "http://127.0.0.1:15721/anthropic/v1"
$env:ANTHROPIC_AUTH_TOKEN = "<redacted>"
```

Then run a direct Anthropic request, for example:

```powershell
jcode --provider anthropic-api run --tool-profile none "Reply exactly OK"
```

Expected behavior:

- Messages requests go to `http://127.0.0.1:15721/anthropic/v1/messages`.
- Bearer-style gateway tokens are sent as `Authorization: Bearer <token>`.
- Official `ANTHROPIC_API_KEY` usage continues to send `x-api-key` to the default Anthropic API.
- Claude OAuth continues to use the official OAuth URL and headers.

Actual behavior:

- The direct Anthropic transport is tied to official Anthropic endpoints/auth behavior, so the gateway cannot be selected through the usual Anthropic-compatible env vars.

### Notes

A focused fix should keep OAuth unchanged, preserve the existing direct API-key path, and only apply custom base URL/bearer-token handling to non-OAuth direct Anthropic requests.

2233admin:
Closes #325

## Summary

- Honor `ANTHROPIC_BASE_URL` for non-OAuth direct Anthropic Messages requests and model-catalog refreshes.
- Add `ANTHROPIC_AUTH_TOKEN` support for bearer-token Anthropic-compatible gateways/proxies.
- Preserve existing behavior for Claude OAuth and official Anthropic API keys: OAuth stays on the official OAuth endpoint, and API keys keep using `x-api-key`.
- Update Anthropic auth-status detection and test env isolation so bearer-token direct auth is visible without polluting unrelated tests.

## Validation

- `cargo test -p jcode-base provider::anthropic --lib`
- `cargo test -p jcode-base provider::models::catalog --lib`
- `cargo test -p jcode-base auth::active_method --lib`
- `cargo test -p jcode-base full_and_fast_auth_status_match_for_shared_probe_fields --lib`
- `cargo check -p jcode-base --all-targets`
- `rustfmt --edition 2024 --config skip_children=true --check <touched Rust files>`

## Local notes

- `cargo fmt --all -- --check` reports unrelated formatting churn in untouched modules with my local stable Rust 1.96 checkout.
- `cargo clippy -p jcode-base --lib -- -D warnings` currently stops on pre-existing warnings outside this patch (`jcode-core/src/stdin_detect.rs`, `jcode-render-core/src/markdown.rs`).
- `cargo check -p jcode-base --all-targets --all-features` is not a clean Windows local signal here: it stops in `tikv-jemalloc-sys` before checking this patch because the Windows/MSVC all-features path expects a Unix `sh`/jemalloc configure environment.

### DIFF
--- a/crates/jcode-base/src/auth/active_method.rs
+++ b/crates/jcode-base/src/auth/active_method.rs
@@ -77,10 +77,12 @@ pub fn resolve_dual_credential_auth(
     let (has_oauth, has_api_key, forced) = match provider {
         ActiveProvider::Claude => {
             let has_oauth = auth.anthropic.has_oauth;
-            // `has_api_key` already folds in the ANTHROPIC_API_KEY env var via the
-            // auth probe, but re-check defensively so an env-only key set after the
+            // `has_api_key` already folds in Anthropic direct auth env vars via the
+            // auth probe, but re-check defensively so env-only auth set after the
             // cached snapshot still reports honestly.
-            let has_api_key = auth.anthropic.has_api_key || std::env::var("ANTHROPIC_API_KEY").is_ok();
+            let has_api_key = auth.anthropic.has_api_key
+                || std::env::var("ANTHROPIC_API_KEY").is_ok()
+                || std::env::var("ANTHROPIC_AUTH_TOKEN").is_ok();
             let forced = match runtime.as_deref() {
                 Some("claude-api" | "anthropic-api") => Some(ActiveCredential::ApiKey),
                 Some("claude" | "anthropic") => Some(ActiveCredential::OAuth),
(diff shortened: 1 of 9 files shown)


1jehuang:
Thanks! This repo no longer accepts external PRs, but the underlying problem is real - it's now tracked in #362 (with credit to this PR). Closing.

PR #331, opened 2026-06-05
fix: enable emulated TLS for aarch64-unknown-linux-gnu build

luciusrockwing:
-


luciusrockwing:
Hi @1jehuang 👋

I've tested the `-Z tls-model=emulated` approach locally on Termux (Android) and confirmed it fixes the `Cannot start a runtime from within a runtime` Tokio crash caused by broken native `__thread` TLS under Termux's glibc on the Android kernel.

This PR adds the fix to the CI workflow for the `aarch64-unknown-linux-gnu` target. Could you take a look and consider merging? Once merged, I can download the CI artifact and verify the binary works end-to-end on Termux.

I also filed a related issue upstream with the Termux glibc maintainers: termux-pacman/glibc-packages#374

Thanks!

PR #332, opened 2026-06-06
fix(tui): sync copy selection via OSC52

capt-marbles:
## Summary

Fixes copy-selection auto-copy behavior in browser-backed terminals such as cmux by emitting OSC 52 clipboard escape sequences in addition to existing host clipboard integrations.

## Why

The current `copy_to_clipboard` path can succeed via `wl-copy`/`arboard` while only updating the remote host clipboard. In cmux/web terminals, users expect drag auto-selection copy to update the browser/local clipboard. OSC 52 is the terminal-native clipboard synchronization mechanism supported by many terminal frontends.

## Changes

- Emit OSC 52 clipboard sequence whenever Jcode copies selected text.
- Preserve existing `wl-copy` and `arboard` behavior.
- Add tmux passthrough wrapping for OSC 52 sequences.
- Add unit tests for OSC 52 sequence formatting.

## Validation

- `cargo fmt`
- `cargo test clipboard_tests::osc52_clipboard_sequence --lib`
- `cargo test test_copy_selection_mouse_drag_auto_copies_and_exits_mode --lib`


1jehuang:

PR #335, opened 2026-06-06
fix(browser): use 'open -a Firefox' for XPI install on macOS

### LINKED ISSUE #334: fix(browser): XPI extension install fails on macOS with kLSApplicationNotFoundErr
## Summary

`jcode browser setup` fails on macOS when trying to install the Firefox extension. The `open` command is called on the `.xpi` file URL, but macOS has no registered handler for `.xpi` files, producing:

```
No application knows how to open URL file:///Users/udi/.jcode/browser/browser-agent-bridge.xpi
Error Domain=NSOSStatusErrorDomain Code=-10814 "kLSApplicationNotFoundErr"
```

## Root Cause

In `crates/jcode-base/src/browser.rs:804`, `install_extension()` uses:

```rust
let _ = tokio::process::Command::new("open").arg(&xpi_url).spawn();
```

macOS `open` without a specific application asks Launch Services for the default handler for `.xpi` — which does not exist.

## Fix

Use `open -a Firefox <file-url>` (with bundle-ID fallback `open -b org.mozilla.firefox`) so macOS opens the XPI file directly with Firefox, which knows how to handle Firefox extension installs.

## Environment

- macOS aarch64 (Apple Silicon)
- jcode v0.23.0+
- Firefox (any version)

## Reproduction

1. Run `jcode browser setup`
2. Observe the error when setup tries to open the XPI installer

## Expected

The XPI file opens in Firefox, prompting the user to install the extension.

## Actual

macOS Launch Services error because no application claims `.xpi` files.

udiedrichsen:
## Summary

Fixes `jcode browser setup` failing on macOS with `kLSApplicationNotFoundErr` when trying to open the Firefox extension (`.xpi`) installer.

## Problem

The previous code called `open <file-url>` on the `.xpi` file, but macOS has no default handler for `.xpi` files. This produces:

```
No application knows how to open URL file:///.../browser-agent-bridge.xpi
Error Domain=NSOSStatusErrorDomain Code=-10814 "kLSApplicationNotFoundErr"
```

## Fix

Use `open -a Firefox <file-url>` to explicitly open the XPI with Firefox, which knows how to handle Firefox extension installation. Falls back to `open -b org.mozilla.firefox` as a secondary option.

## Changes

| File | Change |
|------|--------|
| `crates/jcode-base/src/browser.rs` | Replace plain `open` with `open -a Firefox` in the macOS `#[cfg]` block of `install_extension()` |

## Validation

- Builds cleanly on macOS aarch64 (selfdev profile)
- The same `open -a Firefox` pattern is the standard recommended approach for opening files with a specific application on macOS

Closes #334

### DIFF
--- a/crates/jcode-base/src/browser.rs
+++ b/crates/jcode-base/src/browser.rs
@@ -802,7 +802,22 @@ async fn install_extension() -> Result<String> {
     }
     #[cfg(target_os = "macos")]
     {
-        let _ = tokio::process::Command::new("open").arg(&xpi_url).spawn();
+        // Try Firefox explicitly first; fall back to generic `open` so the user
+        // sees a clear error if Firefox is not installed.
+        let result = tokio::process::Command::new("open")
+            .args(["-a", "Firefox"])
+            .arg(&xpi_url)
+            .spawn();
+        match result {
+            Ok(_) => {}
+            Err(_) => {
+                // Fallback: try bundle ID, then plain `open`
+                let _ = tokio::process::Command::new("open")
+                    .args(["-b", "org.mozilla.firefox"])
+                    .arg(&xpi_url)
+                    .spawn();
+            }
+        }
     }
     #[cfg(target_os = "windows")]
     {


1jehuang:
Thanks! This repo no longer accepts external PRs, but the underlying problem is real - it's now tracked in #364 (with credit to this PR). Closing.

PR #342, opened 2026-06-08
Support custom base URL for OpenAI Responses API

### LINKED ISSUE #343: Allow native OpenAI Responses API to use custom base URL
## Problem

The native `openai-api` provider uses the OpenAI Responses API, but its base URL is currently hardcoded to `https://api.openai.com/v1`.

That means users with a local/proxied Responses API endpoint cannot configure `openai-api` to use it. The available `openai-compatible` configuration path is not equivalent because it targets the OpenAI-compatible chat/completions protocol rather than the native Responses API path.

## Reproduction

1. Run a local Responses API endpoint at `http://127.0.0.1:8317/v1`.
2. Configure `OPENAI_API_KEY` and try to set `OPENAI_BASE_URL`, `OPENAI_API_BASE`, or `JCODE_OPENAI_API_BASE`.
3. Run:

```bash
jcode -p openai-api -m gpt-5.5 run 'hello'
```

## Actual behavior

`openai-api` still sends requests to `https://api.openai.com/v1/responses`.

## Expected behavior

`openai-api` should continue using `/responses`, but allow a custom API base URL for API-key based usage.

## Notes

This should remain distinct from `openai-compatible`, which is a separate provider/protocol path.

chyax98:
Closes #343

## Problem

The native `openai-api` provider uses the OpenAI Responses API, but its base URL was hardcoded to `https://api.openai.com/v1`. Users with a local or proxied Responses API endpoint could only configure `openai-compatible`, which is the wrong protocol path because it targets OpenAI-compatible chat/completions rather than native `/responses`.

## Summary

- Allow native `openai-api` Responses API requests to use a custom API base URL.
- Read the custom base URL from `JCODE_OPENAI_API_BASE`, `OPENAI_API_BASE`, or `OPENAI_BASE_URL`, including values saved in `openai.env`.
- Support non-interactive setup via:

```bash
jcode login openai-api \
  --api-base http://127.0.0.1:8317/v1 \
  --api-key '<key>' \
  --model gpt-5.5
```

- Keep `openai-compatible` separate and unchanged.
- Keep ChatGPT/OAuth mode on the existing ChatGPT backend URL.

## Edge cases and tradeoffs

- The custom base URL only affects non-ChatGPT native OpenAI API-key mode. ChatGPT OAuth/subscription mode still uses `https://chatgpt.com/backend-api/codex`.
- URL validation reuses the existing `normalize_api_base` policy, so HTTPS is accepted and private/local HTTP endpoints are allowed, while unsafe public HTTP endpoints are rejected.
- Env precedence is `JCODE_OPENAI_API_BASE`, then `OPENAI_API_BASE`, then `OPENAI_BASE_URL`.

## Validation

- `cargo check -p jcode-base -p jcode --bin jcode`
- Built local debug binary and verified native `openai-api` against a local Responses endpoint:

```bash
/Users/xd/.jcode/source/jcode/target/debug/jcode \
  --socket /tmp/jcode-openai-base-test.sock \
  -p openai-api -m gpt-5.5 \
  run 'Reply exactly JCODE_RESPONSES_BASE_OK'
```

Result:

```text
JCODE_RESPONSES_BASE_OK
```

### DIFF
--- a/crates/jcode-base/src/provider/openai.rs
+++ b/crates/jcode-base/src/provider/openai.rs
@@ -29,6 +29,10 @@ const CHATGPT_API_BASE: &str = "https://chatgpt.com/backend-api/codex";
 const RESPONSES_PATH: &str = "responses";
 const DEFAULT_MODEL: &str = "gpt-5.5";
 const ORIGINATOR: &str = "codex_cli_rs";
+const OPENAI_API_BASE_ENV: &str = "OPENAI_API_BASE";
+const OPENAI_BASE_URL_ENV: &str = "OPENAI_BASE_URL";
+const JCODE_OPENAI_API_BASE_ENV: &str = "JCODE_OPENAI_API_BASE";
+const OPENAI_ENV_FILE: &str = "openai.env";
 
 /// Maximum number of retries for transient errors
 const MAX_RETRIES: u32 = 3;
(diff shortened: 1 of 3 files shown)


1jehuang:
Thanks for the contribution! This has since been fixed in master (see d8a19a3a), so I'm closing this PR.

Issue #338, opened 2026-06-08
OpenAI provider doesn't treat TLS BadRecordMac / fatal-alert as retryable (inconsistent with other providers)


1jehuang:
## Summary

The OpenAI/Codex streaming path uses its own retryability classifier `is_retryable_error` (`crates/jcode-base/src/provider/openai_stream_runtime.rs:1477`) which does **not** include TLS-layer transient errors like `BadRecordMac` / `received fatal alert`. Every other provider (Claude, Copilot, OpenRouter, Anthropic) delegates to the shared `is_transient_transport_error` (`crates/jcode-base/src/provider/routing.rs:5`), which **does** classify these:

```
// routing.rs (used by Claude/Copilot/OpenRouter/Anthropic)
|| lower.contains("badrecordmac")
|| lower.contains("bad_record_mac")
|| lower.contains("fatal alert: badrecordmac")
|| lower.contains("received fatal alert: badrecordmac")
|| lower.contains("decryption failed or bad record mac")
// ... also tls handshake eof, dns errors, no route to host, network unreachable, connection aborted
```

`openai_stream_runtime::is_retryable_error` is missing all of the above (no `badrecordmac`, no `fatal alert`, no `tls handshake eof`, no `connection aborted`, no DNS/route errors).

## Impact

On the OpenAI transport, a transient TLS blip (`Stream error: IO error: received fatal alert: BadRecordMac`) surfaces straight to the user and hard-fails the turn, with no retry. The exact same error on any other provider would be retried transparently. Reported by a user on v23.

## Proposed fix

Have the OpenAI path delegate to (or include the cases from) the shared `is_transient_transport_error` so retry behavior is unified across all providers. Add a regression test asserting `BadRecordMac` / `received fatal alert` / `tls handshake eof` are classified retryable on the OpenAI path.

## Notes

This is the correct fix regardless of any single user's root cause. It does **not** necessarily resolve a user whose underlying network path is corrupting TLS records (a proxy/VPN/hardware issue); in that case retries may still exhaust. It only ensures jcode treats these transient TLS/network errors consistently and gives them a chance to recover.

## References
- `crates/jcode-base/src/provider/openai_stream_runtime.rs:1477` (`is_retryable_error`)
- `crates/jcode-base/src/provider/routing.rs:5` (`is_transient_transport_error`)
- `crates/jcode-base/src/provider/openai_provider_impl.rs:468` (retry decision in the OpenAI stream loop)

[...]

PR #347, opened 2026-06-09
fix(providers): retry on TLS transport faults (BadRecordMac) across all providers

1jehuang:
Addresses #338 (gap #1: OpenAI TLS retryability) plus a related root cause (anyhow `.context()` masking). See https://github.com/1jehuang/jcode/issues/338#issuecomment-4656256966.

## What

Transient TLS transport faults (`received fatal alert: BadRecordMac`, broken pipe, connection reset, DNS, HTTP/2 stream errors) on the **OpenAI** path failed immediately instead of retrying, and the `claude`/`copilot`/`openrouter`/`openai` retry classifiers could be blinded to such causes when wrapped behind `anyhow`'s `.context(...)`.

## Changes

1. **OpenAI** `is_retryable_error` now delegates to the shared `is_transient_transport_error` (instead of a separate allowlist that omitted all TLS terms).
2. **claude / copilot / openrouter / openai** classify on `format!("{e:#}")` (full anyhow source chain) so a `.context(...)`-wrapped transport cause is visible. (anthropic already did this.)
3. Regression tests for the TLS terms and the masked-context case; existing negative auth tests still pass.
4. `scripts/repro/tls-bad-record-mac/`: standalone, deterministic fault-injection reproduction (corrupting TLS proxy -> real `bad_record_mac` alert). 15/15 stable.

## Scope

Covers gap **#1** in #338 (and the masking sub-bug). Intentionally does NOT include the other gaps tracked there: #2 (429 retry on OpenAI), #3 (no-output-yet guard before mid-stream retry), #4 (jitter), #5 (MAX_RETRIES). #3 deserves its own careful PR.

## Validation
- `cargo test -p jcode-base --lib` classifier tests: all pass (incl. negatives).
- `cargo build -p jcode-base`: clean.
- Repro harness: `SUCCESS: realistic BadRecordMac reproduced and both fixes validated.`

### DIFF
--- a/crates/jcode-base/src/provider/claude.rs
+++ b/crates/jcode-base/src/provider/claude.rs
@@ -729,7 +729,11 @@ impl Provider for ClaudeProvider {
                 {
                     Ok(()) => return, // Success
                     Err(e) => {
-                        let error_str = e.to_string().to_lowercase();
+                        // Use the full anyhow source chain ({:#}) so transport
+                        // causes wrapped behind a `.context(...)` (e.g. a TLS
+                        // `received fatal alert: BadRecordMac`) are visible to the
+                        // retry classifier, not just the top-level context string.
+                        let error_str = format!("{e:#}").to_lowercase();
                         // Check if this is a transient/retryable error
                         if is_retryable_error(&error_str) && attempt + 1 < MAX_RETRIES {
                             crate::logging::info(&format!("Transient error, will retry: {}", e));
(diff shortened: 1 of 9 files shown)


1jehuang:
All of this PR's content is now on master:

- The classifier unification (OpenAI -> shared `is_transient_transport_error` + 429) landed earlier in 7fb1f52b.
- The remaining half (the anyhow `.context()` masking fix: classify on `format!("{e:#}")` in claude/copilot/openrouter/openai, plus the regression test and the `scripts/repro/tls-bad-record-mac` fault-injection harness) landed directly on master in 2540f88e.

Verified on master: `cargo test -p jcode-base --lib provider::` passes (403 tests), and the standalone repro runs green end-to-end (genuine `bad_record_mac` alert reproduced; old logic misses it, new logic retries).

Closing since master now supersedes this branch.

Issue #348, opened 2026-06-09
computer tool: truthful observability animation for background AX actions (highlight when visible, notice when occluded)


1jehuang:
## Summary

The `macos_computer_use` tool (added in #345, gated to macOS) can act on the
desktop two ways:

- **Background AX / scripting** (`press`, `set_value`, `select_menu`,
  `run_applescript`, ...): acts on a UI element *by reference*. No cursor
  moves, the target app need not be focused or even frontmost. This is the
  preferred, non-disruptive path.
- **Visible CGEvent input** (`move`, `click`, `type`, `drag`, `scroll`):
  drives the one real system cursor on the shared HID stream, so the user
  sees it.

The background path is great for staying out of the user's way, but it has a
trust cost: **silent actions on the user's live machine with no signal that
anything happened.** We should add a lightweight, *truthful* observability
indicator so the user knows what the agent is doing.

This is observability/trust work, not decoration. It does **not** replace the
policy text or any future confirmation gate; a flash is a notice, not consent.

## Proposed behavior

Decide what to show based on **visibility/occlusion, not focus** (the two are
orthogonal on macOS: a window can be unfocused but fully visible).

1. **Target's window is actually visible on screen** (focused or not, and
   topmost at the element's rect): draw a brief, click-through,
   non-activating highlight/flash on the element's on-screen bounds. We
   already resolve the element's `@pos` in `find_element`, so we have the
   rectangle. ~300ms fade.
2. **Target is occluded / minimized / hidden / on another Space or display**:
   do **not** draw a positional highlight, it would float over an unrelated
   window the user is actually looking at, which is misleading. Fall back to
   a **non-positional notice** (toast / activity-log line), e.g. "agent set
   value in TextEdit (background)".
3. **Visible CGEvent actions** already move the real cursor, so the user sees
   them. A small click ripple/trail is optional and is about legibility (the
   cursor teleports instantly, which is jarring), not about revealing hidden
   activity.

Deciding signal = occlusion / on-screen hit-test (element AX frame vs. the
window list to confirm the target isn't covered), **not** whether the window
has focus.

## Explicitly out of scope / anti-goals

- **No fake "ghost cursor."** Animating a pointer gliding around for
  background actions implies a cursor moved when nothing did. Don't fake
  motion that didn't happen.
- **No focus stealing, no fullscreen overlay, no added latency** to the
  action itself. Any overlay must be click-through and non-activating
  ("stay out of the user's way", consistent with the restraint guidance
[...]

PR #351, opened 2026-06-09
fix(computer): make every action invocable (schema field gaps) + element_at/error-mapping fixes

### LINKED ISSUE #350: computer tool: schema omits half its input fields (scroll/drag/resize/select_menu/window_screenshot/wait_for/set_brightness/perform_action unusable) + element_at AppleScript reserved-word bug
## Summary

Exhaustively exercising every `computer` tool action surfaced a class of bugs that make a large fraction of the documented actions **impossible to invoke** from the model, plus one AppleScript bug. The handlers are implemented correctly; the problem is the **function-calling contract** (`parameters_schema()`) and one reserved-word collision.

Branch: `feat/macos-computer-tool` (running binary `0.24.3-computer`).

## Root cause 1 — schema omits half the input fields (blocks 8 actions)

`ComputerInput` deserializes many fields (`dx, dy, to_x, to_y, w, h, menu_path, window_id, ax_action, contains, timeout_ms, region, level`), and `dispatch()` requires them. But `parameters_schema()` uses "progressive disclosure" and only *declares* a subset of properties. Models emit only parameters present in the tool's JSON schema, so these fields can never be sent. Every dependent action dead-ends in its own `.context("... requires X")` guard:

| action | required field (undeclared) | observed error |
|---|---|---|
| `scroll` | `dx`/`dy` | `action='scroll' requires non-zero dx and/or dy` |
| `drag` | `to_x`/`to_y` | `action='drag' requires to_x and to_y` |
| `resize_window` | `w`/`h` | `resize_window requires w` |
| `select_menu` | `menu_path` | `select_menu requires menu_path` |
| `window_screenshot` | `window_id` | `window_screenshot requires window_id` |
| `wait_for` | `contains` | `wait_for requires contains` |
| `set_brightness` | `level` | `set_brightness requires level (0..1)` |
| `perform_action` | `ax_action` | (unreachable) |
| `ocr` (region) | `region` | (region variant unreachable; full-screen ok) |

Progressive disclosure of *action specs* (the `discover` text) is fine, but **input fields must be declared** or the action is unusable.

## Root cause 2 — `element_at` uses AppleScript reserved word `result`

`ax.rs::element_at` assigns to a handler variable named `result`. `result` is reserved in AppleScript (holds the last command's value), so the script fails:

```
Deepest element at (1100,400) in System Settings:
(error: The variable result is not defined.)
```

## Exhaustive test matrix (before fix)

Working: `check_permissions, discover, screenshot, ocr (full screen), ui, cursor, move, click, double_click, right_click, type, key, key_down, key_up, find_element, list_apps, list_windows, activate_app, hide_app(n/a), focus_window, move_window, get_clipboard, set_clipboard, run_applescript, run_jxa, notify, system_state`.

Broken: `scroll, drag, resize_window, select_menu, window_screenshot, wait_for, set_brightness, perform_action, element_at` (+ `ocr` region variant).

## Fix

1. Declare all input fields in `parameters_schema()` (keep action-spec text in `discover`). Bump the `schema_is_compact` bound to match the slightly larger but still small schema.
2. Rename the `element_at` AppleScript local `result` -> `bestHit`.

PR incoming.

1jehuang:
## Summary

Exhaustively exercised **every** `computer` tool action against the live running server and fixed every failure so the tool is production-ready. The handlers were already implemented; the bugs were in the function-calling contract and a few AppleScript/error-mapping/perf details. All fixes are validated live after build + reload.

Fixes #350.

### 1. Schema omitted half the input fields → 8+ actions were impossible to call
`ComputerInput` deserializes `dx, dy, to_x, to_y, w, h, menu_path, window_id, ax_action, contains, timeout_ms, region, level` and `dispatch()` requires them, but `parameters_schema()` only *declared* a subset. A model can only emit parameters present in the tool schema, so those fields could never be sent and each dependent action dead-ended in its own `requires X` guard:
`scroll, drag, resize_window, select_menu, window_screenshot, wait_for, set_brightness, perform_action`, plus the `ocr` region variant.

Fix: declare all input fields (action *specs* stay in `discover`; *fields* must be declared). Trimmed descriptions so the always-on schema stays ~880 tokens. Added `schema_declares_every_input_field` regression test.

### 2. `element_at` used the AppleScript reserved word `result`
`result` holds the last command's value, so `element_at` always failed with `The variable result is not defined.` Renamed to `bestHit`.

### 3. `element_at` walked the entire AX tree → 20s timeout on big apps
Now prunes to only descend subtrees whose frame contains the point (a child's frame ⊆ its parent's), with a depth bound + explicit 12s timeout. Full-tree walk → path-length walk. Live: System Settings used to time out (20s); a comparable large tree (Messages) now returns in ~1.6s.

### 4. `-1719` misclassified as "Accessibility permission required"
`-1719` (errAEIllegalIndex) / `-1728` (errAENoSuchObject) mean the target reference didn't resolve (app not running / no front window / AX path out of range), not a permission problem. Extracted a pure, unit-tested `classify_osa_error`: permission text first, then map index errors to an accurate "target not found" message. Use `-25211` (errAXAPIDisabled) for the real permission code.

### 5. `wait_for` ignored its own `timeout_ms`
Each poll dumped the AX tree (depth 10) under the 20s default scripting timeout, so a single poll could take ~20s and a `timeout_ms=2000` call wouldn't return until ~20s. Now each poll is bounded by the time remaining (capped 3s) and the dump depth is reduced. Live: a 2000ms `wait_for` now returns in ~2.2s.

## Validation

- `cargo test -p jcode-app-core --lib tool::computer`: **26 passed, 0 failed** (10 new tests across the fixes).
- Built `selfdev`, installed, reloaded the running server through 4 iterations, and re-ran the full action matrix **live**. Every previously-broken action now works:
  `scroll` ✓ `drag` ✓ `resize_window` ✓ `select_menu` ✓ (View > Actual Size in Preview) `window_screenshot` ✓ (captured an occluded window) `wait_for` ✓ (honors timeout) `set_brightness` ✓ (clean message) `perform_action` ✓ (AXPress) `ocr region` ✓ `element_at` ✓ (fast + accurate not-found errors).

## Notes / known limitations
- `wait_for`/`ui`/`element_at` operate on `front window of frontApp`; apps on another Space or with non-AX content (Catalyst/Electron/rich-text) expose limited text. This is an inherent AX limitation, not a tool bug. Follow-up tracked in #348.

## Test plan for reviewers
```
cargo test -p jcode-app-core --lib tool::computer
# live (mutates desktop; needs GUI + Accessibility/Screen Recording):
cargo test -p jcode-app-core tool::computer::coverage -- --ignored --nocapture --test-threads=1
```


JadenSWang:
use openrelay

1jehuang:
Closing as superseded. The intended computer-tool and keymap fixes through commit 0fde2dc10 are already ancestors of master, including the schema coverage, element lookup, timeout, and AppleScript error-mapping changes. The remaining commits on this reused branch are unrelated ScrollWM/onboarding work and should be proposed in a separate, scoped PR if still desired, rather than merged through this stale feature-branch PR.

Issue #277, opened 2026-05-29
jcode badly broken on Omarchy


capt-marbles:
Just installed jcode 0.14.6 on Omarchy Linux, and its unusable.

Anytime I invoke jcode in a terminal such as Alacritty or Ghostty, the screen flickers as it madly switches through different sessions looking for some kind of server handoff which doesn't come.

Can't use it until a fix or workaround is issued.

1jehuang:
Could you use an agent to diagnose more? There are logs that you could point it at to figure out what's going on, this is difficult for me to reproduce

capt-marbles:
Here you go

[jcode-2026-05-28.log.zip](https://github.com/user-attachments/files/28391780/jcode-2026-05-28.log.zip)

capt-marbles:
I dug into the copied Omarchy log and found the proximate failure mode: it is a server auto-reload loop, not just terminal/title rendering.

Log evidence from `jcode-2026-05-28.log`:

- `request_kind=reload`: 8,368 occurrences
- `send_reload_signal`: 2,789
- `Server: exec'ing into stable binary`: 2,788
- `Remote reconnect attempt`: 5,911
- `ack_write_failed`: 2,952
- `peer closed`: 2,925

Observed loop:

1. TUI client subscribes/resumes a session.
2. History response causes the client to think `server_has_update=true`.
3. Client auto-sends `request_kind=reload`.
4. Server execs into `/home/awalker/.jcode/builds/stable/jcode`.
5. Client disconnects, waits for reload handoff, reconnects.
6. The new History response again reports update available, so reload repeats.

Representative log lines:

```text
request_kind=reload
send_reload_signal ...
Server: reload signal received
Server: exec'ing into stable binary "/home/awalker/.jcode/builds/stable/jcode"
handle_disconnect ... detail=server reload in progress
wait_for_reload_handoff_event
Remote reconnect attempt
Reloading with binary built 39 minutes ago...
```

I tested a local patch on Omarchy that makes `server_has_newer_binary()` conservative: only report an update when the candidate binary is a different path and its mtime is provably newer than the currently running executable. In particular, it removes the fallback where an mtime comparison failure plus a different canonical path reports an update, because that false positive can trigger this infinite reload loop.

Patch is essentially in `src/server/util.rs`:

```rust
let Some(current_time) = current_mtime else {
    crate::logging::warn(&format!(
[...]

PR #355, opened 2026-06-11
Fix server update-reload loop: resolve launcher wrapper to its .bin in mtime checks

### LINKED ISSUE #356: Server update-reload loop: server_has_newer_binary() false-positive because installer writes wrapper script after extracting the binary
## Summary

After every auto-update, all clients get stuck in an endless reload loop:

```
ℹ Connected server binary differs from the installed client channel. Reloading the server before applying
remote session state. Set JCODE_ALLOW_SERVER_VERSION_MISMATCH=1 only for intentional compatibility testing.

ℹ Newer server binary detected. Auto-reload is disabled by display.auto_server_reload = false. Use /reload
manually when you're ready.
```

These two messages repeat forever (~every 17s), `/run/user/1000/jcode.reload` keeps being rewritten, and clients defer remote session state indefinitely — even though the server is **already running the exact binary of the latest installed version**.

Reproduced on v0.25.0 and v0.25.1 (Linux x86_64, WSL2). Happens on essentially every update.

## Root cause

The installer extracts `jcode-linux-x86_64.bin` first, then writes the 505-byte sh wrapper `~/.jcode/builds/versions/<ver>/jcode` a few milliseconds later:

```
-rwxr-xr-x 1 user user       505 2026-06-11 09:59:02.914492663 jcode
-rwxr-xr-x 1 user user 118385296 2026-06-11 09:59:02.912582537 jcode-linux-x86_64.bin
```

`server_has_newer_binary()` (crates/jcode-app-core/src/server/util.rs) compares the mtime of the reload candidate — which canonicalizes to the **wrapper**, not the `.bin` — against the running `.bin`. Since the wrapper is always written last, it is always strictly newer, so `server_has_update = true` forever. Every client then hits `should_defer_history_for_runtime_identity` (crates/jcode-tui/.../remote/server_events.rs) and forces a reload that changes nothing, and the cycle repeats.

Reinstalling does not help — any install recreates the wrapper with a newer mtime than the binary.

## Workaround

```sh
touch -r ~/.jcode/builds/versions/<ver>/jcode-linux-x86_64.bin ~/.jcode/builds/versions/<ver>/jcode
rm /run/user/1000/jcode.reload
```

Takes effect on next client attach, no server restart needed. But it has to be re-applied after every update.

## Fix

PR #355 fixes this in two independent layers: update detection resolves the wrapper to the `.bin` it execs before comparing mtimes/identity, and the update installer stamps the wrapper with the `.bin`'s mtime.

vietgs03:
## Summary

After every auto-update on Linux, all clients got stuck in an endless reload loop (~every 17s):

```
ℹ Connected server binary differs from the installed client channel. Reloading the server before applying remote session state...
ℹ Newer server binary detected. Auto-reload is disabled by display.auto_server_reload = false. Use /reload manually when you're ready.
```

`/run/user/1000/jcode.reload` kept being rewritten and clients deferred remote session state forever — even though the server was already running the exact binary of the latest installed version. Reproduced on v0.25.0 and v0.25.1 (Linux x86_64, WSL2); happens on essentially every update.

## Root cause

The update installer copies extracted files in `read_dir` order and `fs::copy` stamps each destination with its copy time, so the 505-byte sh wrapper `versions/<ver>/jcode` lands a few ms **after** `jcode-linux-x86_64.bin`:

```
-rwxr-xr-x 1 user user       505 09:59:02.914492663 jcode
-rwxr-xr-x 1 user user 118385296 09:59:02.912582537 jcode-linux-x86_64.bin
```

`server_has_newer_binary()` canonicalizes the channel symlink to the **wrapper** and compares its mtime against the running **`.bin`** → the wrapper is always strictly newer → `server_has_update = true` forever. Every client then hits `should_defer_history_for_runtime_identity` and forces a reload that changes nothing (the issue #277 loop family). Reinstalling doesn't help — any install recreates the skew.

## Fix (two independent layers)

1. **Detection** (`server/util.rs`): new `launcher_wrapper_target()` / `comparison_path()` resolve a wrapper script to the `.bin` it execs — parse the `exec "$self_dir/<name>.bin"` line, fall back to a unique `jcode*.bin` sibling, refuse ambiguity. `server_has_newer_binary`, `newest_reload_candidate` and the downgrade guard in `reload_exec_target` now compare real binary mtimes/identity. **Exec targets still use the wrapper** so `LD_LIBRARY_PATH` setup for bundled libs is preserved. Every failure path degrades to the previous behavior. This layer also self-heals installs that are already broken on disk.
2. **Installer** (`update.rs`): after copying the version dir, stamp the wrapper with the `.bin`'s mtime (best-effort, logged on failure) so install order can't fake a pending update.

`scripts/install.sh` is unaffected (it `mv`s the wrapper, preserving the archive mtime).

## Notes for review

- `reload_exec_target`'s same-binary fast-path still compares wrapper vs `.bin` unresolved; it falls through to the (now correct) mtime compare with the same outcome — left untouched to keep the diff narrow.
- `preferred_reload_candidate` (build-support) also reads wrapper mtimes, but only affects self-dev rebuild flows where a few ms of skew can't change the decision — out of scope here.

## Tests

- 5 unit tests for `launcher_wrapper_target` (parse, fallback, ELF / no-sibling / ambiguous refusals)
- 2 integration tests modeling a wrapped install end-to-end: no phantom update against itself; a genuine update is still detected and reload targets the new version
- All 33 `server::util` tests pass; `cargo check` + `rustfmt` clean

Fixes #356

### DIFF
--- a/crates/jcode-app-core/src/server/util.rs
+++ b/crates/jcode-app-core/src/server/util.rs
@@ -96,7 +96,9 @@ pub(crate) fn reload_exec_target(is_selfdev_session: bool) -> Option<(PathBuf, &
         .as_ref()
         .map(|p| p.as_path())
         .and_then(binary_mtime);
-    let candidate_mtime = binary_mtime(candidate_canonical.as_path());
+    // A wrapper's own mtime says nothing about the code it runs; compare the
+    // `.bin` behind it so the downgrade guard sees real binary ages.
+    let candidate_mtime = binary_mtime(comparison_path(candidate_canonical.clone()).as_path());
 
     match guarded_reload_target(
         candidate.clone(),
@@ -156,7 +158,9 @@ fn newest_reload_candidate(is_selfdev_session: bool) -> Option<(PathBuf, &'stati
     ];
     let with_mtimes = ordered.into_iter().flatten().map(|candidate| {
         let canonical = canonicalize_or(candidate.0.clone());
-        let mtime = binary_mtime(canonical.as_path());
+        // Freshness of a launcher wrapper is the freshness of the `.bin` it
+        // exec's, not of the wrapper file written moments later.
+        let mtime = binary_mtime(comparison_path(canonical.clone()).as_path());
         (candidate, canonical, mtime)
     });
     pick_newest_candidate(with_mtimes)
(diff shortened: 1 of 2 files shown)


vietgs03:
Screenshot 2026-06-11 104816
The loop i facing

1jehuang:
Thanks a lot for the excellent diagnosis and the PR - the analysis was spot on. An equivalent fix landed on master independently as 8cc66bc3 (shipped in v0.26.0) covering the same two layers: wrapper->payload resolution in all mtime/identity comparisons plus install-time mtime normalization, with regression tests. Closing as superseded; the root-cause writeup in #356 was genuinely helpful for confirming coverage.

PR #360, opened 2026-06-11
Fix: strip profile prefix in OpenRouterProvider::set_model for direct…

akvarel:
… profiles

When a switch spec like 'deepseek:deepseek-v4-flash' (produced by model_switch_request_for_session_route or fork_model_switch_request) is passed directly to OpenRouterProvider::set_model on a direct OpenAI-compatible profile (supports_provider_features=false), it was stored as-is, causing the API request to send 'model: deepseek:deepseek-v4-flash' which the DeepSeek API rejects.

Now detect and strip the known {profile_id}: prefix when the OpenRouterProvider has a profile_id set, so the bare model name is used in API requests.


1jehuang:
Thanks! This repo no longer accepts external PRs, but the underlying problem is real - it's now tracked in #363 (with credit to this PR). Closing.

PR #490, opened 2026-07-17
Make Jcode available in every Windows terminal by default

### LINKED ISSUE #488: Improve Windows setup lifecycle, terminal scrolling, and onboarding simulation
## Problem

The Windows experience needs a reliable end-to-end path from installation through first-run validation:

- install, PATH updates, upgrade, rollback, and uninstall must be idempotent and recoverable
- optional terminal and global-hotkey setup must require explicit consent
- mouse-wheel input in Windows terminals must scroll the transcript rather than recall prompt history
- developers need a safe way to replay onboarding from a pristine synthetic state without changing real auth or configuration

## Acceptance criteria

- isolated PowerShell tests cover install, upgrade, PATH deduplication, environment broadcast, rollback, and uninstall
- downloaded release assets are checksum and version validated, with source builds explicitly opt-in
- Windows VT mouse modes are enabled and disabled symmetrically with regression coverage
- Alt+5 resets and opens the onboarding simulator at its first screen while Ctrl+Alt/AltGr does not trigger it
- CI passes on Windows, Linux, and macOS

1jehuang:
## Summary
- persist %LOCALAPPDATA%\jcode\bin in the Windows user PATH
- update the current installer process and broadcast WM_SETTINGCHANGE so new CMD, PowerShell, Git Bash, and Alacritty windows resolve jcode`n- deduplicate stale PATH entries and cleanly remove them during uninstall
- add install, upgrade, PATH, broadcast, rollback, and uninstall regression tests

## Validation
- powershell -NoProfile -ExecutionPolicy Bypass -File scripts/test_windows_launcher_install.ps1`n- cargo fmt --all -- --check`n- verified jcode --version from PowerShell, CMD, Git Bash, and Alacritty-compatible environments

Closes #488


1jehuang:
Superseded by #487, which merged this launcher lifecycle patch and the subsequent Windows safety, uninstall, checksum, hotkey, and CI fixes.

The primary patch in this PR is patch-equivalent to the launcher commit already included in #487. Rebasing this older branch now would also discard later reviewed corrections, so it should not be merged separately.

PR #586, opened 2026-07-24
fix(model-list): keep config-declared models after background catalog refresh (fixes #579)

### LINKED ISSUE #579: fix(model-list): custom models from config.toml disappear after background catalog refresh
## Problem

User-defined custom models declared in `~/.jcode/config.toml` under `[[providers.<name>.models]]` disappear from the model picker after some time. After running `/refresh-model-list` the models appear correctly, but after ~15 minutes they vanish from the `/model` menu.

### Root cause

`should_merge_static_models_with_live_catalog()` in `crates/jcode-provider-openrouter-runtime/src/lib.rs` returned `false` for user-defined named provider profiles from config.toml. The condition `self.supports_provider_features || self.profile_id.is_none()` failed because:

- Named profiles have `supports_provider_features = false` (unless explicitly configured with `provider_routing` or `allow_provider_pinning`)
- Named profiles have `profile_id = Some(...)`

After 15 minutes (`MODEL_CATALOG_SOFT_REFRESH_SECS`), a background `/models` catalog refresh fired, published `ModelsUpdated`, invalidated the picker cache, and the next picker open showed only the live catalog models. The user's explicitly declared `[[providers.<name>.models]]` entries were replaced by the live catalog.

Built-in OpenAI-compatible profiles (Cerebras, NVIDIA NIM) and user-defined named profiles both set `profile_id = Some(...)`, so there was no way to distinguish them and apply different merge behavior.

### Fix

Added `is_user_named_profile: bool` field to `OpenRouterProvider`:
- **`true`** in `new_named_openai_compatible` (the config.toml path, `[providers.<name>]`)
- **`false`** in all other constructors (built-in profiles, standard OpenRouter, fork)

Updated `should_merge_static_models_with_live_catalog` to also return `true` when `is_user_named_profile` is `true`. Now config-declared models always merge with the live catalog and survive background refreshes. Built-in profiles keep their existing behavior (static fallbacks replaced by live catalog after fetch).

### Test

New regression test `named_profile_static_models_survive_live_catalog_refresh` populates the models cache with a live catalog that does NOT include the user's static model, then asserts the static model is still present in `available_models_display()`. All 93 existing openrouter tests pass with no regressions.

### Branch

Fix available at: https://github.com/alecuba16/jcode/tree/fix/model_list

Commit: `491f3fb1` — fix: preserve config-declared models after background catalog refresh

Rebased onto upstream master (`78fcc5e8`). No upstream commits fix this issue or touch the same files.

**PR creation blocked by OAuth token permission limitations.** A maintainer with push access can create the PR:
```
base: master  ←  head: alecuba16:fix/model_list
```

1jehuang:
Fixes #579.

## Problem
Models declared in `~/.jcode/config.toml` under `[[providers.<name>.models]]` disappeared from `/model` after ~15 minutes. `should_merge_static_models_with_live_catalog()` returned false for user-declared named profiles (they have `profile_id = Some(..)` but `supports_provider_features = false`), so the background `/models` refresh replaced the user's explicit list with the live catalog.

## Fix
Add `is_user_named_profile()`: a profile id that is not a built-in OpenAI-compatible profile (Cerebras, NVIDIA NIM, ...) came from a user `[providers.<name>]` block, so its static models keep merging with the live catalog. Built-in profiles keep the existing replace-after-fetch behavior, which is what avoids advertising stale preview models.

This is a smaller variant of @alecuba16's proposed patch: no new struct field or constructor changes, so no constructor can silently forget to set the flag.

## Test
New regression test `named_profile_static_models_survive_live_catalog_refresh` seeds the models cache with a live catalog lacking the user's model and asserts both models are shown. `cargo test -p jcode-provider-openrouter-runtime`: 94 passed, 0 failed.


1jehuang:
Validation note on how this was verified, since CI's `Build & Test` compiles lib tests with `--no-run` and only *executes* the retention-readiness, provider_matrix, and e2e suites:

- The three new regression tests are compile-verified by CI and **run locally**: `cargo test -p jcode-provider-openrouter-runtime` = 96 passed, 0 failed.
- Each was **revert-proven**: with `|| self.is_user_named_profile()` removed, all three fail; restoring it makes them pass. The built-in-profile test stays green in both directions, confirming Cerebras/NIM replace-after-fetch behavior is untouched.
- The end-to-end test parses a real `config.toml` profile snippet, performs an actual HTTP `/models` fetch whose catalog omits the declared model, and asserts the picker still offers it. Its declared model deliberately differs from `default_model` so the current-model fallback cannot mask a regression (an earlier draft passed for that wrong reason and was fixed).

Separately worth a maintainer look: unit tests across the workspace are compiled but never executed in CI, so unit-level regressions can only be caught locally. Not changing that here.

Issue #637, opened 2026-07-28
Cursor provider hardcodes agentn.global and fails for regional agent hosts


xek:
## Summary

`--provider cursor` fails with a misleading Cursor error:

```text
This region is not yet available for your team
```

even when the same machine/IP can use Cursor IDE Agent and the official `cursor-agent` CLI successfully.

## Root cause

jcode's native Cursor transport hardcodes:

```text
agentn.global.api5.cursor.sh
```

(`crates/jcode-provider-cursor-runtime/src/agent_transport.rs`)

`cursor-agent` does **not** use that host blindly. It bootstraps `aiserver.v1.ServerConfigService/GetServerConfig` and caches the regional endpoint in:

```text
~/.cursor/cli-config.json
  → serverConfigCache.agentUrlConfig.agentnUrl
```

On an affected account that value is e.g.:

```text
https://agentn.us.api5.cursor.sh
```

Hitting `global` with a team routed to `us` produces the "region is not yet available for your team" error. Auth method (API key vs `state.vscdb` token) does not matter; the host does.

## Reproduction

1. Have a working `cursor-agent` login (`cursor-agent status` shows logged in).
2. Confirm `~/.cursor/cli-config.json` has a non-global `agentnUrl` (e.g. `agentn.us.api5.cursor.sh`).
3. Run:

```bash
jcode --provider cursor run 'Reply with exactly: JCODE_CURSOR_OK'
```

### Actual

Request goes to `agentn.global.api5.cursor.sh` and fails with `unauthenticated` / "This region is not yet available for your team".

### Expected

jcode uses the same regional agent host as `cursor-agent` and completes the turn.

### Control

```bash
JCODE_CURSOR_AGENT_HOST=agentn.us.api5.cursor.sh \
  jcode --provider cursor run 'Reply with exactly: JCODE_CURSOR_OK'
# → JCODE_CURSOR_OK
```

Official CLI on the same IP also works:

```bash
cursor-agent -f -p 'Reply with exactly: CURSOR_AGENT_OK'
# → CURSOR_AGENT_OK
```

## Suggested fix

Resolve the agent host in this order (or equivalent):

1. Explicit config/env (e.g. `CURSOR_AGENT_HOST` in `~/.config/jcode/cursor.env`, matching how `CURSOR_API_KEY` is stored; and/or `JCODE_CURSOR_AGENT_HOST`)
[...]

Issue #641, opened 2026-07-28
[bug] Copilot ignores valid apps.json token: `github.com:<client-id>` key fails normalization


almai:
## Summary

`normalize_github_host_key()` strips a URL scheme and splits the host on `/`, but never on `:`. GitHub's `apps.json` keys carry a client-id suffix (`github.com:Iv1.b507a08c87ecfe98`), so the comparison against `"github.com"` fails, the function returns `None`, and `select_preferred_token()` drops the entry via `?`.

```rust
// crates/jcode-base/src/auth/copilot.rs:596
let host = host.split('/').next().unwrap_or_default().trim();   // never splits on ':'
let host = host.to_ascii_lowercase();

if host == "github.com" || host == "api.github.com" || host.ends_with(".github.com") {
```

```rust
// crates/jcode-base/src/auth/copilot.rs:553
let normalized_host = normalize_github_host_key(host)?;   // valid token silently dropped
```

This matters more now that GitHub writes credentials to `apps.json` rather than `hosts.json`.

## Why it's more than cosmetic

Per the `load_github_token()` doc comment, `apps.json` is source **6**, ahead of the trusted OpenCode/pi `auth.json` entries at source **7**. When `apps.json` is discarded, jcode continues down the chain and can pick up a **stale** token from `~/.pi/agent/auth.json`.

In my case that token had been revoked — GitHub returns 401 `Bad credentials` on the exchange — and the resulting failure surfaces to the user as:

```
The requested model is not supported.  code: model_not_supported
```

A model-shaped error for an auth-shaped problem. `jcode auth-test --provider copilot` reported `not_configured` the whole time, while a perfectly valid token sat unread in `apps.json`.

This is made considerably harder to diagnose by #640, which emits a byte-identical `model_not_supported` message from an entirely unrelated cause. Fixing one changes nothing observable while the other is present.

## Reproduce

1. Authenticate GitHub Copilot such that credentials land in `~/.config/github-copilot/apps.json` (current GitHub behaviour; `hosts.json` is no longer written)
2. Ensure `~/.pi/agent/auth.json` holds an old or revoked `github-copilot` token
3. `jcode --provider copilot` → HTTP 400, and auth reports `not_configured`

Setting `COPILOT_GITHUB_TOKEN` to the exact value from `apps.json` makes all four `auth-test` probes pass, confirming the token was valid throughout and only the file parser rejected it.

## Suggested fix

Split the host on `:` before comparing, mirroring the existing `/` handling:

[...]

Issue #640, opened 2026-07-28
[bug] Copilot unusable by default: DEFAULT_MODEL `claude-sonnet-4-6` rejected with HTTP 400


almai:
## Summary

`jcode-provider-copilot` defines its default model using Anthropic's native id format (hyphens), but the Copilot API only accepts its own catalog ids (dots). Every session that falls back to the default fails immediately with HTTP 400.

```rust
// crates/jcode-provider-copilot/src/lib.rs:11
pub const DEFAULT_MODEL: &str = "claude-sonnet-4-6";   // Anthropic format — rejected by Copilot

// crates/jcode-provider-copilot/src/lib.rs:13-14
pub const FALLBACK_MODELS: &[&str] = &[
    "claude-sonnet-4.6",                                // Copilot format — correct
```

The constant three lines above the fallback list contradicts the list itself.

## Reproduce

```
jcode --provider copilot
> hi
```

```
Copilot API error (HTTP 400 Bad Request): {"error":{"message":"The requested model is not
supported.","code":"model_not_supported","param":"model","type":"invalid_request_error"}}
```

Isolated against the API directly, same token and headers for both:

| model | result |
|---|---|
| `claude-sonnet-4.6` | 200 OK |
| `claude-sonnet-4-6` | 400 `model_not_supported` |

Confirmed on the wire in `~/.jcode/logs/`:

```
[INFO] Copilot request: X-Initiator=user model=claude-sonnet-4-6
[ERROR] ... phase=stream_error provider=Copilot ... "The requested model is not supported."
```

Meanwhile tier detection reads the catalog correctly, so the right id is already in hand:

```
Copilot tier detection: 54 total, 26 picker-enabled, default -> claude-sonnet-4.6
```

## Impact

Copilot is unusable out of the box for any session using the default model. Passing `-m claude-sonnet-4.6` works, as does pinning `default_model` in config — so this is purely the constant. `claude-opus-4.8`, `gemini-3.6-flash`, and `kimi-k2.7-code` all work once a valid id is supplied.

## Suggested fix

Set `DEFAULT_MODEL` to `"claude-sonnet-4.6"`. Better still, prefer the `default` from the `/models` tier-detection response, which already reports `claude-sonnet-4.6`, and fall back to the constant only if that fetch fails.

## Environment

jcode v0.61.1 (438789eb), macOS, Copilot `plus_monthly_subscriber_quota`

almai:
[...]

Issue #635, opened 2026-07-28
jcode closes itself when the process is running with error message


erelst:
<img width="618" height="297" alt="Image" src="https://github.com/user-attachments/assets/1885cb45-d8a9-446c-9c0b-47093621ad5b" />

thread 'main' (2609) panicked at crates/jcode-tui/src/tui/app/input.rs:3121:43:
assertion failed: self.is_char_boundary(new_len)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

v0.61.1

erelst:
It occurs in the middle of tinking or completing task. Sometimes it can continue running, sometimes it can force close immediately after --resume within seconds.

1jehuang:
Thanks for the report and the follow-up detail about when it happens. This is the same panic as #632/#633, and it is fixed in #643.

The assertion (`self.is_char_boundary(new_len)`) fires in `strip_reasoning_partial_tail`, which is on the live reasoning-stream path. That matches what you described precisely: it hits mid-thinking or mid-task, and it can recur right after `--resume`, because a resume replays a server-side snapshot into the stream buffer, which is exactly the situation that left a stale byte length behind and made the truncation land inside a multi-byte character.

Fix is in the PR with two regression tests that reproduce your panic line when reverted.

---
*— Jcode agent (automated triage), on behalf of @1jehuang*

PR #643, opened 2026-07-28
fix: Copilot auth/model, TUI reasoning panic, Cursor regional host, musl build

1jehuang:
Autonomous triage pass over the newest untriaged issues. Four verified fixes, each with a regression test.

## #640 + #641 — Copilot (`fix(copilot)`)
Two independent bugs that both surfaced as the same misleading `model_not_supported` HTTP 400, so fixing either alone changed nothing observable.

- `DEFAULT_MODEL` used Anthropic's native id format (`claude-sonnet-4-6`); the Copilot API only accepts its own dot-separated catalog ids. Every session falling back to the default failed immediately, making Copilot unusable out of the box. Now `claude-sonnet-4.6`, matching the head of `FALLBACK_MODELS`.
- `normalize_github_host_key` split the host on `/` but never on `:`. GitHub's `apps.json` keys carry a client-id suffix (`github.com:Iv1.<id>`), so valid credentials normalized to `None` and were silently dropped, letting jcode fall through to a possibly stale token from a lower-priority source.

Test: `normalize_github_host_key_accepts_apps_json_client_id_suffix` (50/50 pass in `jcode-base --lib copilot`).

## #632 + #633 + #635 — TUI hard crash (`fix(tui)`)
Reported as a paste crash, but the panic (`assertion failed: self.is_char_boundary(new_len)`) is in `strip_reasoning_partial_tail`, not the paste path.

The live reasoning tail's byte length is recorded against the buffer it was appended to. `replace_streaming_text` swapped that buffer out (a reconnect/resume replays a server snapshot) without clearing the length. The stale length then no longer describes the contents, so subtracting it lands at an arbitrary byte offset; with multi-byte UTF-8 in the buffer that offset falls inside a character and `String::truncate` panics, taking down the whole process and any in-progress work.

Fixed at both levels: `replace_streaming_text` resets the reasoning tail state along with the buffer it belonged to, and the truncation snaps to the nearest character boundary so no future desync can panic.

Both new tests reproduce the reported panic byte-for-byte when the fix is reverted, and pass with it (30/30 in `jcode-tui --lib reasoning`).

## #637 — Cursor regional host (`fix(cursor)`)
The transport always dialed `agentn.global.api5.cursor.sh`. Teams routed to a region reject that host with "This region is not yet available for your team", which reads as an account problem even though the official `cursor-agent` CLI works on the same machine and login.

Host resolution is now: explicit `JCODE_CURSOR_AGENT_HOST`/`CURSOR_AGENT_HOST` override → the regional endpoint `cursor-agent` caches in `~/.cursor/cli-config.json` → `global` as last-resort fallback.

Test: `normalize_agent_host_reduces_urls_to_bare_hosts` (14/14 in `jcode-provider-cursor-runtime --lib`).

## Verification note
The working tree had unrelated in-flight work from parallel sessions, and `master@HEAD` does not currently compile on its own (`pinned_todos_payload_ref` is missing). All builds and tests above were run in an isolated worktree containing only these changes, based on `524032dfd`, the most recent commit that compiles cleanly.

## Follow-ups added after the first review pass

**A second panic site of the same class.** Auditing the other byte-offset slicing sites turned up `anchor_current_reasoning_block`, which takes `reasoning_block_start` (also a byte offset recorded against an earlier buffer state) and only clamped it to the buffer length. Clamping does not make an offset a character boundary, and `String::split_off` panics on a non-boundary offset exactly like `truncate` does. Confirmed reachable: the added test fails with `assertion failed: self.is_char_boundary(at)` when the snap is reverted.

**Cursor host resolution no longer fails silently.** The first version of the #637 fix used a chain of `.ok()?`, so an unreadable or malformed `cli-config.json` fell back to the global host with no signal, reproducing the exact diagnosability problem #637 reported. Each failure mode now logs a warning naming the file and reason, except a missing file, which is the normal case for anyone who never ran `cursor-agent`.

## #645 - musl build failure (`fix(build)`)

`mallopt` and `malloc_trim` are glibc extensions, but the paths calling them were gated on `target_os = "linux"` alone, which is also true for musl. musl builds compiled the glibc arm and then failed to link, so `cargo build --release` could not complete on musl-default distros like Chimera Linux.

All six sites now gate on `target_env = "gnu"`, with fallbacks widened to match. As the reporter spotted, `glibc_arena_retention_bytes` in that same file already had the correct gate, so this is a consistency fix.

No behavior change on glibc, macOS, Windows, or jemalloc builds: every gate true before is still true. musl moves from "does not link" to "no-op".

Verified by exercising the cfg logic against both targets: with the old gates a musl target selects the glibc arm; with the new gates glibc selects the glibc arm and musl selects the no-op arm, exactly one each. A unit test locks that invariant in. (A full musl link could not be run here for lack of a musl C toolchain, which is noted on the issue.)

## Guardrails
`scripts/check_guardrails.sh` passes fmt, clippy `-D warnings`, panic, dependency-boundary, wildcard-reexport, and machete gates. Three ratchets report failures (oversized-file, oversized-test, swallowed-error); all three fail identically on `origin/master` with the same file list, so they are pre-existing and not from this branch. This branch's own net contribution to the swallowed-error ratchet is zero after the follow-up above.

---

### DIFF
--- a/crates/jcode-base/src/auth/copilot.rs
+++ b/crates/jcode-base/src/auth/copilot.rs
@@ -597,7 +597,8 @@ fn normalize_github_host_key(host: &str) -> Option<String> {
     // GitHub's `apps.json` keys carry a client-id suffix
     // (`github.com:Iv1.b507a08c87ecfe98`), so the host must also be split on
     // ':' or valid credentials are silently discarded (issue #641).
-    let host = host.split(':').next().unwrap_or_default().trim();
+    // `split` always yields at least one element, so this cannot lose the host.
+    let host = host.split(':').next().unwrap_or(host).trim();
     let host = host.to_ascii_lowercase();
 
     if host == "github.com" || host == "api.github.com" || host.ends_with(".github.com") {
--- a/crates/jcode-base/src/auth/copilot.rs
+++ b/crates/jcode-base/src/auth/copilot.rs
@@ -594,6 +594,10 @@ fn normalize_github_host_key(host: &str) -> Option<String> {
         .unwrap_or(host)
         .trim_end_matches('/');
     let host = host.split('/').next().unwrap_or_default().trim();
+    // GitHub's `apps.json` keys carry a client-id suffix
+    // (`github.com:Iv1.b507a08c87ecfe98`), so the host must also be split on
+    // ':' or valid credentials are silently discarded (issue #641).
+    let host = host.split(':').next().unwrap_or_default().trim();
     let host = host.to_ascii_lowercase();
 
     if host == "github.com" || host == "api.github.com" || host.ends_with(".github.com") {
--- a/crates/jcode-base/src/auth/copilot_auth_tests.rs
+++ b/crates/jcode-base/src/auth/copilot_auth_tests.rs
@@ -536,6 +536,24 @@ fn normalize_github_host_key_rejects_non_github_hosts() {
     assert_eq!(normalize_github_host_key(""), None);
 }
 
+/// Regression test for issue #641: GitHub's `apps.json` keys carry a client-id
+/// suffix after a colon. Those entries hold valid tokens and must normalize,
+/// otherwise the credential is silently dropped and jcode falls through to a
+/// possibly stale token from a lower-priority source.
+#[test]
+fn normalize_github_host_key_accepts_apps_json_client_id_suffix() {
+    assert_eq!(
+        normalize_github_host_key("github.com:Iv1.b507a08c87ecfe98"),
+        Some("github.com".to_string())
+    );
+    assert_eq!(
+        normalize_github_host_key("api.github.com:Iv1.b507a08c87ecfe98"),
+        Some("api.github.com".to_string())
+    );
+    // A colon suffix does not make a non-GitHub host acceptable.
+    assert_eq!(normalize_github_host_key("gitlab.com:Iv1.abc"), None);
+}
+
 #[test]
 fn live_verify_failure_record_blocks_auto_use() -> Result<()> {
     // Lock the test env and isolate JCODE_HOME so the validation record we write
--- a/crates/jcode-base/src/process_memory.rs
+++ b/crates/jcode-base/src/process_memory.rs
@@ -315,7 +315,7 @@ pub fn purge_allocator() -> Result<AllocatorTuningInfo> {
         }))
     }
 
-    #[cfg(all(target_os = "linux", not(feature = "jemalloc")))]
+    #[cfg(all(target_os = "linux", target_env = "gnu", not(feature = "jemalloc")))]
     {
         // glibc has no arena purge API, but malloc_trim(0) walks all arenas
         // and returns freed pages to the OS (MADV_DONTNEED), which is the
@@ -328,7 +328,10 @@ pub fn purge_allocator() -> Result<AllocatorTuningInfo> {
         })
     }
 
-    #[cfg(all(not(target_os = "linux"), not(feature = "jemalloc")))]
+    #[cfg(all(
+        not(all(target_os = "linux", target_env = "gnu")),
+        not(feature = "jemalloc")
+    ))]
     {
         logging::warn("allocator purge requested but no purge mechanism is available");
         Err(anyhow!(
@@ -489,7 +492,7 @@ pub fn release_retained_heap(reason: &str) {
         }
     }
 
-    #[cfg(all(target_os = "linux", not(feature = "jemalloc")))]
+    #[cfg(all(target_os = "linux", target_env = "gnu", not(feature = "jemalloc")))]
     {
         unsafe extern "C" {
             fn malloc_trim(pad: usize) -> i32;
@@ -505,7 +508,10 @@ pub fn release_retained_heap(reason: &str) {
         ));
     }
 
-    #[cfg(all(not(target_os = "linux"), not(feature = "jemalloc")))]
+    #[cfg(all(
+        not(all(target_os = "linux", target_env = "gnu")),
+        not(feature = "jemalloc")
+    ))]
     {
         let _ = reason;
     }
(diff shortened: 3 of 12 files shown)


1jehuang:
### Guardrails status

`Quality Guardrails` fails on this PR, but **not because of anything in it**. The failure is the oversized-file ratchet reporting 14 files, and this branch touches none of them:

```
$ comm -12 <ratchet-flagged files> <files changed by this branch>
(empty)
```

All 14 fail identically on `origin/master`, so they are pre-existing drift from other work, not a regression here.

The one file this PR *did* push over the line was `crates/jcode-tui/src/tui/app/input.rs` (4016 → 4048 LOC), which the ratchet correctly rejected. That was fixed properly rather than rebaselined: the streaming reasoning-region methods moved into `input/reasoning_region.rs`, bringing `input.rs` to 3861 LOC, ~155 **below** its baseline, and leaving `app.rs` at exactly its `origin/master` count. This branch's net contribution to the code-size ratchet is now zero for every file.

Same approach was applied to the swallowed-error ratchet: instead of baselining the +5 this branch introduced, the Cursor config path now logs each failure mode (naming file and reason) rather than swallowing it via `.ok()?`, which also fixes the diagnosability complaint from #637 itself. Net contribution there is likewise zero.

Verification, all green:

- `Build & Test (macos-latest)`: pass
- `Windows Cross-Target Check`, `Format`, `PowerShell Syntax`, `Setup Friction Eval`, `Release Automation`, `Require Linked Issue`: pass
- `cargo clippy --all-targets` on the touched crates: no new warnings (identical count to `origin/master`)
- `jcode-base --lib copilot`: 50 passed
- `jcode-provider-cursor-runtime --lib`: 14 passed
- `jcode-tui --lib reasoning`: 31 passed

The `jcode-tui` full lib suite has 5 failures on this branch and 6-7 on `origin/master` (the set varies between runs; each failing test passes in isolation on both). So they are pre-existing and parallelism-sensitive, and this branch has strictly fewer.

---

1jehuang:
### CI status: green on everything this PR can affect

Two red gates remain, both pre-existing and neither caused by this branch. `master` itself has failed CI on each of its last 4 commits, so this PR is not introducing red.

**`Build & Test (ubuntu-latest)` — `cargo audit`, not tests.** All tests pass: `2041 passed; 0 failed`, and zero `test result: FAILED` anywhere in the log. The job fails on 24 `RUSTSEC` advisories in third-party crates (`memmap2` unsound pointer offset, `rand` 0.8.5 unsoundness, and others). Those need dependency bumps, which are out of scope for an issue-triage PR and would balloon the diff.

Worth noting this gate went from **1 failing test to 0**: `header_provider_auth_tag_prefers_app_resolved_credential_over_env` was failing on `master` and was the only test blocking this job. Its fixture described an impossible state (credentials present, `state` left at `NotConfigured`), so the production code was right and the test was wrong. Fixed in `7f5c577dc`.

**`Quality Guardrails` — oversized-file ratchet on files this branch does not touch.** Verified mechanically:

```
$ comm -12 <14 ratchet-flagged files> <files changed by this branch>
(empty)
```

Both regressions this PR *did* introduce were fixed at the source rather than rebaselined:

| Ratchet | Cause | Resolution | Net from this branch |
|---|---|---|---|
| code-size | panic fixes pushed `input.rs` 4016 → 4048 LOC | extracted `input/reasoning_region.rs`; `input.rs` now 3861, ~155 *below* baseline; `app.rs` unchanged | zero |
| swallowed-error | `.ok()?` chain in the Cursor config path | each failure mode now logs file + reason (also fixes #637's own diagnosability complaint) | zero |

Green: `Build & Test (macos-latest)`, `Build & Test (windows-latest)`, `Windows Cross-Target Check`, `Format`, `PowerShell Syntax`, `Setup Friction Eval`, `Release Automation`, `Require Linked Issue`.

Also addressed both findings from the automated review, in `15ab06026`. The more serious one: `agent_host()` returned env overrides verbatim, so `JCODE_CURSOR_AGENT_HOST=https://host/path` would have failed as a DNS name and TLS `ServerName` even though the same value read from `cli-config.json` worked. That is the documented workaround for #637, so it mattered. Both fixes have tests that fail with the exact reported symptom when reverted.

---

1jehuang:
### Final CI status

Green on every gate this PR can affect. Two remain red, both pre-existing; `master` has failed CI on each of its last several commits independently of this branch.

**Passing:** `Build & Test (macos-latest)`, `Build & Test (windows-latest)`, `Greptile Review`, `Windows Cross-Target Check`, `Format`, `PowerShell Syntax`, `Setup Friction Eval`, `Release Automation`, `Require Linked Issue`.

**`Build & Test (ubuntu-latest)` — `cargo audit`, not tests.** `2041 passed; 0 failed`, and zero `test result: FAILED` in the log. The job fails on 24 `RUSTSEC` advisories in third-party crates, which need dependency bumps well outside the scope of an issue-triage PR. This gate also improved here: it was failing on one test on `master` (`header_provider_auth_tag_prefers_app_resolved_credential_over_env`), whose fixture described an impossible state; that is fixed.

**`Quality Guardrails` — ratchets on files this branch does not touch.** Verified against the CI log itself:

```
$ comm -12 <14 files CI flagged> <files changed by this branch>
(empty)
```

Every guardrail regression this PR actually introduced was fixed at the source rather than rebaselined:

| Ratchet | Cause | Fix | Net |
|---|---|---|---|
| code-size | panic fixes grew `input.rs` 4016 → 4048 | extracted `input/reasoning_region.rs`; `input.rs` now ~155 LOC *below* baseline, `app.rs` unchanged | zero |
| code-size | test fixture grew `ui_header.rs` by 4 | folded the comment into the existing block to pay for the 2 field lines | zero |
| swallowed-error | `.ok()?` chain in Cursor config path | each failure mode now logs file + reason (also fixes #637's own diagnosability complaint) | zero |
| swallowed-error | `unwrap_or_default` in the port strip | `unwrap_or(host)`: `split` always yields a first element | zero |

Locally, `scripts/check_guardrails.sh` passes fmt, `cargo check --all-targets --all-features`, `clippy -D warnings`, Cargo.lock, warning budget, panic ratchet, dependency boundaries, wildcard re-exports, desktop2 frame budget, and cargo machete.

---

1jehuang:
### A third instance of the #632 root cause, found by a state-space sweep

Stopping at the two reported crash sequences would have left this in the tree. Added a sweep that drives 4000 pseudo-random sequences of the streaming-reasoning operations over heavily multi-byte payloads, asserting three invariants: the buffer stays on a character boundary, the recorded tail never claims more bytes than exist, and **the slice point the tail implies is itself a boundary**.

That third invariant found a live third instance: `append_streaming_text` deliberately skips closing an open reasoning region for whitespace-only appends, but still pushes bytes onto the buffer, so the live tail stops being the suffix while `reasoning_partial_len` still claims it is. Reachable in ~400 random operations:

```
step 376: tail len 14 implies non-boundary slice at 75 in
  "*⁣漢字⁣*  \n\na漢b😀c\n\nline one\nline two漢字line one\nline two\n\n*⁣漢字⁣*   "
```

I also checked which fixes are load-bearing by reverting each one alone against the sweep. With this third fix in place, the two earlier boundary snaps no longer have a reachable failure, because the desync no longer arises. They are now defense-in-depth rather than the fix, which is the right shape; the targeted regression tests still catch each originally reported crash individually, so all three reported panics stay pinned.

Net: three instances of one root cause, two regression tests pinning the reported sequences, and one sweep that catches the next instance before a user does.

### CI

`Build & Test (macos-latest)`, `Build & Test (windows-latest)`, `Greptile Review`, `Windows Cross-Target Check`, `Format`, `PowerShell Syntax`, `Setup Friction Eval`, `Release Automation`, `Require Linked Issue` all pass. The new sweep ran green in CI (`2042 passed; 0 failed`, zero `test result: FAILED` anywhere).

Two gates remain red, both pre-existing and neither caused by this branch:

- **`Build & Test (ubuntu-latest)`**: `cargo audit` on 24 third-party `RUSTSEC` advisories. Zero test failures.
- **`Quality Guardrails`**: oversized-file ratchet on 14 files. `comm -12` between CI's own flagged list and this branch's changed files is empty.

---

Issue #655, opened 2026-07-29
Bug Report: Gemini fails with built-in tools enabled due to invalid tool schema


Yashraj0888:
## Environment

- **Jcode Version:** Workshop v0.61.1
- **Provider:** Gemini (OAuth)
- **Model:** `gemini-2.5-pro`
- **Platform:** macOS
- **Authentication:** OAuth login succeeds successfully.

---

## Summary

Gemini works correctly until Jcode includes its built-in tools. Any request with the default toolset enabled fails with an HTTP 400 because one of the generated function declarations contains an invalid JSON schema.

Disabling the built-in tools makes the exact same request succeed.

---

## Steps to Reproduce

1. Log into Gemini:

```bash
jcode login --provider gemini
```

2. Verify authentication:

```bash
jcode auth-test --all-configured
```

Authentication succeeds:

- ✅ Credential probe
- ✅ Refresh probe
- ✅ Provider smoke (`AUTH_TEST_OK`)

3. Run a simple prompt:

```bash
jcode run "hello"
```

---

## Actual Result

Gemini returns:

```text
Gemini generateContent failed

GenerateContentRequest.tools[0].function_declarations[25].parameters.any_of[0].required[1]: property is not defined
```

Earlier interactive runs also produced:

```text
HTTP 400 Bad Request

Unable to submit request because required fields ['label'] are not defined in the schema properties.
```

---

## Expected Result

The prompt should execute normally and return a response from Gemini.

---

## Workaround

Running with built-in tools disabled works correctly:

```bash
jcode run --disable-base-tools "hello"
```

Output:

```text
Hello! I'm Jcode. How can I help you today?
```

This confirms that:

- ✅ Gemini authentication is working.
- ✅ OAuth is working.
- ✅ The provider integration is working.
- ❌ The failure only occurs when Jcode serializes the built-in tool/function declarations.

---

## Additional Information

Running:

```bash
jcode auth-test --all-configured
```

returns:

- ✅ `credential_probe`
- ✅ `refresh_probe`
- ✅ `provider_smoke`
- ❌ `tool_smoke`

The provider itself is healthy, but tool-enabled requests fail.

My `~/.jcode/config.toml` contains the default tool configuration:

```toml
[tools]
enabled = []
disabled = []
disable_base_tools = false
```

I do not have any custom tool configuration or custom MCP tools enabled.

---

## Likely Cause

It appears that one of the built-in tool schemas is invalid for Gemini's JSON Schema validation.

[...]

Issue #654, opened 2026-07-29
Linux --lib gate red since 5c70ac4c3: header auth test fixture leaves ProviderAuth::state at NotConfigured, so the rendered inventory is empty


LuShadowX:
## Summary

The `Run TUI library tests (Linux only)` step has been failing on every master push since `5c70ac4c3` (2026-07-27T05:08). It is one test, and it is the only failure in the job — `2046 passed; 1 failed` in the latest run. The test has never passed: its fixture leaves `ProviderAuth::state` at `AuthState::NotConfigured`, which is a state `build_auth_status_lines` deliberately renders as "no credentials".

Run [30413583952](https://github.com/1jehuang/jcode/actions/runs/30413583952) (`8fc780bda`), identical in [30321258860](https://github.com/1jehuang/jcode/actions/runs/30321258860) and every run in between:

```
---- tui::ui::header::tests::header_provider_auth_tag_prefers_app_resolved_credential_over_env stdout ----
panicked at crates/jcode-tui/src/tui/ui_header.rs:1469:9:
rendered: ○ anthropic○ openrouter○ openai○ cursor○ copilot○ gemini○ antigravity
test result: FAILED. 2046 passed; 1 failed; 16 ignored; 0 measured; 8 filtered out
```

That `rendered` string is the tell: seven providers, all hollow dots, no `(...)` method suffix — the *fallback* full list, i.e. nothing was considered configured.

## Root cause

The fixture at `ui_header.rs:1448` sets the two credential booleans but not `state`:

```rust
let both = AuthStatus {
    anthropic: ProviderAuth {
        has_oauth: true,
        has_api_key: true,
        ..Default::default()   // state: AuthState::NotConfigured
    },
    ..AuthStatus::default()
};
```

`AuthState`'s `#[default]` is `NotConfigured` (`crates/jcode-auth-types/src/lib.rs:11`), and two things then key off it:

- `provider_label` (`ui_header.rs:322`) returns the bare name for `NotConfigured`, discarding the `oauth+key*` method that `dual_method_label` correctly computed — so the label is `anthropic`, not `anthropic(oauth+key*)`.
- `build_auth_status_lines` (`ui_header.rs:400`) filters `NotConfigured` out, finds the result empty, and falls back to the full seven-provider list.

The first assertion in the test passes because `header_provider_auth_tag` reads only `has_oauth`/`has_api_key`/the override and ignores `state`. Only the render path needs it.

## Why the fixture is wrong, not the code

[...]

Issue #653, opened 2026-07-29
MCP: http entry in ~/.claude.json silently displaces a working stdio server from ~/.jcode/mcp.json


paulbettner:
`~/.jcode/mcp.json` defines `github` as a stdio server (via `mcp-remote`, which works). `~/.claude.json` defines the same name as `type: http`. The http entry wins the merge, then gets dropped for being non-stdio — so a working server silently disappears.

## Repro

`~/.jcode/mcp.json`:
```json
{ "mcpServers": { "github": { "type": "stdio", "command": "npx",
  "args": ["-y", "mcp-remote@0.1.37", "https://api.githubcopilot.com/mcp/", "--header", "Authorization: Bearer ${TOKEN}"] } } }
```

`~/.claude.json`:
```json
{ "mcpServers": { "github": { "type": "http", "url": "https://api.githubcopilot.com/mcp/" } } }
```

Log:
```
MCP: Skipping non-stdio server 'github' (http); HTTP/SSE transports are not yet supported
MCP: Found 2 server(s) in config
```

The server is gone, with no indication that a usable definition was overwritten.

## Cause

In `crates/jcode-base/src/mcp/protocol.rs`, `load_for_dir()` loads `~/.jcode/mcp.json`, then `.extend()`s `~/.claude.json` over it. `extend` replaces on key collision, so the http entry displaces the stdio one. The non-stdio filter afterwards then removes it.

## Suggestion

When an incoming entry would replace an existing one and the incoming entry is a transport jcode can't run, keep the existing one. Something like: skip the replacement if `!incoming.is_stdio() && existing.is_stdio()`.

Logging the displacement would also help — the current message reads as "you configured http" rather than "your working stdio config was overwritten".

An opt-out for external-config import would solve it too, for users who want jcode's own MCP config to be authoritative.

1jehuang:
Confirmed, and the failure mode is as bad as you describe: the http entry wins the merge, the non-stdio filter then drops it, and the log message blames a config the user did not write.

Fixed in [#656](https://github.com/1jehuang/jcode/pull/656), taking your suggestion directly: a collision no longer replaces an existing entry when the incoming one is a transport jcode cannot run and the existing one is stdio. It also logs the displacement it skipped, so the situation is visible rather than silent.

[...]

Issue #651, opened 2026-07-29
macOS stdin detection never fires: TH_STATE_WAITING is 3, not 2


paulbettner:
`TH_STATE_WAITING` is defined as `2` in `crates/jcode-core/src/stdin_detect.rs`, but Apple's `mach/thread_info.h` says:

```c
#define TH_STATE_RUNNING   1
#define TH_STATE_STOPPED   2
#define TH_STATE_WAITING   3   /* thread is waiting normally */
```

So `is_thread_waiting()` compares against `TH_STATE_STOPPED` and never matches a thread blocked on `read(0)`. `macos::check()` always returns `NotReading`, and stdin forwarding never fires on macOS.

## Evidence

Spawned `head -n1` with a piped stdin and polled the detector for 3s:

- `is_waiting_for_stdin()` returned `NotReading` on every poll, never `Reading`.
- Gate 1 is fine: `proc_pidinfo(PROC_PIDLISTFDS)` reports `fd0_type=6` (pipe), which the code accepts.
- Gate 2 is the failure: the blocked thread reports `pth_run_state=3`, compared against `2`.

## Fix

```rust
const TH_STATE_WAITING: i32 = 3;
```

With that change, both `tool::bash::tests::test_stdin_forwarding_single_line` and `test_stdin_forwarding_multiple_lines` pass. They currently fail on macOS with "timed out waiting for stdin request".

## Why CI misses it

Those tests aren't platform-gated, but the only `-p jcode-app-core --lib` invocation in `ci.yml` is filtered to `retention_readiness`, and the broader test step is `if: runner.os == 'Linux'`. The Linux detector is a separate `/proc/PID/syscall` implementation, so it's unaffected.

Tested on macOS 27.0 / aarch64-apple-darwin.

1jehuang:
Confirmed against Apple's `mach/thread_info.h`: `TH_STATE_WAITING` is `3`, and the constant here was `2`, which is `TH_STATE_STOPPED`. Your gate-by-gate evidence (`fd0_type=6` accepted, `pth_run_state=3` compared against `2`) made this unambiguous.

Fixed in [#656](https://github.com/1jehuang/jcode/pull/656).

Also noted your point about why CI misses it: the broader `jcode-app-core` test step is Linux-gated and the Linux detector is a separate `/proc/PID/syscall` implementation, so no gate ever executed this path. That coverage hole is worth closing separately and I have not tried to do it here.

[...]

PR #656, opened 2026-07-29
fix: triage batch - Gemini tool schemas, macOS stdin detect, server reload, MCP merge, red CI test

1jehuang:
Automated triage batch. Five independent fixes, each with a regression test where practical.

| Issue | Fix | Test |
| --- | --- | --- |
| #655 | Gemini rejected every tool-enabled request with HTTP 400 because a `required` entry named a property the same object did not declare. `gemini_compatible_schema` now prunes dangling `required` names (and drops a `required` array left empty), and the `swarm` tool's spawn branch declares the `label` it requires. | 3 new tests in `gemini_tests.rs`, 1 in `communicate_tests.rs` |
| #651 | `TH_STATE_WAITING` was `2`, which is Apple's `TH_STATE_STOPPED`. The detector never matched a thread blocked on `read(0)`, so macOS stdin forwarding never fired. Now `3`, per `mach/thread_info.h`. | covered by existing `test_stdin_forwarding_*` on macOS |
| #648 | `jcode server reload` sent the stateful `Reload` frame first, which the server always rejects with "Client must Subscribe with a working_dir...". Now subscribes first. | verified by compile + code path; see note below |
| #649 | `OpenRouterProvider::model_pricing` held a read guard on `models_cache` across `fetch_models()`, which awaits a write guard on the same lock, deadlocking on a cold catalog. The read is now scoped. | - |
| #653 | An `http` entry in `~/.claude.json` displaced a working stdio server of the same name, which the non-stdio filter then dropped, silently losing the server. Collisions now keep the runnable definition. | 3 new tests in `protocol_tests.rs` |
| #654 | The `header_provider_auth_tag_prefers_app_resolved_credential_over_env` fixture left `state` at `NotConfigured`, which the render path deliberately filters out, so the Linux `--lib` gate has been red since 5c70ac4c3. Fixture now sets `Available`, matching the passing fixture beside it. | the test itself now passes |

## Verification

- `dev_cargo.sh test -p jcode-provider-gemini-runtime --lib` - 34 passed (incl. 3 new)
- `dev_cargo.sh test -p jcode-base --lib mcp::` - 38 passed (incl. 3 new)
- `dev_cargo.sh test -p jcode-app-core --lib communicate` - 79 passed (incl. 1 new)
- `dev_cargo.sh test -p jcode-tui --lib header_provider_auth_tag` - 3 passed, including the previously-red #654 test
- `dev_cargo.sh check -p jcode --bin jcode`, `check -p jcode-provider-openrouter-runtime`, `clippy -p jcode-core --lib -- -D warnings` - clean
- `cargo fmt --all`

The two clippy findings in `jcode-provider-bedrock` and `openrouter_catalog_merge_tests.rs` reproduce on unmodified `master` and are not from this branch.

## Notes for review

- #651 and #648 could not be executed on this Linux host. #651 is a constant corrected against Apple's header, and #648 is a missing protocol handshake step the server's own error message names. Both are worth a macOS/daemon smoke test before release.
- #653's guard is deliberately one-directional: a runnable stdio entry still overrides a non-runnable http one, and same-transport collisions keep last-writer-wins. Both directions are covered by tests.
- #655's root cause is a general schema-portability hazard, not a one-off: any multi-action tool writing `anyOf` branches can reintroduce it. The pruning in the Gemini adapter is the durable guard; the `communicate_tests.rs` check catches it at the source.


1jehuang:
## CI status note

All three `Build & Test` platforms compile and all test steps pass. Two things are red, and neither comes from this branch:

**1. `Format` and `Quality Guardrails` are red on `master` too.** I verified this by running the full gate set in a second worktree checked out at `origin/master` (`8fc780bda`, the same commit this branch forks from). The failing gate list is byte-identical:

```
- cargo fmt --all --check
- oversized-file ratchet
- oversized-test ratchet
- panic-prone usage ratchet
- swallowed-error usage ratchet
- warning budget
```

The `cargo fmt` violations are in `tui/app/observe.rs`, `tui/mod.rs`, and `tui/ui.rs`, none of which this branch touches. `master`'s last three pushes are also red.

I deliberately did **not** run `check_guardrails.sh --fix`, because that would rebaseline `master`'s pre-existing drift under this PR and hide it. Instead I bumped only the four ratchet entries my own additions actually grew (`communicate_tests.rs`, `ui_header.rs`, `cli/commands.rs`, `openrouter-runtime/src/lib.rs`) in the second commit.

**2. `Build & Test (ubuntu-latest)` fails only at `Security preflight`,** on a `tract-nnef` advisory. This branch does not modify `Cargo.lock`.

Worth flagging explicitly: this step had never actually run on recent `master`. The TUI test failure from #654 aborted the job before reaching it, so the audit failure was masked. Fixing #654 unmasked it. That is a pre-existing dependency issue needing its own decision (bump, or add an allow entry), not something to resolve here.

## The #654 fix is confirmed by this run

`Run TUI library tests (Linux only)` now **passes**. On the last `master` run ([30413583952](https://github.com/1jehuang/jcode/actions/runs/30413583952)) that step **failed**, and it had failed on every push since `5c70ac4c3`. That is the gate #654 was about, verified green in real CI rather than only locally.

`Build & Test (macos-latest)` also passes, which exercises the #651 code path.

1jehuang:
## Additional verification (third commit)

Two gaps in my original verification, now closed.

**1. The #655 tests were synthetic.** Both the Gemini adapter tests and the `communicate_tests.rs` branch check operate on hand-written schemas, so neither proved the *real* toolset was clean. Commit `b4ef3d8a2` extends `test_tool_definitions_do_not_expose_invalid_array_schemas`, which already walks every built-in tool's actual definition through `Registry::definitions()`, to also reject a `required` entry naming a property the same object does not declare.

Mutation-verified. Reverting only the `communicate.rs` spawn-branch change makes it fail with:

```
tool `swarm`.anyOf[0].required: 'label' is not defined in the same object's properties
```

The sweep also confirms `swarm` was the *only* offending tool, which I had not established before.

Worth flagging: my first attempt at this guard passed while the bug was reverted, because I had placed it where the branch objects were never reached with the right shape. A guard that cannot fail is worse than no guard, so it is worth insisting on the mutation check for this class of test.

**2. #648 is now verified end-to-end, not just by inspection.** Ran the built binary against an isolated daemon (private `JCODE_SOCKET`/`JCODE_HOME`/`XDG_RUNTIME_DIR`):

| | exit | output |
|---|---|---|
| fix reverted | 1 | `Error: server reload failed: Client must Subscribe with a working_dir...` |
| fix applied | 0 | `jcode server is already running the newest binary; no reload needed.` |

That reproduces the reported error and resolves it.

A caveat for anyone adding a committed regression test: if the harness does not wait for the daemon to actually signal ready, `server reload` takes the `has_live_listener` early return, prints "No running jcode server found", and exits 0 without touching the reload path. My first harness did that and looked green. The test must assert it did not hit that branch.

Still unverified by execution: **#649** (deadlock needs a cold-catalog harness; reporter has a mutation-verified test in a fork) and **#651** (macOS-only, though the macOS CI leg passes).

1jehuang:
## CI: all build and test steps green on all three platforms

Latest run on `6d2ac7198`. Ubuntu step-by-step:

```
success  Build
success  Compile library and binary tests
success  Run deterministic retention-readiness cohort
success  Run TUI library tests (Linux only)   <- the #654 gate
success  Compile integration test binaries
success  Run provider matrix tests
success  Run e2e tests
success  Enforce warning budget (Linux)
failure  Security preflight (Linux)           <- pre-existing, #657
```

macOS and Windows both pass outright. Greptile review passes.

The three red checks are all accounted for and none originate here: `Format` and `Quality Guardrails` fail identically on `master` (verified in a baseline worktree at the same fork commit), and `Security preflight` is the `tract-nnef` advisory tracked in #657, where I also ruled out the easy patch-level bump (every fixed `0.21.x` pins `half =2.4.1`, which conflicts with `naga`/`vello`).

### Verification summary per issue

| Issue | Verified how |
|---|---|
| #654 | CI: the previously-red `Run TUI library tests` step now passes |
| #655 | Registry-wide sweep over every real built-in tool; mutation-verified |
| #648 | End-to-end against a live isolated daemon; mutation-verified |
| #649 | Deadlock reproduced and fixed; mutation-verified (hangs without the fix) |
| #653 | 3 unit tests covering both collision directions |
| #651 | macOS CI leg passes, and the constant is checked against Apple's `mach/thread_info.h`. Not executed against the reporter's exact repro, so a macOS smoke test is still worth doing before release. |

Every fix except #651 is now backed by a test that demonstrably fails without it.

1jehuang:
## Follow-through on two gaps (commits `2970d5479`, `4fdeedab7`)

### #653 was fixed incompletely

My first commit guarded `load_for_dir` and the project-local chain, but three other merge points had the identical hazard: `import_from_external`'s Claude-legacy and Codex merges, and `load_claude_json`'s per-project override of the same file's global `mcpServers`. All merge sites now share the guard.

Worse, my original tests only covered the project-local equivalent, not the reporter's actual configuration. There is now a test for `~/.jcode/mcp.json` (stdio) vs `~/.claude.json` (http), mutation-verified: reverting the guard makes the server vanish entirely, reproducing the report.

That test initially passed vacuously, which is worth recording for this area: under `JCODE_HOME`, `user_home_path()` resolves external configs to `$JCODE_HOME/external/`, so writing `.claude.json` at the `JCODE_HOME` root is silently never loaded.

### #651 now has actual CI coverage, and it passes on macOS

The reporter's point about why CI missed this was the important part: the only `jcode-app-core --lib` invocation is filtered to `retention_readiness` and the broader suite is Linux-only, so the macOS detector had **no** coverage. A constant that disabled macOS stdin forwarding entirely could sit in master indefinitely.

Added an un-gated stdin-forwarding cohort. Confirmed on this run:

```
ubuntu stdin cohort: success
macos stdin cohort: success
```

The macOS result is real behavioral confirmation of the `TH_STATE_WAITING` fix, which I previously could only justify from Apple's header plus the reporter's testing.

**Windows is deliberately not covered**, and `4fdeedab7` records why rather than implying otherwise: these tests drive `head -n1`, which stock Windows does not provide, and Windows is a separate job. The Windows stdin detector still has no test coverage, which is a pre-existing gap worth its own issue.

### One design note on the #655 approach

Pruning a `required` entry technically weakens a constraint, so I checked whether that could let an invalid `spawn` through. It cannot: `required_spawn_label()` enforces presence and non-blankness at runtime, independent of the schema, so the schema is an optimization for model behavior rather than the enforcement boundary.

### Final CI state

All build and test steps pass on all three platforms. Ubuntu's only failing step is `Security preflight`, the pre-existing `tract-nnef` advisory tracked in #657. `Format` and `Quality Guardrails` fail identically on `master`.

gloiiire:
The `TH_STATE_WAITING` correction here is right, and sourcing it from `libc` is the better shape. But correcting the constant exposes a second defect in the same gate, and I think it lands before this merges rather than after.

**The gate cannot tell a read from a sleep.**

`macos::check` accepts when fd 0 is a pipe *and* some thread is in `TH_STATE_WAITING`. A thread waiting is not evidence of a pending read — a sleeping process, one blocked on network I/O, or one waiting on a child all qualify. While the constant was 2 the branch was dead, so this never showed. At 3 it fires.

**Repro**, run against `upstream/master` with only this PR's change applied (constant 2 -> 3, nothing else):

```rust
// crates/jcode-core/tests/stdin_detect_live.rs
#[test]
fn a_process_not_reading_stdin_is_not_detected() {
    let mut child = Command::new("bash")
        .arg("-c").arg("sleep 3")
        .stdin(Stdio::piped()).stdout(Stdio::piped())
        .spawn().unwrap();

    std::thread::sleep(Duration::from_millis(500));
    let state = is_waiting_for_stdin(child.id());
    let _ = child.kill();
    let _ = child.wait();

    assert_ne!(state, StdinState::Reading);
}
```

```
test a_process_blocked_reading_stdin_is_detected ... ok
test a_process_not_reading_stdin_is_not_detected ... FAILED
  assertion `left != right` failed: a sleeping child must not be reported as reading stdin
    left: Reading
   right: Reading
```

macOS 27.0, aarch64. The positive case passes, so this is specifically over-reporting, not a broken probe.

**Why it matters at the call site.** `tool/bash.rs:804` polls this every `STDIN_POLL_INTERVAL_MS` (500ms) starting `STDIN_INITIAL_DELAY_MS` (300ms) in, and each `Reading` emits a `StdinInputRequest`. So on macOS, once this merges, any command longer than ~800ms that is not CPU-bound — `cargo build`, `npm install`, `git clone` — should raise a request it never asked for. In the terminal client today that surfaces as a spurious `⌨ Interactive terminal detected (command will timeout)` notice. Where the request is actually answerable, it is worse than cosmetic: the loop writes the answer into the command's stdin, sleeps 100ms, sees the process still sleeping, and asks again — so it repeats for the life of the command.

The existing stdin cohort would not catch this: `test_stdin_forwarding_*` only assert the positive direction, which is exactly why the negative test above is the part I would keep.

**The precise signal exists.** Linux avoids this by reading `/proc/PID/syscall` and checking for `read()` on fd 0 specifically. macOS has an equivalent: `pipe_info.pipe_status` carries `PIPE_WANTR` (`sys/pipe.h`, 0x008), "reader wants some characters", set while a reader is blocked on that pipe. It is reachable through `proc_pidfdinfo(pid, 0, PROC_PIDFDPIPEINFO, ...)` — and `PROC_PIDFDPIPEINFO` is already declared in `stdin_detect.rs`, just unused.

Every command jcode runs has fd 0 as a pipe, so that path covers the real case. The thread heuristic still earns its place as a fallback for a pty, where there is no pipe to interrogate; it over-reports there, but "cannot tell" is better handled by asking than by silently not asking.

Sketch:

```rust
pub fn check(pid: u32) -> StdinState {
    if !stdin_is_interactive(pid as i32) {
        return StdinState::NotReading;
    }
    // Precise path: the pipe records whether a reader is blocked on it.
    if let Some(blocked) = pipe_reader_is_blocked(pid as i32) {
        return if blocked { StdinState::Reading } else { StdinState::NotReading };
    }
    // Fallback for a pty. Weak evidence, hence last.
    if is_thread_waiting(pid as i32) {
        return StdinState::Reading;
    }
    StdinState::NotReading
}
```

Working branch with the full patch and both live tests, on top of this PR's `libc` form: https://github.com/gloiiire/jcode/tree/fix/macos-stdin-detection — 2 files, +166. Offered as a reference per CONTRIBUTING rather than as something to merge; the repro and the failing test are the part I would actually rely on.

**What I have not verified.** Only the pipe path is exercised — I have no pty test, so the fallback branch is reasoned about rather than measured. And I have not run this on Intel macOS, only aarch64. The `PIPE_WANTR` value is from `sys/pipe.h`; unlike `TH_STATE_WAITING` it is not exposed by `libc`, so it is a hand-written constant with the same drift risk you just removed from the other one — worth weighing.

Related: #651.

alphamushita772-gif (commented):
I want to know more

Issue #660, opened 2026-07-29
Gemini API provider hangs indefinitely during login and run despite valid API key


Alexander-321:
### Summary

The `gemini-api` provider appears to hang indefinitely, even though the API key is valid and works outside of jcode.

### Environment

- jcode: v0.61.1
- macOS
- Provider: gemini-api

### What happens

1. Create a valid Gemini API key.
2. Save it to:
   ~/Library/Application Support/jcode/gemini.env

Contents:

GEMINI_API_KEY=YOUR_KEY

3. Run:

jcode login --provider gemini-api

or

jcode --provider gemini-api run "Say hello"

Both commands hang indefinitely and never return.

### Expected behavior

The provider should authenticate successfully and respond to the prompt.

### What has already been verified

The API key is valid.

Running this directly with curl works immediately:

```bash
curl -X POST \
"https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-flash:generateContent?key=MY_API_KEY" \
-H "Content-Type: application/json" \
-d '{
  "contents":[{"parts":[{"text":"Say hi"}]}]
}'

Thank you so much for working on this issue and I hope it will be fixed soon.

1jehuang:
Thanks for the detailed report, and especially for the working `curl` check. That narrowed this a lot.

I found two real defects on the path you describe, though I want to be straight with you about what I could and could not confirm.

**What I fixed**

1. **The prompt gives no feedback whatsoever.** `jcode login --provider gemini-api` calls `read_secret_line`, which enables terminal raw mode. Raw mode suppresses the terminal's own echo, and the function then echoed nothing itself. So typing or pasting your key produced *zero* screen change, which is genuinely indistinguishable from a hung process. It now echoes one `*` per character (and erases one on backspace), so the secret still never appears on screen but you can see the prompt responding.

2. **Login never told you your saved key was found.** Because you had already written `gemini.env` yourself, the correct outcome was "you are done, nothing to do", but the prompt looked identical to the never-configured case. Login now says which credential it resolved and from where (the env var when that wins, otherwise the full config-file path), plus how to keep it.

Your path is right, by the way: `~/Library/Application Support/jcode/gemini.env` is exactly where `app_config_dir()` resolves on macOS, and `GEMINI_API_KEY` is the name the `gemini-api` profile reads. Your setup was correct.

**What I could not verify**

[...]

Issue #659, opened 2026-07-29
Fix typo in repository About description (effiecent → efficient)


doyonghoon:
## Summary

The repository About section currently contains a typo:

> "The most RAM **effiecent** harness"

`effiecent` should be `efficient`.

## Context

This is a documentation-only typo in the repository metadata (About description). It is visible on the repository homepage and in search results, so correcting it would improve professionalism and discoverability.

Current:

> The most RAM **effiecent** harness

Expected:

> The most RAM **efficient** harness

## Suggestion

Update the About description by replacing:

* `effiecent` → `efficient`

No code changes are required.

## Not urgent

This does not affect functionality and is purely a documentation polish item. Thank you for building and maintaining such an amazing tool!

---

*— Community contribution*

1jehuang:
Fixed, thanks for the sharp eye.

The About description now reads "The most RAM efficient harness". While checking, I found the same typo in a second place you could not see from the repo homepage: `README.md` had "happens to be very effiecent" in the Handterm scrolling note. That one is fixed on the triage branch, so the repo no longer contains the misspelling anywhere.

Verified with `gh repo view --json description` and a repo-wide search for the old spelling, which now returns nothing.

--- *— Jcode agent (automated triage), on behalf of @1jehuang*

Issue #657, opened 2026-07-29
Security preflight gate fails on tract-nnef advisory (was masked by the #654 test failure)


1jehuang:
## Summary

`Security preflight (Linux)` in `Build & Test (ubuntu-latest)` fails on a `cargo-audit` vulnerability in `tract-nnef 0.21.10`, plus several unmaintained/unsound warnings. It runs `scripts/security_preflight.sh --strict`, so the finding is a hard gate.

## Why this has not been noticed

This step had not actually executed on recent `master`. The job aborts at an earlier step, `Run TUI library tests (Linux only)`, which has failed on every push since `5c70ac4c3` (that is #654). Because `Security preflight` runs after it in the same job, the audit failure was masked by the test failure.

Fixing #654 in #656 lets the job reach the audit step, which then fails. Verified on [run 30435943571](https://github.com/1jehuang/jcode/actions/runs/30435943571): every test step passes and `Security preflight` is the only failing step. For contrast, on the last master run ([30413583952](https://github.com/1jehuang/jcode/actions/runs/30413583952)) that step is `skipped`.

So this is not a new regression, and it is not caused by #656. It is a pre-existing dependency issue that was hidden behind an unrelated red gate.

## The finding

One error-level advisory:

```
Crate: tract-nnef  Version: 0.21.10   error: 1 vulnerability found!
```

Warning-level (unmaintained/unsound): `bincode`, `paste`, `rustybuzz`, `ttf-parser`, `anyhow 1.0.100` (RUSTSEC-2026-0190), `lru 0.12.5` (RUSTSEC-2026-0002), `memmap2 0.8.0` (RUSTSEC-2026-0186), `rand 0.8.5` (RUSTSEC-2026-0097).

## Why I am not fixing it

This needs a maintainer decision rather than an agent's guess:

1. Bump `tract-nnef`, which may pull a breaking `tract` API change wherever it is used.
2. Add a scoped ignore to the audit config, accepting the risk with a documented rationale.
3. Drop or feature-gate the dependency if the functionality is optional.

Choosing between those depends on how the crate is used and what risk is acceptable, and the several warning-level advisories deserve a deliberate pass too. Picking one unilaterally would be the wrong call.

## Suggestion

Whatever the resolution, it is worth considering whether `Security preflight` should run in a job that is not gated behind the full test suite. A failure that can only surface once an unrelated test passes is a blind spot, and this issue is a concrete instance of it.

[...]

Issue #649, opened 2026-07-29
OpenRouter model_pricing deadlocks on cold catalog: holds read lock across fetch_models


paulbettner:
`OpenRouterProvider::model_pricing` can deadlock: it holds a read guard on
`models_cache` and then calls `fetch_models()`, which awaits a write guard on
that same lock.

In `crates/jcode-provider-openrouter-runtime/src/lib.rs`:

```rust
async fn model_pricing(&self, model_id: &str) -> Option<ModelPricing> {
    let cache = self.models_cache.read().await;   // read guard acquired
    if cache.fetched
        && let Some(model) = cache.models.iter().find(|m| m.id == model_id)
    {
        return Some(model.pricing.clone());
    }
    // ... disk cache attempt ...

    if let Ok(models) = self.fetch_models().await  // still holding `cache`
        && let Some(model) = models.iter().find(|m| m.id == model_id)
    {
        return Some(model.pricing.clone());
    }
    None
}
```

`cache` is still alive at the `fetch_models().await`, because it is never
dropped and the early returns above are the only paths that release it. And
`fetch_models` does:

```rust
pub async fn fetch_models(&self) -> Result<Vec<ModelInfo>> {
    // ...
    let cache = self.models_cache.read().await;
    // ...
    let mut cache = self.models_cache.write().await;   // waits forever
```

With `tokio::sync::RwLock`, a pending writer blocks, so the task waits on a
lock it is itself holding a reader for.

## When it triggers

Only on a cold catalog: `cache.fetched` false and no usable disk cache entry.
Both early-return paths release the guard, which is why this does not reproduce
once the catalog is warm.

It needs `model_catalog = true`, which is why I suspect it has gone unreported.

## Possible fix

Dropping the guard before the fetch is enough — release it explicitly, or scope
the cache read so it ends before `fetch_models()` is awaited.

## Notes

I have this change and a regression test running in a fork if a reference is
useful. Filing as an issue rather than a PR per CONTRIBUTING, since the
mechanism is visible from the source.

1jehuang:
Confirmed. The read guard is live at the `fetch_models().await`, `fetch_models` takes a write guard on the same `models_cache`, and a pending tokio writer blocks, so the task waits on a lock it holds a reader for. Your note that only a cold catalog reproduces it, because both early-return paths release the guard first, is exactly right.

Fixed in [#656](https://github.com/1jehuang/jcode/pull/656) by scoping the cache read to a block so the guard drops before the fetch.

[...]

PR #663, opened 2026-07-29
fix(ci): restore master CI - fmt, header auth fixture, clippy debt, ratchets, vacuous test (fixes #662, #654, #659)

### LINKED ISSUE #662: CI on master is fully red: fmt drift + failing header auth test, with clippy/audit/ratchet failures queued behind them
Every CI run on `master` has been failing for a while. The first failing step (rustfmt) masks a queue of further failures behind it, so the visible red is only part of the story. Full inventory, verified locally:

**Failing today:**
1. **rustfmt drift** — the key-to-paint commits (`ad7093d`, `8fc780b`) landed unformatted; `cargo fmt --all -- --check` rejects `app/observe.rs`, `tui/mod.rs`, `tui/ui.rs`. Fails both the `Format` job and `Quality Guardrails` step 1.
2. **`header_provider_auth_tag_prefers_app_resolved_credential_over_env`** — the test fixture claims `has_oauth + has_api_key` while leaving `ProviderAuth::state` at `NotConfigured` (an impossible probe result, since `state` is "best of available methods"). `build_auth_status_lines` filters on `state`, so it renders the nothing-configured fallback list (`○ anthropic ○ openrouter …`) instead of `anthropic(oauth+key*)`. The sibling test `auth_status_line_marks_active_credential_when_both_configured` sets `state: Available` on the same kind of fixture. Fails `Build & Test (ubuntu-latest)`.

**Queued behind those (each verified by running the gates on current master):**

3. **Clippy debt** — `cargo clippy --all-targets --all-features -- -D warnings` fails across seven crates. Some errors fire on Linux and would block CI (the orphaned `pinned_todos_*` fields on `App` left behind by `96a4a91`; the Windows command-line helpers in `jcode-terminal-launch` whose vestigial `test` cfg leaves them dead on unix test builds). The rest sit in macOS-only cfg blocks that ubuntu clippy never compiles but any Mac checkout trips over.
4. **Security preflight** — `cargo audit` now flags RUSTSEC-2026-0217 (`tract-nnef` 0.21.10). The patched tract line is unresolvable: `tract-data 0.21.16` pins `half = "=2.4.1"` against `naga`'s `half ^2.5` (desktop2 GPU stack), so this needs a triage entry per `docs/SECURITY_DEPENDENCIES.md` until the tract stack can move to 0.22/0.23.
5. **Stale ratchet baselines** — `check_code_size_budget`, `check_test_size_budget`, `check_panic_budget`, and `check_swallowed_error_budget` all fail on current master: files grew past their baselines while CI was stuck at the fmt step (desktop2, agent_tests, mermaid runtime, …).

## Fix

A complete fix branch is ready and validated (fmt, full-workspace clippy `-D warnings`, TUI lib suite with CI's exact skip flags, retention cohort, desktop2 frame budget, warning budget, all six ratchet scripts, `cargo machete`, `cargo metadata --locked`, `security_preflight.sh --strict` — all green on macOS/stable):

**https://github.com/1jehuang/jcode/compare/master...y4hyya:jcode:fix/ci-fmt-and-header-test**

Five separable commits: rustfmt pass, test-fixture fix, mechanical clippy sweep (no behavior changes), audit triage per your documented process, and the ratchet baseline refresh last so it can be dropped in favor of refactoring if you prefer.

Opening the PR from the fork currently fails with a permissions error (PR creation looks restricted to collaborators), so filing this issue per `require-issue.yml` and as the entry point — happy to open the PR the moment it's possible, or the branch can be pulled directly.

1jehuang:
Restores CI on `master`, which has been fully red. Fixes #662, #654, #659, and part of #660.

`master` failed at rustfmt, and because that is the first step in both the `Format` job and `Quality Guardrails`, every later gate was masked. Each one turned out to be independently broken. Verified locally by running CI's own commands.

| # | Problem | Fix |
| --- | --- | --- |
| 1 | `cargo fmt --all --check` rejected `app/observe.rs`, `tui/mod.rs`, `tui/ui.rs` (landed unformatted in `ad7093d`/`8fc780b`) | rustfmt pass, no behavior change |
| 2 | #654: `header_provider_auth_tag_prefers_app_resolved_credential_over_env` built a `ProviderAuth` with `has_oauth + has_api_key` but `state: NotConfigured`, which no real probe can return, so the render path filtered it and produced the nothing-configured fallback | fixture sets `Available`, matching the passing test beside it |
| 3 | Workspace clippy `-D warnings` failed on Linux | fixed at cause, see below |
| 4 | Four ratchet baselines stale (files grew while CI sat at the fmt step) | rebaselined in a separate final commit |
| 5 | `remote_reasoning_delta_burst_is_paced_not_dumped` fails on unmodified `master`, and was **not** in #662's inventory | pinned reasoning display, see below |

## Clippy findings fixed at cause, not silenced

- `pinned_todos_payload` / `_rendered_hash` / `_checked_at` were orphaned by `96a4a91ed`, which removed the trait override while the fields and both initializers stayed. Nothing read them; removed.
- `current_full_frame_redraw_reason` had no call site, but its doc comment explains exactly why it exists: `last_full_frame_reason` is sticky and keeps reporting a notice that expired minutes ago, making "why is this repainting at 60fps" undiagnosable. Now surfaced as `redraw_schedule.full_frame_reason` in `draw-stats`, which is what it was written for.
- Four copy-badge tests held a `CapturedClipboard` guard and never asserted through it, leaving `clipboard` unused and `CapturedClipboard::text` dead. They asserted only the badge and status notice, **both of which render fine when the clipboard write silently fails**. They now assert the copied text.
- `subscribe_nudge`'s test module sat before `impl App`, and asserted `0 + week`. Module moved to the end; the duplicated fresh-state assertion replaced with the clock-skew case (`now < last_shown`) that the `saturating_sub` exists to handle.
- Two `StatusSpinnerRenderer` fixtures assigned fields after `Default::default()`; now struct literals.

## Two things beyond the reported scope

**The copy-badge cohort is no longer skipped in CI.** It was excluded as needing a real desktop clipboard. Once all six tests route through the in-process sink (two had no guard at all and did reach for the OS clipboard), they pass headless, verified with `DISPLAY`, `WAYLAND_DISPLAY` and `XDG_SESSION_TYPE` unset. These were the tests most worth running, being the only end-to-end coverage of copy-badge to clipboard.

**A test that was passing vacuously now enforces its invariant.** `remote_reasoning_delta_burst_is_paced_not_dumped` fails on master because the reasoning-display default became `Off` in `166e4444f`: with it off the delta is dropped rather than buffered, so no pacing code runs at all. `742a3dc42` pinned the sibling tests via `with_reasoning_current_home` but missed this one. I confirmed the pinned test enforces real behavior by replacing the per-frame reveal clamp with "reveal the whole backlog", which makes it fail as intended.

## Also included

- #659: the About description typo is fixed on the repo, and the same `effiecent` typo in `README.md` is fixed here.
- #660 (partial): `jcode login` enables raw mode, which suppresses the terminal's echo, and then echoed nothing itself, so pasting a key produced zero screen change and looked like a hang. Now echoes `*` per character. Login also now reports an already-resolved key and its source, since the prompt was previously identical whether or not a credential was found. 5 new tests; the notice tests fail when the lookup is stubbed to its pre-fix behavior. `login.rs` **shrinks** by 12 lines, as it is already over the size budget, and the extracted `next_step` hint gained tests it never had.

## Verification

- `scripts/check_guardrails.sh` (full, including slow gates): **all gates pass** — fmt, `cargo check --all-targets --all-features`, clippy `-D warnings`, Cargo.lock, warning budget, all six ratchets, dependency boundaries, wildcard re-exports, desktop2 frame budget, `cargo machete`
- `cargo test -p jcode-tui --lib` with CI's exact skip flags: **2053 passed, 0 failed** (this is the gate that has been red)
- `cargo test -p jcode-app-core --lib retention_readiness`: 3 passed
- `cli::login::` cohort: 18 passed

## Not addressed

- #657 (`tract-nnef` RUSTSEC-2026-0217) is deliberately untouched. It is labeled `autonomous: no`, and the prior investigation established the `0.21.16` bump is blocked by `tract-data`'s exact `half =2.4.1` pin against `naga`'s `half ^2.5`. Choosing between a `tract 0.23` major bump, a scoped ignore, and feature-gating the embedding stack is a maintainer decision. `Security preflight` will therefore still fail on this branch. Everything else in front of it is now green, so that finding is finally visible rather than masked.
- #660's `run "Say hello"` hang is unconfirmed. I could not reproduce an indefinite hang on Linux, and `run` is a different path from `login`. I have asked the reporter for logs.


1jehuang:
**Merge-order note for the two open PRs blocked behind this one.**

Both #643 and #656 have been failing `Build & Test (ubuntu-latest)` and `Quality Guardrails` purely because of the master breakage this PR fixes, not because of anything in their own changes. I checked how each merges against this branch:

- **#643** (`agent/triage-fixes-20260728`, Copilot auth/model, TUI paste panic, Cursor regional host, musl): **merges cleanly**, no conflicts. It should just go green once this lands and it is rebased.
- **#656** (`triage/issues-2026-07-29`): **conflicts in two files**, both expected and both trivial:
  - `crates/jcode-tui/src/tui/ui_header.rs` — we both fix #654. The fixes are equivalent (`state: Available`), so take either side. Mine additionally comments why the original fixture was impossible.
  - `scripts/code_size_budget.json` — both refresh the baseline. Take either, then re-run `python3 scripts/check_code_size_budget.py --update`.

Suggested order: land this PR, rebase #643 (clean), then rebase #656 and resolve those two files.

Worth flagging that #656 also carries fixes for #655, #651, #648, #649 and #653, so it is the larger backlog unblock. It has an unskipped-tests change to `.github/workflows/ci.yml` too, which does not conflict with the `libfontconfig1-dev` step added here.

1jehuang:
**Rebased onto current master, and the scope grew.** Master moved while this was open (v0.62.1 plus the perf work and f59d04025), and two of those commits left master red on their own. Both are fixed here, since this PR's purpose is a green master.

**1. `JCODE_AUTO_POKE` was missing from `CONFIG_ENV_KEYS`.** f59d04025 added the env override for #664 but did not register it in the list the config cache fingerprints. That is not cosmetic: the fingerprint is what makes an env change take effect, so the new setting could be silently ignored until something else forced a reload. The repo already guards this, and `config_env_fingerprint_tracks_every_apply_env_override_var` **currently fails on master**:

```
CONFIG_ENV_KEYS must include every env var read by Config::apply_env_overrides;
missing: ["JCODE_AUTO_POKE"]
```

**2. The code-size baseline went stale again.** `tui/mod.rs`, `tui/ui.rs`, `ui_frame_metrics.rs`, `config-types/lib.rs` and `run_shell.rs` all grew past their entries in the perf commits. Refreshed; none of that growth is from this branch.

## Also folded in: the #664 session-level test

My separate #664 PR (#665) is closed as superseded by f59d04025, which is the better implementation because it also covers headless `jcode run`. But master's version tests only the config layer: that the field parses, and that the env var overrides it. **Neither would fail if `auto_poke_incomplete_todos` had stayed hardcoded `true`**, which was the actual bug. So the two behavioral tests moved here instead of being dropped:

- a new session honours `auto_poke = false`, and `/poke on` still re-enables it for that session
- auto-poke stays on when nothing disables it

Mutation-checked: restoring the hardcoded `true` makes the first fail.

That also brought over a latent leak in `with_temp_jcode_home`, which restored `JCODE_HOME` but never invalidated the config cache (keyed by content, not by home). A config written inside a temp home stayed cached for later tests in the same process, which is why writing one broke two unrelated auto-poke tests under the suite while both passed in isolation (refs #596).

Finally, the `StatusSpinnerRenderer` fixture is now a struct update rather than an exhaustive literal. The exhaustive form broke the moment master added `seeded_animation_area`, and a fixture asserting one field should not have to name every other one.

## Re-verified after the rebase

- `scripts/check_guardrails.sh`: all gates pass
- `cargo test -p jcode-tui --lib` with CI's exact skip flags: **2061 passed, 0 failed**
- auto-poke cohort: 22 passed; `run_shell::`: 12 passed; `config_env_fingerprint`: 2 passed

## Merge order

- **#666** (tract 0.23, clears the last `Security preflight` failure) is stacked on this branch and rebased on top of it.
- **#643** merged cleanly against this branch when I last checked; re-check after this lands since master has moved.
- **#656** conflicts in `ui_header.rs` (we both fix #654, equivalent fixes) and `scripts/code_size_budget.json` (both refresh the baseline). Take either side, then re-run `check_code_size_budget.py --update`.

1jehuang:
**Heads up: `master` is broken again, and by the same mechanism as `96a4a91ed`.** This blocks #663, #666, #643 and #656, so flagging it rather than working around it.

`221159294` ("desktop2: show sent versus acknowledged for user messages") declares two modules whose files were never committed:

```rust
// crates/jcode-desktop2/src/main.rs
mod frame_meter;      // line 20
mod scroll_profile;   // line 38
```

Neither `crates/jcode-desktop2/src/frame_meter.rs` nor `scroll_profile.rs` exists anywhere in history (`git log --all --diff-filter=A` finds nothing). The consequences on clean `origin/master`:

```
$ cargo fmt --all -- --check
Error writing files: failed to resolve mod `frame_meter`:
  crates/jcode-desktop2/src/frame_meter.rs does not exist

$ cargo check -p jcode-desktop2
error[E0583]: file not found for module `frame_meter`
error[E0583]: file not found for module `scroll_profile`
```

So the `Format` job and the `desktop2 frame budget` gate both fail on master again, which is exactly what #662 was about.

**This one cannot be fixed by deleting the `mod` lines**, unlike `96a4a91ed`. Master already has live consumers:

- `main.rs:143` — `frame_meter: frame_meter::FrameMeter`
- `main.rs:172` — `frame_meter::FrameMeter::from_env()`
- `cli.rs:648-649` — `scroll_profile::sweep()` / `scroll_profile::report(&costs)`

The two files **do** exist, untracked, in the `/home/jeremy/jcode` working tree (230 and 321 lines). I did not commit them, for two reasons. They are someone's in-flight work rather than mine to land, and the local copies have already moved past what master expects: dropping them in gives

```
error[E0061]: this method takes 1 argument but 2 arguments were supplied
  --> crates/jcode-desktop2/src/main.rs:1563:47
       state.render(&scene, &mut self.frame_meter)
```

so `render()` gained a parameter locally that master's call site does not have. Committing them blind would trade two errors for a different one.

**What is needed from you:** commit those two files together with the matching `render()` signature change, since only you know which local state is the intended one. Once that lands I will rebase #663 and #666 on top.

Worth noting for the recurrence, since this is the third instance (`c9ccb4f01`, `96a4a91ed`, now `221159294`): all three come from assembling a commit by copying files out of a worktree, which picks up some files and not their siblings. A pre-push hook running `cargo fmt --all -- --check` would catch this class in a second, because unresolvable `mod` declarations fail rustfmt before they fail the build.

Separately, #663 is otherwise green and now also fixes two things master broke while this PR was open:

- `JCODE_AUTO_POKE` was missing from `CONFIG_ENV_KEYS` after `f59d04025`, which makes `config_env_fingerprint_tracks_every_apply_env_override_var` fail on master.
- The code-size baseline went stale again from the perf and tui-style work.

1jehuang:
Added a gate for the recurrence itself, since this is now the third time an incomplete commit has taken master's Format job down.

`scripts/check_module_files.py` verifies that every `mod x;` in a tracked `.rs` file actually resolves to a file. It handles `#[path = "..."]` overrides, inline `mod x { ... }` blocks, and the non-`mod.rs` parent-directory form. It needs no compiler and runs in about 0.6s, and it is wired in **ahead of rustfmt** in both CI Format jobs and `check_guardrails.sh`.

Ordering is the point. Today the failure surfaces as:

```
Error writing files: failed to resolve mod `frame_meter`:
  crates/jcode-desktop2/src/frame_meter.rs does not exist
```

which reads like a formatting problem, and because rustfmt is step 1 of both `Format` and `Quality Guardrails`, that one omission makes every later gate unreachable. Exactly the masking dynamic this issue was filed about. Now the first line of output is `mod frame_meter; has no file`.

**Being straight about its scope.** This catches the `mod`-without-file variant only. I checked the two earlier instances of the same habit and it reports **clean** on both: `c9ccb4f01` and `96a4a91ed` landed a *reference* without its definition (`pinned_todos_payload_ref`, and the pinned-todos consumer), which only the compiler can catch. So this is not a general "incomplete commit" detector. It targets the one variant that disables rustfmt and therefore hides everything else, which is the variant worth a dedicated fast gate.

Verified it discriminates rather than just matching everything:

| Tree | Result |
| --- | --- |
| current `master` | fails, naming both missing files |
| `d9ca2a232` (v0.62.1 release) | clean |
| `c9ccb4f01` (1st instance) | clean |
| `96a4a91ed~1` (2nd instance) | clean |

Both #663 and #666 are rebased onto current master and pushed. They will still show the `frame_meter` / `scroll_profile` failure until those two files are committed, since that is a master problem I deliberately did not paper over. Everything else on #663 is green, and #666's `Security preflight` passes locally.

One more thing worth doing separately: `core.hooksPath` is set to `.githooks`, but that directory does not exist in the repo, so local hooks silently never run. Dropping this check plus `cargo fmt --all -- --check` into a real `.githooks/pre-push` would stop this class from reaching master at all, rather than catching it after the fact.

1jehuang:
Follow-up with the exact fix, since I verified it rather than guessing.

Master is still red on the missing `frame_meter` / `scroll_profile` modules. I confirmed the minimal fix is **three files** from the `/home/jeremy/jcode` working tree, not two:

```
crates/jcode-desktop2/src/frame_meter.rs      (untracked, 230 lines)
crates/jcode-desktop2/src/scroll_profile.rs   (untracked, 321 lines)
crates/jcode-desktop2/src/render.rs           (modified)
```

The third is the one that makes it work. Master's `main.rs:1563` already calls the two-argument form:

```rust
state.render(&scene, &mut self.frame_meter)
```

and the uncommitted `render.rs` is what provides that signature:

```rust
-    pub fn render(&mut self, scene: &Scene) -> Result<()> {
+    pub fn render(&mut self, scene: &Scene, meter: &mut crate::frame_meter::FrameMeter) -> Result<()> {
```

So master's call sites for `frame_meter` and `scroll_profile` were committed while all three definition-side files were left behind. With those three in place I confirmed locally:

- `cargo check -p jcode-desktop2`: clean
- `scripts/check_module_files.py`: `Module declarations OK`
- only `cargo fmt` remains, and purely mechanically: `render()`'s signature now exceeds the line limit and wants wrapping, plus one unrelated line in `tests/delivery.rs`. `cargo fmt --all` handles both.

**I did not commit them.** `scroll.rs` and `scroll_bench.rs` are also modified in that tree, the files were last written 13 minutes before I checked, and committing another agent's actively-edited work would be the same mistake that caused this. So: `git add` those three plus whatever else belongs with them, run `cargo fmt --all`, and master should go green.

Once it does, #663 and #666 need no changes; they are already rebased and will just need a re-run.

1jehuang:
**Third independent break on master, same class again.** Both PRs are rebased onto `d7d81d308` and pushed, but master now has two separate incomplete commits, not one.

**New: `jcode-app-core` does not compile.**

```
error[E0599]: no method named `resolve_invocation` found for struct
              `std::sync::Arc<SkillRegistry>`
   --> crates/jcode-app-core/src/agent/turn_execution.rs:777:46
777 |    if let Some(invocation) = skills.resolve_invocation(input) {
```

Reproduced on clean `origin/master` with `cargo check -p jcode-app-core`. A call site landed without the method it calls, so this is the same failure mode as `c9ccb4f01` and `96a4a91ed` rather than the `mod`-without-file variant. It kills the whole `jcode-tui --lib` gate, since that crate depends on it.

For the record, I checked whether my `git status` clue applied here too: the main worktree has `crates/jcode-base/src/skill.rs` modified and `crates/jcode-base/src/skill/invocation.rs` **deleted**, which is very likely where `resolve_invocation` lives or lived. But unlike the desktop2 case I could not verify a fix, because the definition is not sitting in the tree waiting to be committed; it appears to have been moved or removed mid-refactor. So this one needs whoever owns that change.

**Still open: the desktop2 modules.** Unchanged from my previous comment. The verified fix is three files (`frame_meter.rs`, `scroll_profile.rs`, and the modified `render.rs` that supplies the two-argument `render()` signature master already calls), then `cargo fmt --all`.

**Current state of master's gates:**

| Gate | Status | Cause |
| --- | --- | --- |
| `Format` (rustfmt) | red | missing desktop2 module files |
| `desktop2 frame budget` | red | same |
| `jcode-tui --lib` | red | `resolve_invocation` missing in jcode-app-core |
| `config_env_fingerprint` | red | `JCODE_AUTO_POKE` absent from `CONFIG_ENV_KEYS` (fixed on #663) |

#663 fixes the fourth and adds a gate that names the first two in 0.6s instead of as a confusing rustfmt error. The first three are other people's in-flight work, and I have deliberately not committed any of it.

This is now four instances of the same root cause in roughly a day (`c9ccb4f01`, `96a4a91ed`, `221159294`, and this one), all from assembling commits out of a worktree that has other agents' changes in it. The `core.hooksPath` point is worth repeating: it is set to `.githooks`, which does not exist, so local hooks silently never run. A real `.githooks/pre-push` running `cargo fmt --all -- --check`, `scripts/check_module_files.py`, and `cargo check --workspace` would have caught every one of these before push.

1jehuang:
**Correction: one of the two failures I attributed to master was mine.** Retracting that part of my previous comment.

I claimed `jcode-app-core` failed on master with `E0599: no method named resolve_invocation`. It does not. `origin/master` has both `mod invocation;` in `skill.rs` and `resolve_invocation` in `skill/invocation.rs`, and `cargo check -p jcode-app-core` is clean there. The breakage was on **my branch**, and the cause is the exact thing this PR is about:

My rustfmt commit was assembled from a worktree holding other agents' in-flight changes, so three files came along at a pre-refactor state and silently reverted work already on master:

| File | What my branch wrongly reverted |
| --- | --- |
| `crates/jcode-base/src/skill.rs` | Removed `mod invocation;` and re-inlined an older `SkillInvocation`, deleting the module that defines `resolve_invocation` (reverting 268913473 / 1c45117ca) |
| `crates/jcode-base/src/soft_interrupt_store_tests.rs` | Dropped the `images` field from three `SoftInterruptMessage` literals (reverting part of #627) |
| `crates/jcode-base/src/config.rs` | Master had already added `JCODE_AUTO_POKE` to `CONFIG_ENV_KEYS` in sorted position; mine duplicated it out of order |

All three are restored to `origin/master` verbatim in `1eba0a693`. `cargo check -p jcode-app-core` and `-p jcode-base` are clean, and `config_env_fingerprint` passes on master's own version, so my earlier commit for that was redundant too.

This is uncomfortable to write given that I filed the recurrence report and added a gate for it, but it is the honest record: I made the same mistake in the same session, and my own gate did not catch it because this variant is a stale-content revert, not a `mod` without a file. Only compiling the workspace catches it. That strengthens rather than weakens the `.githooks/pre-push` argument: `cargo check --workspace` in a hook is the check that would have caught all four instances, mine included.

**What is genuinely still master's, verified on clean `origin/master`:**

| Failure | Cause |
| --- | --- |
| `cargo fmt --all --check` | missing `crates/jcode-desktop2/src/frame_meter.rs` and `scroll_profile.rs` |
| `desktop2 frame budget` | same |
| `module declarations resolve` (new gate) | same |
| `test_configured_palette_recolors_a_real_rendered_frame` | fails on clean `origin/master`, unrelated to this PR |

That last one is new since my earlier comment and I have not touched it: `assertion left != right failed: configuring a color role should change the rendered frame`. It reproduces on `origin/master` alone, so it belongs to the tui-style work.

Both PRs are rebased on `d7d81d308` and pushed. With the three restorations, everything on #663 that is within its scope passes; the remaining red is the desktop2 files plus that palette test.

1jehuang:
Closing the verification loop properly, since my earlier "verified" claims were thinner than they sounded. Two gaps found and fixed, plus an audit that shows this branch is now clean.

## The #660 masking had no test at all

`read_secret_line` only masks on the raw-mode TTY branch. Under `cargo test` stdin is not a terminal, so **every possible unit test takes the piped-line fallback and exercises none of the fix**. The reporter's actual symptom was unguarded.

Fixed with a real pty test (`secret_input_pty_tests.rs`): forks a child running the function with a pty slave as stdin, types at it, and asserts on the bytes the terminal received. Mutation-verified both ways, and confirmed end to end against built binaries:

| Binary | Bytes echoed while typing 15 chars |
| --- | --- |
| pre-fix | `''` — **zero**, which is precisely "hangs indefinitely" |
| fixed | `'***************'` — 15 masks, key saved intact |

The test also pins that the secret never reaches the screen, that masking does not corrupt the returned value, and that backspace erases one mask and one character.

## The extracted hint strings were not pinned

Moving the next-step hints into `next_step` turned three `eprintln!` literals into `\`-continued strings. A continuation swallows the following indentation, so a mis-indented line changes user-facing text while still compiling, and nothing would have caught it. Added exact-equality assertions against the strings `login.rs` printed before extraction, so the refactor is provably invisible to users.

## Audit of this branch against master

After the stale-file incident I checked mechanically rather than by eye.

**Every line this branch removes from master** (210 of them): all accounted for as intentional edits or baseline numbers. `rate_limit_notice_with_nudge` was moved, not deleted, and its caller in `server_events.rs:1254` still resolves.

**Every baseline I raised** (26 code-size, 5 test-size), attributed by comparing against master's own file lengths:

- **29 of 31** match master's current content exactly, i.e. master's growth, not mine.
- 2 apparent exceptions are files this branch makes **smaller** than master: `tui/app.rs` 2516 → 2508 and `tui_lifecycle.rs` 1302 → 1296, from removing the orphaned `pinned_todos_*` fields.
- The one genuine increase of mine is `scroll_copy_02/part_01.rs` (+23), and all 23 lines are the clipboard assertions those tests were missing.

So no gate was loosened to accommodate my own code.

## Current test state

- `cargo test -p jcode-tui --lib` (CI's skip flags, plus master's broken palette test): **2074 passed, 0 failed**
- `cargo test -p jcode-base --lib`: **1195 passed, 0 failed**
- `security_preflight.sh --strict` on #666: passes

Still red and still not mine, both reproduced on clean `origin/master`: the missing `frame_meter.rs` / `scroll_profile.rs`, and `test_configured_palette_recolors_a_real_rendered_frame`.

1jehuang:
Caught the obvious failure mode in my own two new tests before claiming them: **neither would have run in CI.**

## Both new tests were dead on arrival

| Test | Why it would never execute |
| --- | --- |
| `secret_input_pty_tests` (#660) | `jcode-base --lib` is not run anywhere on Linux CI. The only jcode-base test invocation in the whole workflow is a Windows-only `power_inhibit::tests::windows_` filter. |
| `minilm_embedding_is_numerically_stable...` (#657) | `jcode-embedding` tests are not run at all, **and** the test skips itself when the model is absent, which it always is on a runner. |

That is the identical shape as the two vacuous gates this PR already fixes: the warning budget passing because `rg` was missing, and the macOS stdin detector in #651 that no job ever executed. I would have added two more tests that looked like coverage and provided none.

Both now have cohorts. Details worth flagging:

**The pty cohort is bounded so it cannot hang.** These tests exist because `read_secret_line` can *appear to hang*; a test for a hang that hangs would burn the 300s CI timeout instead of failing. The forked child now calls `alarm(20)`. Verified by making the function block forever behind an env guard: the cohort fails in **6 seconds** with `got 0 for 15 chars` rather than hanging.

**The embedding cohort fetches the model and asserts it did not skip.** ~87MB, two curls. The guard matters more than the run, and I checked which half actually proves anything against a model-less `HOME`:

- `grep "... ok"` — **still passes on a skip**, because a skipped test reports `ok`. It only proves the binary built and the harness ran it.
- `grep "skip: MiniLM model not present"` — **correctly fails the step**. This is the real check.

The comments in `ci.yml` now say which check proves what, rather than implying the first one is sufficient. That distinction is the whole lesson of #662, and I nearly reintroduced it.

## Robustness checks on the pty cohort

- passes with **no controlling terminal** (`setsid`), which is the CI shape
- passes under the **default parallel harness**, not just `--test-threads=1`
- **10 consecutive runs**, zero flakes
- clippy `-D warnings` clean on jcode-base; test-size, panic and swallowed-error ratchets all unchanged
- `ci.yml` parses as valid YAML

Both branches are rebased and pushed. On #666 the embedding cohort passes on tract 0.23.4 against reference values pinned on 0.21.10, so the numeric guarantee is now enforced by CI rather than by my say-so.

Issue #596, opened 2026-07-25
TUI lib tests share process-global state, so they only pass serially


1jehuang:
Follow-up to #592, which enabled the TUI library tests in CI. That gate has to run with `--test-threads=1` because the suite is not parallel-safe.

## Evidence

- `cargo test -p jcode-tui --lib -- --test-threads=1` → 1977 passed, 0 failed.
- Default (parallel) → 2-4 failures per run, **a different set each run**, and every failing test passes when run alone.

Observed failing across runs: `overscroll_reveal_does_not_relayout_transcript`, `test_remote_mixed_catalog_keeps_jcode_subscription_separate_from_other_providers`, `test_tui_cerebras_paste_key_lifecycle_has_no_degraded_success_messages`, `test_save_and_restore_reload_state_preserves_split_view_mode`, `test_tui_openai_compatible_empty_catalog_does_not_switch_to_profile_default`, `test_changelog_overlay_repeated_renders_are_stable`, `gather_ambient_info_filters_to_session_reminders_when_ambient_disabled`.

## Contended state

- the process-global model catalog / route memo
- `AMBIENT_INFO_CACHE`, whose stale-while-revalidate path spawns a background thread that reloads under whatever `JCODE_HOME` is current when it runs, not when it was queued
- render state (partly addressed already by the shared `render_state_test_lock`)
- `JCODE_HOME` itself, when a test sets it without holding the shared env lock

`helpers_tests.rs` already documents this: an "unrelated in-flight cache refresh can otherwise overwrite the cleared process-global cache with data loaded under another test's JCODE_HOME".

## What worked in `jcode-base`

Two tests with exactly this shape were fixed by scoping, not redesign:

- wrap the body in `with_clean_provider_test_env`
- hold the shared test-env lock plus a private `JCODE_HOME` (matching the neighbouring test that already did)

`cargo test -p jcode-base` went from 1-2 intermittent failures per run to 1130 passing across three consecutive runs.

## Why it is worth doing

1. Ordering-dependent tests hide real regressions: a genuine break looks like the usual flake.
2. Serial costs ~40s in CI where parallel would be ~15s.
3. The `--skip` list in the CI step is blunt. Six `copy_badge_shortcut` tests are skipped because they drive a real clipboard; if they were injectable behind a clipboard trait they could run headlessly instead of being excluded.

## Suggested order

1. Scope the catalog/ambient tests the way the `jcode-base` ones were.
[...]

PR #665, opened 2026-07-30
feat(config): persistent features.auto_poke setting (fixes #664)

### LINKED ISSUE #664: Feature Request: Add a persistent config option to disable auto-poke
First, thank you for building and maintaining Jcode. It is a genuinely useful product, and I appreciate the amount of work that has gone into the agent workflow and TUI.

I would like to request a persistent configuration option for disabling auto-poke.

At the moment, I can run:

```text
/poke off
```

However, this only disables auto-poke for the current running instance. Since auto-poke is enabled by default again when Jcode starts or reloads, I have to remember to disable it repeatedly.

This can be disruptive during normal conversations with the agent. I may be discussing the task, providing additional context, or preparing to redirect it, while the harness continues sending automatic follow-up messages because there are incomplete todos. In some cases, several poke messages can be sent before I notice what is happening. This happens pretty often when I've being switching conversations between multiple agents, i would ask a question, switch to another agent, come back and see that auto poke has triggered multiple times. Sometimes the agent would refuse to continue (which is nice), but sometime the agent would be mistaken and assume my authorization for resuming, and implement stuff that is not properly planned out yet.

It would be helpful to support a persistent setting such as:

```toml
[features]
auto_poke = false
```

Ideally, the behavior would be:

* The configured value determines the default when Jcode starts.
* `/poke on` and `/poke off` can still override it for the current session.
* Users who like the current behavior can keep auto-poke enabled by default.

Thanks again for the product and for considering this request.

1jehuang:
Implements the persistent auto-poke setting requested in #664, using the exact shape the reporter proposed.

**Stacked on #663.** This branch is based on the CI-restoration branch, because master's CI is fully red and nothing can be verified against it otherwise. Merge #663 first.

## The change

`features.auto_poke` (default `true`) sets whether auto-poke starts enabled in a new session:

```toml
[features]
auto_poke = false
```

The behavior matches the request exactly:

- The configured value determines the default when jcode starts.
- `/poke on` and `/poke off` still override it for the current session.
- Users who like the current behavior get no change, since the default is `true`.

`JCODE_AUTO_POKE` works as an env override, consistent with every other `[features]` flag, and the setting is documented in the shipped config template and shown in `/config`.

The actual defect was small: `auto_poke_incomplete_todos` was initialized to a hardcoded `true` in two places, so no config value could have reached it.

## Why `FeatureConfig` moved

`crates/jcode-config-types/src/lib.rs` is already over the code-size budget, so adding a field there is precisely the growth the ratchet exists to catch. Rather than refresh the baseline, I extracted `FeatureConfig` into its own `features` module. No baseline in this PR needed updating.

## Two problems found while testing

- **A latent state leak in the shared test helper.** `with_temp_jcode_home` restored `JCODE_HOME` on exit but never invalidated the config cache, which is keyed by content rather than by home. A config written inside a temp home therefore stayed cached for subsequent tests in the same process. This is why writing a config in one test made `test_auto_poke_starts_enabled_by_default` and `test_ctrl_p_toggles_auto_poke_locally` fail under the suite while passing in isolation. Now invalidated on both entry and exit (refs #596).
- **An existing gate did its job.** `config_env_fingerprint_tracks_every_apply_env_override_var` failed with `missing: ["JCODE_AUTO_POKE"]`, catching that the new variable was absent from `CONFIG_ENV_KEYS`. Without that entry, changing the env var would not have invalidated the config cache. Fixed.

## Verification

- 5 new tests: default/parse, env override, shipped-template round-trip (parses the real generated template rather than matching a substring), session default, and `/poke on` still overriding.
- The session-default test is **mutation-checked**: restoring the hardcoded `true` makes it fail, so it detects the actual bug rather than only restating the config value.
- `scripts/check_guardrails.sh` (full): all gates pass, no baseline changes.
- `cargo test -p jcode-tui --lib` with CI's skip flags: 2058 passed, 0 failed.
- `cargo test -p jcode-base --lib config::`: 70 passed.

## Note

The default stays `true`, so this is purely opt-in. If you would rather auto-poke default to off given the reporter's description of it acting on unplanned work, that is a product call and a one-line change.


1jehuang:
Closing as superseded: @1jehuang landed the same feature directly on master in f59d04025 while this was in review. Same design, same `[features] auto_poke` key, same `/poke on|off` override semantics, and it additionally covers the headless `jcode run` path that I had missed.

I have moved the two pieces of this branch that master does not have onto #663 rather than dropping them:

1. **A test that the session default actually works.** Master's version tests the config layer only (the field parses, `JCODE_AUTO_POKE` overrides it). Neither of those would fail if `auto_poke_incomplete_todos` had stayed hardcoded `true`, which was the real bug. The ported test asserts a new session honours `auto_poke = false` and that `/poke on` still overrides, and it is mutation-checked: restoring the hardcoded `true` makes it fail.

2. **The `with_temp_jcode_home` config-cache leak.** The helper restored `JCODE_HOME` but never invalidated the config cache, which is keyed by content rather than by home, so a config written inside a temp home stayed cached for later tests in the same process.

Also worth flagging, and now fixed on #663: f59d04025 did not add `JCODE_AUTO_POKE` to `CONFIG_ENV_KEYS`. That is not cosmetic, since the cache fingerprint is what makes an env change take effect, and `config_env_fingerprint_tracks_every_apply_env_override_var` is currently **failing on master** with `missing: ["JCODE_AUTO_POKE"]`. Good gate, and it caught the same omission in my own version earlier.

No objection to master's implementation winning; it is the better one for covering `jcode run`.

PR #666, opened 2026-07-30
fix(deps): tract 0.23 to clear RUSTSEC-2026-0217 (fixes #657)

### LINKED ISSUE #657: Security preflight gate fails on tract-nnef advisory (was masked by the #654 test failure)

1jehuang:
Clears RUSTSEC-2026-0217, the last remaining CI failure once #663 lands.

**Stacked on #663.** Merge that first.

## Why 0.23 rather than a patch bump or an ignore

The prior investigation on #657 established that the in-line fix is unreachable, and I confirmed it: `tract-data 0.21.16` and `0.21.17` pin `half =2.4.1` **exactly**, while `naga` (via `vello` in `jcode-desktop2`) requires `half ^2.5`. Neither direction resolves without dropping the desktop2 GPU stack. The 0.23 line drops that pin.

I went with the upgrade rather than a scoped ignore because the exposure is not the comfortable "we only parse bytes we shipped" case. `load_from_dir` downloads `model.onnx` from huggingface.co when it is absent and then parses the fetched artifact, so the vulnerable NNEF parser does run over data retrieved at runtime.

## The bump is mechanical

`jcode-embedding` is the only consumer. Three renames, all compiler-directed:

- `SimplePlan<TypedFact, Box<dyn TypedOp>, Graph<..>>` becomes `Arc<RunnableModel<TypedFact, Box<dyn TypedOp>>>`. 0.23 dropped `SimplePlan`'s third (model) type parameter, and `into_runnable` now returns the `Arc<RunnableModel<..>>` alias directly.
- `TValue::to_array_view` becomes `to_plain_array_view`, 2 call sites.

Diff is 3 files: the version bump, those renames, and `Cargo.lock`.

## Verification

- **`scripts/security_preflight.sh --strict` passes.** `cargo tree -i tract-nnef` reports `0.23.4` and the advisory is gone. Only the 11 pre-existing allowed warnings remain.
- **Inference is correct, not merely compiling.** This matters more than the compile: a silently-wrong embedding would pass a build. The local MiniLM model happens to be present on this host, so `minilm_related_beats_unrelated_if_present` **actually executed** rather than taking its skip path, and its cosine-similarity assertion against the real model still holds. `cargo test -p jcode-embedding` gives 5 passed, and the `--ignored` latency probe runs real inference end to end (model load 164ms, 380ms per embed).
- Full `scripts/check_guardrails.sh`: all gates pass, no baseline changes.
- `cargo metadata --locked` is clean.

`docs/SECURITY_DEPENDENCIES.md` records the resolution and the `half` conflict, so nobody repeats the 0.21.16 attempt.

## Worth your judgement

This is a **major version bump of an inference dependency**, which is why #657 was labeled `autonomous: no`. I am putting it up because it turned out to be three renames rather than a real migration, and because the alternative was an ignore over a runtime-downloaded artifact. But the semantic check that gives me confidence depends on a locally-present model, and that test **skips silently** on a machine without it, including CI. So the quality evidence is from my machine, not from the pipeline. If you would rather ship the ignore and schedule the upgrade deliberately, that is entirely reasonable and this branch can just be closed.

Separately: that test skipping rather than failing when the model is absent is the same "gate that looks green because it never ran" pattern as the warning budget in #663. Probably worth making the model a CI fixture, but that is out of scope here.


1jehuang:
Ran the check that actually mattered for this bump and had not been done: **do tract 0.21 and 0.23 produce the same embedding numbers?**

This is not covered by "it compiles" or by `minilm_related_beats_unrelated_if_present`. Persisted memory embeddings are keyed by `model_id`, which stays `all-MiniLM-L6-v2` across a tract bump, so the model-mismatch filter in `memory.rs:854` cannot notice a numeric change. If 0.23 embedded text differently, every stored vector would silently become incomparable with freshly computed ones, and the related-vs-unrelated assertion would still pass because both vectors move together.

## Measured, on both versions, same model file

Ran an identical probe on `fix/issue-662-ci-red` (tract 0.21.10) and this branch (0.23.4):

```
text="how do I set the cargo build profile"
  dim=384 norm=1.000000
  head=[0.040730, 0.103227, 0.065710, 0.032524, -0.012075, 0.005709, 0.029490, -0.038528]
```

Across all three probe texts and 8 reported dimensions, exactly **one** value differs:

```
0.21.10:  ... 0.028180 ...
0.23.4:   ... 0.028181 ...
```

A single 1e-6 delta, i.e. float32 rounding. Worst case if every one of the 384 dimensions drifted that much: `|error| ≈ 2.0e-5`, cosine deviation from 1.0 of about **1.9e-10**. That is roughly 10^8 times smaller than any retrieval threshold in the codebase. **Existing persisted embeddings stay compatible**; no re-embedding is needed.

## Landed as a test, calibrated by mutation

`minilm_embedding_is_numerically_stable_across_inference_engines` pins dimensionality, L2 norm, and the first 8 components against the 0.21.10 reference at 1e-4 tolerance. Verified it is neither vacuous nor over-tight:

| Condition | Result |
| --- | --- |
| tract 0.21.10 | passes |
| tract 0.23.4 | passes |
| non-uniform perturbation injected into the pooled output | **fails**: `dim 0 drifted beyond float32 rounding: diff 8.3e-4` |

So it tolerates the real rounding and catches a change three orders of magnitude smaller than one that would actually degrade retrieval. It sits on #663 rather than here, so the reference values are pinned on the pre-bump engine and this PR has to satisfy them.

One honest limitation, unchanged: like its neighbours, the test skips when the MiniLM model is absent, which includes CI. Making the model a CI fixture is the fix, and is worth doing separately.

Issue #695, opened 2026-08-01
Reset todos window to the beginning when doing another task, after the task is completed


erelst:
<img width="1286" height="419" alt="Image" src="https://github.com/user-attachments/assets/3887f153-6c0a-4a27-a5e9-d10c60c1b0f0" />

In the same session, the "todos" or window on the right in the screenshot (which contains confidence scores, etc.) doesn't change at all even if I perform other tasks. If the previous task is completed, shouldn't those todos be replaced with the todos from the current task? I've done many different tasks and sometimes I see in thinking/reasoning that there are actually new todos running, but the todos window on the right still displays the ones from the initial task of the session. Actually I'm confused whether this is a feature request or a bug, please check it.

1jehuang:
It was a bug, thanks for flagging it. The goal assessments (confidence panel) are deliberately retained when a todo write does not mention them, so a partial update cannot erase them, but nothing pruned goals whose todo group had already disappeared. Goals from your first task therefore stayed visible after the list moved on.

Fixed in #701: goals whose group no longer exists in the current todo list are dropped, while goals-only writes and flat (ungrouped) lists are unaffected. If you still see stale entries after that lands, please reopen with a screenshot.

--- *— Jcode agent (automated triage), on behalf of @1jehuang*

1jehuang:
Update: the fix is merged to master (#701). Follow-up end-to-end tests are in #702. It will ship in the next release.

--- *— Jcode agent (automated triage), on behalf of @1jehuang*

1jehuang:
One more finding while verifying this, worth flagging because it can produce the same symptom through a different path.

There is a second mechanism: when a single todo write both creates and completes a group without an `end_to_end_ownership` score, the ownership quality gate rejects the entire write and the previously stored todo list is kept. I confirmed this against the real tool: after such a write the stored todos still describe the *previous* task, so the panel looks frozen even though the agent moved on.

That gate is deliberate design (it is what forces the agent to validate before declaring a group done), so I have not changed it. Leaving it for @1jehuang to decide whether a rejected write should still be visible somehow, for example by surfacing the rejection in the panel instead of silently keeping stale content.

The goal-retention half of this issue is fixed and merged.

--- *— Jcode agent (automated triage), on behalf of @1jehuang*

Issue #694, opened 2026-08-01
Bug Report: Custom `openai-compatible` models incorrectly tagged with `copilot:` prefix in model picker


rahulchirumamilla:
# Bug Report: Custom `openai-compatible` models incorrectly tagged with `copilot:` prefix in model picker

## Summary
When using custom `openai-compatible` providers (e.g., local MLX servers via the `omlx` profile), the `/model` picker incorrectly labels the models with `PROVIDER = Copilot` and constructs internal model IDs with the `copilot:` prefix (e.g., `copilot:KAT-Coder-V2.5-Dev-OptiQ-4bit`). This causes "model not found" errors when selecting the model.

## Reproduction
1. Configure a custom provider in `~/.jcode/config.toml`:
   ```toml
   [providers.omlx]
   type = "openai-compatible"
   base_url = "http://127.0.0.1:18000/v1"
   default_model = "KAT-Coder-V2.5-Dev-OptiQ-4bit"
   ```
2. Start jcode with the profile: `jcode --provider-profile omlx`
3. Press `/model` inside the TUI.
4. Observe that the model table shows `Copilot` in the PROVIDER column and the status bar shows `Copilot:KAT Coder V2.5 Dev OptiQ 4bit`.
5. Selecting the model results in: `Error: Model copilot:KAT-Coder-V2.5-Dev-OptiQ-4bit not found`.

## Expected Behavior
Models from custom `openai-compatible` profiles should be tagged with their configured profile name (`omlx`) or the generic `openai-compatible` label in the picker, and the model ID should be passed cleanly to the runtime without the `copilot:` prefix.

## Actual Behavior
The picker merges custom profile models into a catalog that incorrectly inherits the `copilot` provider label from `ProviderChoice::Copilot`.

## Code Location
- `src/cli/provider_init.rs`: Defines `ProviderChoice::Copilot => "copilot"` and handles `OpenaiCompatible` registration via `resolve_openai_compatible_profile`.
- `src/cli/provider_init.rs:1811`: `init_provider_and_registry` likely performs the catalog merge.
- `crates/jcode-tui/src/tui/info_widget_model.rs` (or related TUI picker components): Renders the table with the incorrect PROVIDER column.

## Environment
- jcode v0.64.2 (6c6fbba1c)
- macOS
- Custom `openai-compatible` profile pointing to local MLX server.

1jehuang:
[...]

PR #702, opened 2026-08-01
test: end-to-end coverage for the #694/#695 triage fixes, and fix a pre-existing red test

1jehuang:
Follow-up to #701, which merged while these were still being written.

The three fixes in #701 landed with unit tests. This adds the end-to-end coverage that exercises the real user-visible paths, plus a fix for one pre-existing red test.

### End-to-end regression guards

- **#695**: drives the real `TodoTool` through the reporter's scenario. Complete task one, start task two, then assert the stored todos *and* goals describe task two only. This is what the todos panel renders, so it covers the reported symptom rather than just the helper.
- **#694**: writes a real config.toml containing the reporter's `omlx` profile, then asserts the full picker route set (`remote_model_routes_fallback`) contains the `openai-compatible:omlx` route and no copilot route at all.

Both tests were confirmed to fail when their fix is temporarily reverted, so they are real regression guards rather than tests shaped to match the implementation.

### Pre-existing red test

`schema_advertises_intent_and_todos` asserted a lowercase `requirement-to-check` against schema text that starts the sentence with `Requirement-to-check`, so it failed on master independently of any of this work. Now compared case-insensitively.

### Verification

`cargo test -p jcode-base --lib` (1214 passed), `-p jcode-app-core --lib tool::todo` (38 passed), `-p jcode-tui --lib` serially (2117 passed), clippy clean for the touched crates. Size ratchets rebaselined only for the lines this branch adds.


1jehuang:
Added live verification against a real running binary (not just unit tests), using the debug socket to inject keys into an actual TUI client on this branch's build:

- Input `hello`, cursor at end, Ctrl+D: input stays `hello`, process does not exit.
- Cursor moved left twice, Ctrl+D: input becomes `helo` with the cursor held at position 3.
- Empty input, two rapid Ctrl+D presses: the client exits, so the intentional quit path is preserved.

Also extended coverage since the first push: the remote key handler (the path the normal `jcode` TUI actually uses, since it runs as a client with `is_remote = true`), both model-route sources for #694, and the stale ungrouped goal that the todos panel renders unconditionally for #695.

Issue #726, opened 2026-08-02
selfdev build fails with exit 101 when clang is absent: dev_cargo.sh forces clang even in system-linker fallback mode


MatrixMagician:
## Problem

`selfdev build` fails with exit code 101 on any Linux machine that does not have
`clang` installed, even though the script explicitly reports that it is falling
back to the system linker.

```
$ selfdev build target=tui
Build did not complete successfully (state: failed)
Reason: Command exited with code 101
```

Running the underlying command directly shows the contradiction:

```
$ scripts/dev_cargo.sh build --profile selfdev -p jcode --bin jcode
dev_cargo: no supported fast linker detected; using system linker settings
error: linker `clang` not found
  |
  = note: No such file or directory (os error 2)

error: could not compile `libc` (build script) due to 1 previous error
error: could not compile `serde` (build script) due to 1 previous error
error: could not compile `proc-macro2` (build script) due to 1 previous error
... (every build script in the graph fails identically)
```

The log line and the behaviour disagree: it says "using system linker settings"
and then tells cargo to invoke a linker that is not present.

## Root cause

In `configure_linux_linker` (`scripts/dev_cargo.sh`), the linker export happens
*before* the mode dispatch, so it applies to all three modes including `system`:

```sh
selected_linker_mode="$mode"
export CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER="${CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER:-clang}"

case "$mode" in
  lld)    append_rustflags "-C link-arg=-fuse-ld=lld"  ;;
  mold)   append_rustflags "-C link-arg=-fuse-ld=mold" ;;
  system) # nothing appended, but clang is already exported above
    log "no supported fast linker detected; using system linker settings" ;;
esac
```

The `auto` branch correctly requires `clang` to be present before selecting
`mold` or `lld`, and correctly falls back to `system` when it is missing. But
that fallback is then undone by the unconditional export. Only `mold` and `lld`
need clang as the linker *driver*; `system` mode should leave the choice to
cargo's default (`cc`).

This affects an explicit `JCODE_FAST_LINKER=system` too, not just the
auto-detected fallback, so there is no way to build through this script without
clang installed.

## Reproduction

On a Linux machine with a working Rust toolchain and `cc`, but no `clang`
(verified on the commit before the fix):

```sh
$ command -v clang || echo "no clang"
no clang

[...]

Issue #724, opened 2026-08-02
jcode subscription catalog omits live claude-opus-5


caipira113:
## Summary

The public jcode router catalog advertises `claude-opus-5` for every paid tier, but jcode v0.65.0 and the current `master` subscription catalog omit it. As a result, paid jcode subscribers cannot select a model that the router says they are entitled to use.

Explicitly passing `--model claude-opus-5` does not report that the model is unsupported. It silently selects `claude-opus-4-8` instead.

## Reproduction

Environment:

- jcode v0.65.0 (`d847d62e4`)
- macOS arm64
- jcode account tier: Ultra (active)

List the jcode subscription models:

```bash
jcode --provider jcode model list --json | jq '{
  selected_model,
  count: (.models | length),
  opus_models: [.models[] | select(test("opus"; "i"))]
}'
```

Actual result:

```json
{
  "selected_model": "claude-opus-4-8",
  "count": 19,
  "opus_models": [
    "claude-opus-4-8"
  ]
}
```

Compare that with the public router catalog:

```bash
curl -fsSL https://api.solosystems.dev/v1/models | jq '{
  count: (.data | length),
  opus_models: [
    .data[]
    | select(.id | test("opus"; "i"))
    | {id, tiers}
  ]
}'
```

The router currently returns 20 models, including:

```json
{
  "id": "claude-opus-5",
  "tiers": [
    "plus",
    "pro",
    "max",
    "ultra",
    "flagship"
  ]
}
```

Trying to select the missing model explicitly:

```bash
jcode --provider jcode --model claude-opus-5 model list --json   | jq '{selected_model}'
```

Actual result:

```json
{
  "selected_model": "claude-opus-4-8"
}
```

## Expected behavior

The jcode subscription model catalog should expose `claude-opus-5` when the live router catalog authorizes it for the account tier.

If a requested model is not present in the client catalog, jcode should report that explicitly instead of silently running a different model.

## Current source state

The current `master` HEAD checked during reproduction was `8f614c600006a996301cda7dd08f637b5537b182`.

`crates/jcode-base/src/subscription_catalog.rs`:

- contains `claude-opus-4-8`
- does not contain `claude-opus-5`
- asserts `CURATED_MODELS.len() == 19`

The pricing page states that its live model list comes directly from the same router that enforces plan access:

https://jcode.sh/pricing

## Impact

- Subscribers cannot use a model that the router lists as included in their plan.
- The pricing/router catalog and CLI picker disagree.
[...]

Issue #722, opened 2026-08-02
Bash tool schema doesn't expose `justification` field, so models can't discover the Confirm-tier bypass


rameshbaskaran:
## Summary

`jcode-command-risk`'s destructive-command gate (`bash_destructive_gate.rs` / `gate.rs`) accepts a `justification` field on the `Bash` tool call that can unlock `Confirm`-tier findings (e.g. glob-pattern targets like `find ~/.foo -iname "*bar*"`, computed/runtime paths) when it's substantive (25+ chars, not a bare "yes"/"ok"). This is real, working, intentional escape-hatch infrastructure related to the #604 fix.

The problem: **the `Bash` tool's user/model-facing schema never mentions `justification`.** The advertised parameters are only `command`, `run_in_background`, and `timeout`. A model has no way to discover this field exists except by reading `jcode-command-risk`'s Rust source directly (`bash_parameters_schema()`), which most sessions never do.

## Impact

Every session independently rediscovers (or fails to discover) that read-only commands with globs — `find`, `grep`, `ls` with wildcard targets, etc. — get hard-blocked with a "This command was not run. It is irreversible" message that reads like a false positive for genuinely non-destructive commands. Models tend to work around it (switch to Node REPL, avoid globs, ask the user) rather than use the actual designed escape hatch, because the escape hatch is invisible.

This was traced end-to-end in a live session: confirmed via `strings` on the running binary that the gate logic and messaging match `crates/jcode-command-risk/src/{lib,paths,gate}.rs` exactly, then empirically verified that adding `justification: "<substantive reason>"` to a blocked `find ... -iname "*config*"` call on the same attempt (no retry needed) immediately allows it through.

## Suggested fix

Add `justification` (optional string) to the advertised `Bash` tool schema/description, with a short note on when it's needed (Confirm-tier findings: globs, computed paths, truncating redirects) and what "substantive" means (not a bare "yes"/"proceed", should reference what specific request it serves). This lets models supply it proactively on the same call instead of hitting the block, reading the error text, and guessing.

Optionally: have the block's own error message name the field explicitly (it currently says "re-issue the same call with a `justification` field" — this part is correct and is how the field was discovered here — but the tool schema itself should already list it as a first-class documented parameter, not something surfaced only in an error string after a failure).

## Note on Catastrophic-tier

[...]

Issue #719, opened 2026-08-02
fix(niri): setup-hotkey injects global binds into recent-windows


tingfeng347:
## Summary

`jcode setup-hotkey` inserts global niri launch bindings into the first textual `binds {` block in `config.kdl`. When that block belongs to `recent-windows`, niri rejects the injected `{ spawn ... }` actions and the entire configuration fails to load.

## Reproduction

Use a split niri configuration whose main file contains a nested `recent-windows.binds` block before an included global bindings file:

```kdl
recent-windows {
    binds {
        Mod+Tab { next-window; }
        Mod+Shift+Tab { previous-window; }
    }
}

include "binds.kdl"
```

Run:

```console
jcode setup-hotkey
niri validate
```

## Actual result

The managed jcode region is inserted into `recent-windows.binds`. niri reports that it expected `next-window` or `previous-window` when parsing the injected `spawn` action.

## Expected result

Global launch bindings should only be inserted into a top-level `binds` node. If the main file has no top-level `binds` node, jcode should append one rather than using a nested compositor-specific `binds` node.

Re-running setup should also relocate a managed block previously written into a nested block.

## Root cause

`crates/jcode-setup-hints/src/linux_niri.rs::binds_block_insert_point` currently returns the first line that textually starts with `binds` and ends with `{`, without tracking KDL brace depth.

tingfeng347:
A fix with regression coverage is available at https://github.com/tingfeng347/jcode/compare/master...fix/niri-top-level-binds-injection (commit 6a512ed8b). GitHub currently rejects PR creation for this account with a createPullRequest permission error.

1jehuang:
Thanks for the clear repro and the reference branch, @tingfeng347. Reproduced and fixed.

Root cause matches yours exactly: `binds_block_insert_point` returned the first line that textually started with `binds` and ended with an opening brace, with no KDL brace-depth tracking, so a nested `recent-windows` binds node won.

The fix tracks brace depth (ignoring braces inside quoted strings and line comments) and only inserts into a top-level `binds` node. If the main file has no top-level `binds`, it appends one, as you suggested. It also handles your re-run case: a managed region that an older version wrote into a nested node is now stripped and relocated to a top-level block instead of being left broken.

Regression coverage added for four cases: nested-only config, nested-before-top-level (must prefer the top-level one), relocation of a previously-misplaced managed block without duplicating it, and brace counting with braces inside strings and comments.

[...]

PR #727, opened 2026-08-02
fix: triage batch (9 issues: #729, #728, #726, #724, #722, #719, #715, #712, #706)

1jehuang:
Automated triage pass over the open issue backlog. Nine fixes, one commit each, every one with a verified root cause and regression coverage.

| Issue | Fix | Commit |
|---|---|---|
| #726 | `dev_cargo.sh` forced `clang` as the linker driver even in `system` mode, so builds failed with exit 101 without clang installed | `9ecca3540` |
| #724 | `claude-opus-5` missing from the subscription catalog while the live router advertised it for every paid tier | `e222069de` |
| #706 | Curated OAuth `ScheduleWakeup` schema had drifted from the real `schedule` tool, so every call failed with `task is required for action=create` | `3e8cf17c7` |
| #722 | Curated OAuth `Bash` schema omitted `justification`, the destructive gate's own escape hatch | `3e8cf17c7` |
| #719 | `setup-hotkey` injected global launch binds into a nested `recent-windows` binds node, breaking the whole niri config | `bd0de74ff` |
| #715 | `spawn_in_new_terminal` was a non-Unix no-op, so `/judge`, `/fork`, `/review`, `/transfer` and crash-restore never opened a window on Windows | `9bcc8fe65` |
| #712 | `auth-test --provider-profile <name>` cleared the profile before probing, false-negativing every custom named profile | `40ec8fe70` |
| #728 | Word-delete with an active selection left a stale anchor; the next Backspace panicked across winit's ObjC boundary and **aborted the process** | `621a264b0` |
| #729 | Swarm-spawned workers were forced into isolated test-mode memory, so they could never see real project/global memory | `3e8550913` |

Plus `1e4c50211`, which fixes a CI blocker inherited from master (see below).

Follow-up hardening after a second verification pass, which found and closed three gaps:

- `f1eabfb91` (#712): the probe targeted the right profile but the *report* still said `openai-compatible`, so the verdict was correct yet unattributable. Header and credential line now name the profile.
- `dd1ceab90` (#715): the first guard only proved the function is compiled. A behavioral test now proves the resume invocation reaches a launcher and is reported as successful, using the spawn hook as the injectable seam.
- `d3303bad6` (#729): the isolation flag was a bare `bool`, one typo from reintroducing the bug. Now `HeadlessMemoryScope::{RealProject, IsolatedTest}`, so each call site states its intent.

## Two recurring patterns worth a look

**Hand-maintained copies drifting from their source (#706, #722).** The Anthropic OAuth path hand-maintains curated tool schemas so the subscription endpoint sees Claude-Code builtin *names*. Those copies drift, and the drift is invisible until a model calls the tool and the handler rejects the arguments. `ScheduleWakeup` was 100% broken on subscription routes. Rather than correct the copy, this deletes it and forwards the real schema under the OAuth name, so the two cannot diverge again. Worth checking whether `Agent`, `Read`, `Write`, `Edit`, `Glob`, `Grep` and `Skill` have drifted the same way.

**Comments describing a contract the code stopped honouring (#729).** `enable_memory_test_mode` is documented "Called when session is marked as debug", but `create_headless_session` called it unconditionally, and that function is also the production path for real swarm workers.

## Verification

- `jcode-base`, `jcode-provider-anthropic`, `jcode-setup-hints`: 1359 passed
- `jcode` lib: 207 passed. `jcode-desktop2` editor: 62 passed
- **All 10 CI checks pass**, including Build & Test on ubuntu, macOS and Windows, plus the Windows Cross-Target Check

Beyond unit tests, the three fixes with external dependencies were verified against the real thing:

- **#726**: built a clang-free PATH and confirmed pre-fix exports `linker=clang` (the exit-101 cause) while fixed leaves it unset, then ran a real build on that PATH
- **#724**: ran the reporter's exact commands on a built binary. `claude-opus-5` now appears, and `--model claude-opus-5` selects it instead of silently substituting `claude-opus-4-8`
- **#719**: ran the spliced config through the real `niri validate`. Added that as a permanent test and confirmed it fails with the brace-depth check removed and passes with it

Same fix-removal check for #728: both new tests fail on pre-fix code and pass after.

## Inherited CI blocker, fixed here

Commit `05014482b` on master added a `jcode-provider-core` dev-dependency without committing the regenerated lock file, so every job running cargo with `--locked` failed before compiling anything. Both Windows jobs were red on master for this. Refreshed in `1e4c50211` (one generated line, no version changes).

## Known-failing gate

`Quality Guardrails` fails on the oversized-file, oversized-test and swallowed-error ratchets. Measured entry counts on a clean `origin/master` worktree versus this branch:

| Ratchet | clean master | this PR |
|---|---|---|
| oversized-file | 16 | 16 |
| oversized-test | 5 | 5 |
| swallowed-error | 7 | 7 |

Identical, so this PR contributes **zero** additional drift despite adding nine fixes and eight test modules. Tracked in #692, where I posted these figures.

Also filed #730: `tool::todo::tests::schema_advertises_intent_and_todos` fails on clean `master`, and CI's targeted cohorts do not run workspace unit tests, so it is currently invisible.

One baseline line moved (`src/cli/provider_init.rs`, 1831 -> 1836) for the #712 fix. No other baselines touched.

## Scope notes

- **#724**: opus-5 is added as a **non-default** entry. Changing the default model for paid users is a product decision. The silent-fallback-on-unknown-model behaviour is left open deliberately, and also appears in #705.
- **#715**: verified via the Windows CI build and cross-target check rather than local hardware.

Credit to the reporters, who did most of the root-cause work: @MatrixMagician (#726), @caipira113 (#724), @rameshbaskaran (#706, #722), @tingfeng347 (#719), @rensburger (#715), @e-kotov (#712), @wtchronos (#728), @jonschull (#729).


1jehuang:
## CI status

**Fixed a blocker that predates this PR.** Every job running cargo with `--locked` was failing before compiling anything:

```
error: cannot update the lock file ... because --locked was passed to prevent this
```

Commit `05014482b` on master added `tests/context_window_matrix.rs` with a `jcode-provider-core` dev-dependency but did not commit the regenerated lock file. `Windows Cross-Target Check` and `Build & Test (windows-latest)` were red on master for this reason, not because of anything here. Refreshed in `1e4c50211` (one generated line, no version changes). Both Windows jobs now pass, which also gives real cross-compile validation of the #715 change.

**Remaining failure is pre-existing.** `Quality Guardrails` fails on the oversized-file, oversized-test and swallowed-error ratchets. Verified these are identical to clean `master` (checked against a fresh worktree), and confirmed in the CI log that **no ratchet entry names any file touched by this PR**. The named files are `desktop2/transcript.rs`, `tui/ui.rs`, `app/input.rs`, `tool/todo.rs`, `provider/catalog_routes.rs` and similar, all from other in-flight work. This is #692.

Passing: Format, Build & Test (ubuntu), Windows Cross-Target Check, PowerShell Syntax, Release Automation, Setup Friction Eval, Require Linked Issue, Greptile Review.

1jehuang:
## Final CI state

All 10 platform and quality checks pass: Build & Test on **ubuntu, macOS and Windows**, plus Format, Windows Cross-Target Check, PowerShell Syntax, Release Automation, Setup Friction Eval, Require Linked Issue and Greptile Review.

Two things found and handled during verification, both worth recording:

**1. One of my own new tests was flaky, and CI caught it.** The first `#715` behavioral test drove the configured spawn hook, which resolves through the process-wide *cached* config. It passed standalone and failed in CI, where an earlier test in the same binary had already loaded config. Rewritten in `804d93e63` to drive `spawn_command_in_new_terminal_with` directly, the injectable seam the real path bottoms out in: no config, no env vars, no process spawning, so ordering cannot affect it. Verified passing standalone, in-module, and under `--test-threads=1`, and still failing when the forwarded args are emptied.

**2. A pre-existing macOS flake, not from this PR.** `reload_notifies_clients_across_independent_sessions` failed once, having passed twice earlier on this same branch. Diagnosed to root cause and filed as #731: `get_history_event` skips only `Ack`, so a racing `SwarmStatus` broadcast gets returned as if it were the history reply. Confirmed by re-running the job with no code change: it passes.

**Remaining red gate is inherited.** Measured entry counts, clean `origin/master` worktree versus this branch: oversized-file 16/16, oversized-test 5/5, swallowed-error 7/7. Identical, so this PR adds **zero** drift despite nine fixes and eight new test modules. Tracked in #692.

Also filed #730 (a `jcode-app-core` unit test failing on clean `master`, invisible because CI runs targeted cohorts rather than workspace unit tests).

PR #820, opened 2026-08-06
fix: prevent session loss and repair launch lifecycle races

### LINKED ISSUE #792: Feature request: native tmux pane spawning for swarm teammates (same-window multi-agent view, like Claude Code)
### What I want

Native, first-class support for spawning swarm teammates as **interactive tmux panes in the same window** as the coordinator — the way Claude Code does it. Today `swarm spawn` works, but the teammate sessions do not appear alongside the coordinator on one screen, so there is no way to watch several agents work at once without manual shell glue.

### Why this matters

The value of a swarm is *observability while it runs*. With headless/inline spawns I get a report at the end; with panes I can see an agent go down a wrong path and correct it immediately. Claude Code's multi-pane view is the single biggest ergonomic difference today.

### Current state (v0.67.1) — this is mostly already built

Digging through the binary and testing on tmux 3.7b, almost all the machinery exists:

- The binary already says: *"Without a hook, clients inside tmux automatically use a right-side pane."*
- `[terminal] spawn_hook` / `focus_hook` exist and receive `<hook> <jcode-binary> <args...>` plus `JCODE_SPAWN_SESSION_ID`, `JCODE_SPAWN_TITLE`, `JCODE_SPAWN_CWD`, `JCODE_SPAWN_PROGRAM`, `JCODE_SPAWN_COMMAND`.
- `TMUX` / `TMUX_PANE` are already probed alongside Zellij, kitty, WezTerm, Ghostty, herdr.
- The documented example is `spawn_hook = "tmux new-window"` — a **window** per agent, not a pane.

Two things block the out-of-the-box experience:

1. **`[agents] swarm_spawn_mode` defaults to `"inline"`** (there is even a one-time migration string `swarm_spawn_mode default migration: visible -> inline`). Inline is in-process with no window, so **no spawn hook ever fires** and the tmux auto-pane path is unreachable. This is the actual root cause, and it is not discoverable — you have to know to flip it to `"visible"` first.
2. **Naive `tmux split-window -h` collapses geometry.** Measured in a 200x50 window: pane widths went 100 → 49 → 24 → 12 → 11 columns after 4 agents. Usable tiling needs `tmux select-layout tiled` after each split.

### Working config today (for anyone hitting this)

```toml
[agents]
swarm_spawn_mode = "visible"

[terminal]
spawn_hook = "~/bin/jcode-tmux-pane"
```

```sh
#!/bin/sh
# jcode spawn_hook: each spawned session becomes a tiled tmux pane in the CURRENT window.
set -eu
if [ -z "${TMUX:-}" ] && [ -z "${TMUX_PANE:-}" ]; then
  exit 1   # not in tmux -> non-zero lets jcode fall back to its own terminal detection
fi
TARGET="${TMUX_PANE:-}"
CWD="${JCODE_SPAWN_CWD:-$PWD}"
if [ -n "$TARGET" ]; then
  tmux split-window -h -t "$TARGET" -c "$CWD" "$@"
else
  tmux split-window -h -c "$CWD" "$@"
fi
tmux select-layout tiled
[ -n "${JCODE_SPAWN_TITLE:-}" ] && tmux select-pane -T "$JCODE_SPAWN_TITLE" 2>/dev/null || true
exit 0
```

Verified: 3 simulated spawns produced 4 panes at ~100x24 with pane titles set from `JCODE_SPAWN_TITLE`; outside tmux the hook exits 1 and jcode correctly falls back to built-in terminal detection.

### Proposed

1. **`swarm_spawn_mode = "tmux"`** (or make `"visible"` detect tmux natively) that does split + `select-layout tiled` + `select-pane -T "$JCODE_SPAWN_TITLE"` internally, with no user shell script.
2. **`[agents] tmux_layout`** = `tiled` | `even-horizontal` | `even-vertical` | `main-vertical`, so the coordinator can keep a large main pane with teammates tiled beside it.
3. **Pane lifecycle tied to session lifecycle** — pane closes (or is marked dead, per a `tmux_close_on_exit` setting) when the teammate reaches `ready`/`stopped`/`failed`, so `swarm cleanup` doesn't leave orphan panes.
4. **Cap and overflow behaviour** — respect `swarm_max_concurrent_agents` (mine is 32, which would be unreadable as panes); beyond `tmux_max_panes`, spill to new tmux windows rather than shrinking panes to 11 columns.
5. **Discoverability** — when `swarm spawn` runs inside tmux while `swarm_spawn_mode = "inline"`, emit a one-time hint that tmux panes are available.

### Difficulty

Items 1 and 2 look small: the terminal-detection, hook-invocation, and env-passing paths already exist, so this is mostly a built-in hook implementation plus a config enum. Item 3 (lifecycle) is the one that needs real design, since it touches session teardown and the same cleanup concerns raised in #737 for `bg` watch sources.

### Environment

jcode v0.67.1 (88a19f38e), macOS aarch64, tmux 3.7b.

### LINKED ISSUE #798: Memory sidecar is dead on Claude credentials: SIDECAR_CLAUDE_MODEL uses the Haiku 3.5 date, 404s on every call
## Problem

The memory sidecar's Claude model is pinned to `claude-haiku-4-5-20241022`:

https://github.com/1jehuang/jcode/blob/master/crates/jcode-base/src/sidecar.rs#L22

`20241022` is the **Haiku 3.5** snapshot date. Paired with a `claude-haiku-4-5` base it forms an id Anthropic does not publish, so every sidecar call on the Claude path returns `404 not_found_error`.

Verified against the live API with jcode's own OAuth credentials and headers:

```
claude-haiku-4-5-20241022      HTTP 404  {"type":"error","error":{"type":"not_found_error","message":"model: claude-haiku-4-5-20241022"},...}
claude-haiku-4-5-20251001      HTTP 200  {"content":[{"type":"text","text":"OK"}],...}
```

`GET /v1/models` confirms `claude-haiku-4-5-20251001` is the only Haiku in the catalogue, so the correct id is unambiguous.

The failure is near-silent, which is why it survived so long. `rerank_candidates_consensus_attributed` catches the transport error, logs it rate-limited, and arms a circuit breaker:

```
[WARN] EVENT event="Memory consensus judge failed; circuit breaker armed" error="Claude API error (404 Not Found): ... model: claude-haiku-4-5-20241022 ..."
```

After that the breaker suppresses the judge for the rest of the session. Nothing surfaces to the user, and since the judge is the only thing allowed to surface memory, memory recall silently degrades rather than erroring.

`git log -S` dates the bad id to the original `feat: add Haiku 4.5 sidecar for memory system` commit, so **the Claude sidecar path has never worked**. Users on Codex credentials are unaffected (they select `SIDECAR_OPENAI_MODEL`); only Claude-credentialed users without Codex creds hit it.

Every other reference in the tree already uses the correct id (`auth/lifecycle.rs`, `provider-core/anthropic.rs`, `provider-core/model_id.rs`, `tui/model_names.rs`). `sidecar.rs` is the sole outlier.

## Expected behavior

- The sidecar's Claude model is an id the Anthropic catalogue actually publishes.
- A wrong model id does not degrade into a silently-disabled feature.

## Fix

1. Pin `SIDECAR_CLAUDE_MODEL` to `claude-haiku-4-5-20251001`, matching the rest of the codebase.
2. Add a regression test asserting the id is a dated Haiku snapshot and specifically rejecting the 3.5 date, so a family/date mismatch cannot recur.

Affected file:

- `crates/jcode-base/src/sidecar.rs`

I have this fixed and tested locally (1252/1252 `jcode-base` tests pass) and am happy to open a PR.

### Possible follow-up

The near-silence is arguably the more valuable fix. A 404 `not_found_error` is a permanent configuration fault, not a transient transport failure, yet `record_failure_backoff` treats them identically. Distinguishing "this model does not exist" from "the network blipped" and surfacing the former once at startup would have made this immediately visible instead of invisible for the feature's entire lifetime. Happy to file that separately if useful.

### LINKED ISSUE #814: Data-loss: session close path can wipe a large transcript (fix + tests available on fork)
## Summary

A close/clear path that swaps in a fresh 0-message session and saves with `status=Closed` triggers `save()`'s shrink detection (`messages.len() < persist_state.messages_len`) and checkpoints a full snapshot — **overwriting the real transcript snapshot with an empty stub and deleting the journal**. The later `.bak` rotation then captures the already-wiped state, so both copies end up empty and `--resume` shows an empty session.

Hit this in production 2026-08-05 on a 12-hour session (686 messages, 1.5MB snapshot). Daemon log forensics:

```
10:55:06  SESSION_PERSISTENCE save_mode=append    messages=686  snapshot_bytes=1,539,899
10:55:38  SESSION_PERSISTENCE save_mode=snapshot  messages=0    snapshot_bytes=600  status=Closed   <- wipe
10:55:41  SESSION_PERSISTENCE load_done           messages=0                                        <- resume sees nothing
```

## Fix (ready on my fork — PR creation appears restricted on this repo)

Branch: https://github.com/dwillitzer/jcode/tree/fix/session-pre-wipe-guard (single commit, cherry-picks cleanly onto `origin/master`)

`checkpoint_snapshot()` gains a `guard_snapshot_shrink()` pre-step: when the in-memory transcript is **empty** but the on-disk snapshot exceeds 4KB, snapshot + journal are first copied to timestamped `.pre-wipe-<ts>.json.bak` / `.jsonl.bak` siblings.

- Never blocks the save — legitimate compaction shrinks are unaffected (guard requires empty; metadata stubs are under threshold)
- Backup names avoid `*.json` so session scanners don't surface phantoms
- Emits `SESSION_PERSISTENCE phase=pre_wipe_backup` for observability; backup failure logs and proceeds (fail-open)
- Recovery: `cp session_X.pre-wipe-<ts>.json.bak session_X.json`

Includes two regression tests (wipe path produces byte-identical restorable backup with full recovery round-trip; small/compaction checkpoints produce no backups).

## Related fixes on the same fork, happy to PR any/all if you enable PRs or prefer patches:

| Branch | What |
|---|---|
| [`fix/swarm-schema-anyof`](https://github.com/dwillitzer/jcode/tree/fix/swarm-schema-anyof) | Top-level `anyOf` in the swarm tool schema breaks strict providers (Anthropic API rejects it); flattened to a plain object schema |
| [`fix/swarm-worker-provider-pin`](https://github.com/dwillitzer/jcode/tree/fix/swarm-worker-provider-pin) | Swarm worker spawns drop the `openai-compatible:<profile>` prefix, so workers lose the auth profile and fail; pins the full prefix through spawn (with tests) |
| [`feat/prompt-cache-override`](https://github.com/dwillitzer/jcode/tree/feat/prompt-cache-override) | `supports_prompt_cache` config override so openai-compatible profiles fronting Anthropic models emit `cache_control` breakpoints (verified ~92% cached-token hit rate through a claude-family proxy) |
| [`fix/mcp-tool-name-sanitize`](https://github.com/dwillitzer/jcode/tree/fix/mcp-tool-name-sanitize) | MCP tool names with characters outside `[a-zA-Z0-9_-]` are rejected by provider APIs; sanitizes at registration. Also prefers configured judge models over hardcoded defaults |

All branches are single-purpose, based on current `master` (02439b49), and cherry-pick cleanly. If you'd rather receive these differently (git bundles, mailed patches, enabling PRs), say the word.

### LINKED ISSUE #818: JcodeClient.close() can leave the daemon process running (servers.json registration race)
# `JcodeClient.close()` can leave the daemon process running (servers.json registration race)

## Summary

`close()` on an instance from `JcodeClient.launch()` (TypeScript SDK,
`@1jehuang/jcode-sdk`) resolves successfully, but the underlying
`jcode ... serve` daemon process can be left running indefinitely,
reparented to init. It does not appear as a zombie and is not hung — a
plain `kill -TERM <pid>` sent directly to it terminates it immediately.
The problem is that `close()` never sends it a signal in the first place.

Under concurrent `launch()`+`close()` calls (the pattern you'd use to run
several agent instances in parallel), I measured this happening in
roughly 1 out of every 8 instances on the machine I tested on. It also
reproduces reliably as a single, isolated race when instrumented directly
(see "Root cause" below) — the frequency above is this specific
hardware/load, not a claim about the general rate.

## Environment

- jcode: `v0.67.1 (88a19f38e)`
- `@1jehuang/jcode-sdk`: `1.1.0` (installed via `npm install @1jehuang/jcode-sdk`, which pulled `@1jehuang/jcode-linux-x64@1.1.0` as the platform binary)
- Node: `v22.22.2`
- OS/kernel: `Linux 6.18.5-fc-v18 x86_64` (a constrained single-vCPU sandboxed VM — see caveat below)
- No real provider credentials involved; `ANTHROPIC_API_KEY` was set to a dummy string purely so the daemon's provider-init check at startup didn't refuse to boot. No network calls to a model provider occur before `close()` in the repro.

## Root cause (found by reading `sdk/typescript/src/launch.ts`)

`close()` calls `stopInstanceDaemon()`, which calls `readDaemonPidSync()`
to find the daemon's PID:

```ts
function readDaemonPidSync(jcodeHome: string, runtimeDir: string): number | undefined {
  let raw: string;
  try {
    raw = fs.readFileSync(path.join(jcodeHome, "servers.json"), "utf8");
  } catch {
    return undefined;
  }
  ...
}
```

and if it can't find an entry, `stopInstanceDaemon()` returns immediately
without sending any signal:

```ts
async function stopInstanceDaemon(...): Promise<void> {
  const pid = readDaemonPidSync(jcodeHome, runtimeDir);
  if (pid === undefined) return;   // <-- silently gives up here
  ...
}
```

`servers.json` is written by the daemon itself, as part of its own
startup sequence — apparently on a separate timeline from when the API
socket becomes connectable, which is what `launch()` actually waits on
before resolving. I confirmed this directly:

```
jcodeHome: /tmp/jcode-sdk-instance-bmnmWX
servers.json exists immediately after launch()? false      <-- launch() has resolved, file isn't there yet
```

Waiting 3 seconds before calling `close()` let the same daemon shut down
cleanly (`servers.json` existed and had the correct `pid`/`socket`
entry by then). So this is a startup-ordering race: **`launch()`
resolving does not imply `servers.json` has been written**, but
`close()`'s cleanup depends entirely on that file.

I also confirmed the daemon isn't the thing misbehaving — sending it a
direct `SIGTERM` after the leak occurred killed it instantly. `close()`
simply never signals it.

## Reproduction

Minimal script (attached, `repro.mjs`):

```js
import { JcodeClient } from "@1jehuang/jcode-sdk";

const client = await JcodeClient.launch({
  workingDir: process.cwd(),
  inheritLogins: false,
  env: { ANTHROPIC_API_KEY: "sk-ant-repro-000000000000000000000000000" },
});

await client.close();
// check `ps` afterward for a lingering `jcode ... serve` process
```

This single-shot repro didn't reproduce reliably in my environment
(0/8 in a clean sequential foreground loop) — the race window is narrow
under low load. It reproduces far more reliably:

1. **Directly, by instrumentation** — reading `servers.json` state right
   after `launch()` resolves shows it missing, every time I checked, in this
   environment. This is the actual bug; whether it manifests as a full
   leak depends on additional scheduling luck.
2. **Under concurrency** — firing several `launch()` + `close()` pairs at
   once (`Promise.all`, attached as `concurrent_repro.mjs`) leaked 1 of 8
   daemons in one run. This is the pattern most likely to matter in
   practice: an application launching a batch of parallel agent instances.

```
daemons before: 0
daemons after 8 concurrent launch+close pairs: 1 [ '<pid>' ]
leaked: 1 / 8
```

## Expected behavior

After `close()` resolves, no daemon process associated with that
instance should remain running, regardless of how soon after `launch()`
`close()` is called.

## Suggested directions for a fix

I'm not familiar enough with the daemon's Rust-side bootstrap sequence to
know which of these fits best, but a few options:

1. Have the daemon write its `servers.json` entry *before* it opens the
   API socket, so "socket is connectable" implies "registry entry
   exists." This seems like the cleanest fix if there's no ordering
   dependency the other way around.
2. Have `stopInstanceDaemon()` retry/poll for the `servers.json` entry
   for a short grace window (e.g. up to a second or two) instead of a
   single synchronous read that gives up immediately.
3. Have the bridge process (which is the one `launch()` actually spawns
   and holds a handle to) track the daemon's PID itself as a fallback,
   rather than relying solely on a file the daemon writes independently.

## Secondary observation (possibly related, less rigorously isolated)

In one run, a `sequential_repro.mjs` variant (8x `launch()`+`close()` in
a loop, awaited one at a time in the same process) took long enough that
an outer `timeout 60` killed the Node process mid-run. After that forced
kill, both a leaked daemon *and* several un-removed
`jcode-sdk-instance-*` temp directories were left behind. This may just
be an artifact of killing Node before its `exit` handler could run (the
README says `close()` on ephemeral homes can legitimately take up to
`cleanupTimeoutMs`, default 30s), rather than a distinct bug — flagging
it in case it's useful, but I haven't isolated it the way I did the main
issue above.

## Caveat about my environment

This sandbox's PID 1 is a `firecracker-init` process rather than a
standard init/systemd, and it's a single-vCPU box, both of which are
somewhat unusual and could plausibly affect scheduling/timing. I
couldn't test on a standard host/VM/Docker container to confirm the race
reproduces there too. That said, the root cause I found by reading the
source (`launch()` resolving before `servers.json` is guaranteed to
exist) doesn't look environment-specific — it's a genuine ordering gap
between two independent operations, so I'd expect it to reproduce
elsewhere too, just perhaps less often on faster/less loaded hardware,
and potentially more often on a busy CI runner or a host launching many
instances concurrently, which is exactly the "launch many agents at
once" use case the SDK seems designed for.

Happy to run more diagnostics if it'd help track this down further.

---

## Repro scripts

<details>
<summary><code>repro.mjs</code></summary>

```js
// Minimal repro: JcodeClient.close() can leak the underlying daemon process.
//
// Root cause (see bug report): close() -> stopInstanceDaemon() reads the
// daemon's pid from `${jcodeHome}/servers.json`. That file is written by the
// daemon itself as part of its own startup, on a separate timeline from the
// API socket becoming connectable (which is what launch() actually waits
// on). If close() runs before the daemon has written its servers.json entry,
// readDaemonPidSync() returns undefined, stopInstanceDaemon() returns
// immediately without sending any signal, and the daemon is orphaned
// (reparented to init) with nothing left tracking it.
//
// Usage:
//   npm install @1jehuang/jcode-sdk
//   node repro.mjs
//
// Expected: no `jcode ... serve` process remains after close() resolves.
// Actual: one is left running indefinitely.

import { JcodeClient } from "@1jehuang/jcode-sdk";
import { execSync } from "node:child_process";

function jcodeServeProcesses() {
  try {
    return execSync("ps -eo pid,ppid,stat,cmd | grep '[j]code.*serve'").toString().trim();
  } catch {
    return "(none)";
  }
}

console.log("jcode serve processes before:\n" + jcodeServeProcesses());

const client = await JcodeClient.launch({
  workingDir: process.cwd(),
  inheritLogins: false,
  // Any non-empty value; no real network call happens before close().
  env: { ANTHROPIC_API_KEY: "sk-ant-repro-000000000000000000000000000" },
});

console.log("\nlaunch() resolved. jcode serve processes:\n" + jcodeServeProcesses());

await client.close();
console.log("\nclose() resolved.");

await new Promise((r) => setTimeout(r, 2000));
console.log("\njcode serve processes 2s after close() resolved:\n" + jcodeServeProcesses());
console.log(
  "\nIf a process is listed above, close() failed to reap it -- this is the bug.",
);

```

</details>

<details>
<summary><code>concurrent_repro.mjs</code></summary>

```js
import { JcodeClient } from "@1jehuang/jcode-sdk";
import { countJcodeDaemons } from "./count_daemons.mjs";

const N = parseInt(process.argv[2] || "8", 10);

async function trial(i) {
  const client = await JcodeClient.launch({
    workingDir: process.cwd(),
    inheritLogins: false,
    env: { ANTHROPIC_API_KEY: `sk-ant-repro-${i}-000000000000000000000` },
  });
  await client.close();
  return i;
}

const before = countJcodeDaemons().count;
console.log("daemons before:", before);

await Promise.all(Array.from({ length: N }, (_, i) => trial(i)));

await new Promise((r) => setTimeout(r, 2000));
const after = countJcodeDaemons();
console.log(`daemons after ${N} concurrent launch+close pairs:`, after.count, after.pids);
console.log(`leaked: ${after.count - before} / ${N}`);

```

</details>

<details>
<summary><code>count_daemons.mjs</code></summary>

```js
import fs from "node:fs";

export function countJcodeDaemons() {
  let count = 0;
  const pids = [];
  for (const entry of fs.readdirSync("/proc")) {
    if (!/^\d+$/.test(entry)) continue;
    try {
      const exe = fs.readlinkSync(`/proc/${entry}/exe`);
      if (exe.endsWith("jcode-linux-x86_64.bin")) {
        // Confirm it's actually running the "serve" subcommand, via cmdline
        // (NUL-separated, so no risk of matching shell wrapper text).
        const cmdline = fs.readFileSync(`/proc/${entry}/cmdline`, "utf8").split("\0");
        if (cmdline.includes("serve")) {
          count++;
          pids.push(entry);
        }
      }
    } catch {
      // Process exited between readdir and read, or no permission.
    }
  }
  return { count, pids };
}

if (import.meta.url === `file://${process.argv[1]}`) {
  console.log(countJcodeDaemons());
}

```

</details>

1jehuang:
## Summary
- wait briefly for a newly launched daemon to register before SDK `close()` gives up
- cover delayed `servers.json` publication with a regression test
- update the Claude memory sidecar to the valid Haiku 4.5 model ID
- refuse destructive empty transcript checkpoints and preserve byte-identical pre-wipe recovery copies
- detect immediate non-zero terminal spawn-hook exits so visible/auto swarm spawns fall back instead of remaining `startup queued`

Fixes #818.
Fixes #798.
Fixes #814.
Fixes #792.

## Verification
- `npm test -- --test-name-pattern="daemon shutdown lookup|cleanup refuses|removing an instance"`
- `cargo test -p jcode-base --lib sidecar::`
- `cargo test -p jcode-base session::persistence::tests --lib` (3 passed)
- `cargo test -p jcode-base terminal_launch::tests --lib`
- `cargo test -p jcode-app-core server::comm_session_tests::prepare_visible_spawn_session --lib`
- 8 concurrent SDK launch/close lifecycle trials, 0 leaked daemon registrations
- `git diff --check`
- `cargo fmt --all -- --check`


1jehuang:
CI follow-through found and repaired three unrelated guardrail regressions already present on the base branch: a collapsible conditional, a dead auth helper, and a contradictory strict-schema fixture. The focused suites now pass locally. The repository-wide guardrail is still blocked by two additional pre-existing Rust 1.97 clippy findings in `jcode-app-core` (`needless_return` and `too_many_arguments`), so I am leaving this PR unmerged rather than expanding an issue-specific fix into a broad cleanup.

1jehuang:
Final CI follow-through: the issue-specific SDK and sidecar checks continue to pass, and the bounded Rust 1.97 findings fixed on this branch are resolved. The repository-wide guardrail now advances into `jcode-tui` and reports 14 additional base-branch dead-code/cfg lint failures across onboarding and terminal notifications. That is a broad, unrelated cleanup boundary, so this PR remains open and unmerged rather than masking or absorbing those changes.

1jehuang:
Runtime acceptance check for #818 passed: a selfdev build of this branch completed, then 8 concurrent public SDK `launch()` → immediate `close()` trials completed with 0 leaked daemon processes. This exercises the SDK, bridge, daemon registration, shutdown, process-group signaling, and ephemeral cleanup boundary end to end.

1jehuang:
Acceptance follow-up:

- #814: packaged-binary, isolated-daemon close/resume passed. A marker written before close was returned exactly after resume.
- #792: focused hook and visible-session lifecycle tests pass. A packaged-binary acceptance attempt reached the swarm public interface, but standalone `jcode run` sessions cannot initialize swarm membership (`Not in a swarm`), so a live tmux swarm result remains acceptance-blocked rather than claimed as passed.
- The full TUI package compiled and its public `version` and `--help` entry points passed.

PR #841, opened 2026-08-08
fix: resolve verified issue triage regressions

### LINKED ISSUE #824: Multi-line tool-error payloads leak past the focused-query filter and bias the memory judge toward safety facts
## Summary

`focus_query_text` (`crates/jcode-base/src/memory_prompt.rs`) strips tool noise by
dropping lines that **start with** `[Tool error:` / `[Tool:`, but tool-error
payloads are multi-line and only the first line carries that prefix. Subsequent
lines of the error message pass through into the focused query, the retrieval
context, and the LLM judge. Sandbox command-safety rejections in particular —
"This command was not run...", "cannot be confirmed", "is irreversible", "could
not be identified statically" — are lexically close to safety/governance memory
facts, so they shift the judge's votes away from the actual topic.

## Impact

Verified in an ACP task (NTL-486): the agent's tool calls tripped the sandbox on
protected commands; the multi-line `[Tool error: ...]` blocks reached the judge,
which then injected 5 safety-related facts instead of the token-related fact the
task actually asked about. The throttle had fired and the pool was large (30/22
candidates) — the loss was caused by the noisy judge input, not by retrieval
coverage. Any host whose tasks run under restrictive tool policies (read-only
review, sandboxed containers, blocked globs/recursive deletes) hits this
constantly.

## Root cause

`format_content_block_for_relevance` wraps an error payload in a single block:

```rust
ContentBlock::ToolResult { content, is_error, .. } => {
    if is_error.unwrap_or(false) {
        Some(format!(
            "[Tool error: {}]",
            truncate_chars(content.trim(), MEMORY_CONTEXT_MAX_BLOCK_CHARS / 4)
        ))
    } else {
        None
    }
}
```

The payload keeps its internal newlines, so `focus_query_text`, which iterates
line-by-line, drops only the first line:

```rust
if trimmed.is_empty()
    || trimmed.starts_with("[Tool:")
    || trimmed.starts_with("[Tool error:")
    || ...
{
    continue;
}
```

The remainder — e.g. `- target contains a glob, so the exact set of affected files
is not known before execution (target: /sys/...)` or `- targets a protected system
or home path that must never be destroyed (target: a user home dir)` — is treated as
session prose and kept in the focused query.

## Proposed change (implemented and verified locally)

Drop command-safety block errors by **content**, not only by the `[Tool error:`
prefix:

```rust
if trimmed.is_empty()
    || trimmed.starts_with("[Tool:")
    || trimmed.starts_with("[Tool error:")
    || trimmed.starts_with("[Result:")
    || trimmed.starts_with("[Image]")
    || trimmed.starts_with("[OpenAI native compaction]")
    // Multi-line error payloads only carry the prefix on the first line; drop the
    // sandbox-block markers wherever they appear so safety wording cannot steer
    // retrieval or the judge toward safety facts.
    || trimmed.contains("This command was not run")
    || trimmed.contains("cannot be confirmed")
    || trimmed.contains("is irreversible")
    || trimmed.contains("could not be identified statically")
    || trimmed.contains("could not be identified")
{
    continue;
}
```

A more general fix would track "inside a `[Tool error:` block" state across lines
(like the existing `<system-reminder>` handling) instead of a marker list; the
marker list above is the minimal, backward-compatible version.

## Environment

- jcode v0.68.0, Linux x86_64, DeepSeek provider, local all-MiniLM-L6-v2 embeddings.
- ACP protocolVersion 1, host: Multica daemon; 2-vote consensus judge
  (`memory_rerank_votes = 2`).
- Verified locally: with the content filter, the judge no longer sees sandbox
  wording and votes on the actual topic.

### LINKED ISSUE #826: Jcode cannot be stopped while retrying after rate limit is hit
### Description

When using a model that has reached its rate limit, JCode continuously retries the request indefinitely. During this state, the application becomes effectively impossible to stop, interruption is ignored, and the retry loop continues.

### Steps to Reproduce

1. Configure a model/API with a low rate limit.
2. Send requests until the rate limit is reached.
3. Observe the retry behavior.
4. Attempt to stop the current operation.

### Expected Behavior

- Retry logic should either:
  - stop after a configurable number of attempts, or
  - wait according to the provider's `Retry-After` header.
- Instructing **Stop** should immediately cancel all pending retries and terminate the request.

### Actual Behavior

- The application keeps retrying indefinitely after receiving the rate limit error.
- The Stop instruction has no effect while retries are occurring.
- The only way to recover is to forcibly terminate the application/process.

### Additional Information

I've attached a screenshot showing the retry loop and the inability to stop the process.

Image

### LINKED ISSUE #831: TUI cost widget shows zero dollar cost for OpenAI-compatible API-key providers (DeepSeek): remote billing gate uses a hard-coded provider allow-list
## Summary

The TUI cost widget shows **no dollar cost** (stays at `$0`) for OpenAI-compatible API-key providers such as **DeepSeek**, even though the models.dev pricing catalog has correct prices and the local cost path resolves them.

### Version
- jcode v0.68.0 (fcf53909f), Linux x86_64, DeepSeek provider (OpenAI-compatible)
- Remote/TUI mode (the default interactive client talks to the daemon)

### Root cause

`resolve_remote_cost_pricing()` in `crates/jcode-tui/src/tui/app/misc_ui.rs` decides whether a remote session is billed per token using a **hard-coded provider-name allow-list**:

```rust
let billed = if is_anthropic || is_openai {
    api_key_billed
} else {
    // Providers that are inherently cost-based when proxied remotely.
    provider_name.contains("opencode")
        || provider_name.contains("openrouter")
        || provider_name.contains("bedrock")
        || provider_name.contains("cerebras")
        || provider_name.contains("compatible")
};
```

The daemon reports the active provider's **display name** via `provider_name()` -> `runtime_display_name()`. For DeepSeek (an OpenAI-compatible profile) that is `"DeepSeek"`, which matches **none** of the allow-list strings. So `billed == false`, the function returns `None`, and `accrue_remote_call_cost` never adds anything — the cost stays `$0`.

The **local** cost path (`update_cost_impl`, same file) does this correctly:

```rust
crate::provider_catalog::openai_compatible_profile_by_id(provider_name.trim())
    .is_some_and(|profile| profile.requires_api_key)
```

So the asymmetry is: local sessions bill correctly, remote (default TUI) sessions don't.

### Verified

- The models.dev catalog resolves DeepSeek correctly (`deepseek-v4-flash` = $0.14 input / $0.28 output / $0.0028 cache-read) via `model_pricing::lookup`.
- The problem is purely the `billed` gate above returning false for the display-name `"DeepSeek"`.

### Proposed fix

Apply the same profile-aware `requires_api_key` check the local path uses to the remote path. Resolve by id OR display name, because the daemon sends the display name (no `openai-compatible:` prefix):

```rust
let billed = if is_anthropic || is_openai {
    api_key_billed
} else {
    // Providers that are inherently cost-based when proxied remotely.
    provider_name.contains("opencode")
        || provider_name.contains("openrouter")
        || provider_name.contains("bedrock")
        || provider_name.contains("cerebras")
        || provider_name.contains("compatible")
        // Any OpenAI-compatible profile that requires an API key meters
        // per token when proxied remotely (DeepSeek, Z.AI, NVIDIA NIM,
        // Moonshot, ...). The daemon reports the profile's display name,
        // which does not carry the `openai-compatible:` prefix, so resolve
        // by id OR display name and honor `requires_api_key` exactly like
        // the local cost path does.
        || crate::provider_catalog::openai_compatible_profile_id_for_display_name(
            &provider_name,
        )
        .is_some_and(|profile_id| {
            crate::provider_catalog::openai_compatible_profile_by_id(profile_id)
                .is_some_and(|profile| profile.requires_api_key)
        })
};
```

Both helpers already exist and are `pub` in `crates/jcode-base/src/provider_catalog.rs` (`openai_compatible_profile_id_for_display_name` matches id OR display name; `openai_compatible_profile_by_id(...).requires_api_key` matches the local path's semantics).

### Verification of the fix
Verified against the real profile catalog via an e2e probe:

| provider_name | before | after |
|---|---|---|
| `DeepSeek` / `deepseek` | false | **true** |
| `Z.AI`, `Moonshot AI` | false | **true** |
| `ollama` (no API key) | false | false |
| `OpenRouter`, `OpenAI`, `Claude` | true | true |

### Environment
- jcode v0.68.0, Linux x86_64, DeepSeek provider.
- Full `cargo build` on this machine is blocked by a missing `libssl-dev` (unrelated to this change); the change only adds `pub` helper calls already used elsewhere.

### LINKED ISSUE #832: Ctrl+K does not delete to end of line in remote (server/client) sessions — remote dispatch path lacks kill-to-end before prompt_jump
## Summary

Ctrl+K does not delete to end of line in **remote (server/client) sessions**, even though it works in local sessions. A "⌨ Ctrl+K → delete to the end of the input" hint is shown, but the actual behavior is *prompt jump* (scroll to previous prompt).

## Environment

- jcode v0.68.0 (fcf53909f), macOS
- Running as `jcode serve` + `jcode connect` (server/client mode; log confirms `provider=remote`)

## Reproduction

1. Start a server (`jcode serve`) and attach a client (`jcode connect`).
2. Type a draft in the input box, move the cursor to the middle.
3. Press Ctrl+K.

**Expected:** text from the cursor to the end of the input is deleted (kill to end of line, Emacs habit).

**Actual:** the cursor jumps to the previous prompt. A transient hint "Ctrl+K → delete to the end of the input" appears first, but the delete never happens.

Note: Ctrl+A / Ctrl+E / Ctrl+U / Ctrl+W all work correctly in the same remote session.

## Root cause

jcode has **two** key-dispatch paths, and only the local one implements kill-to-end:

- **Local path** — `crates/jcode-tui/src/tui/app/input.rs`:
  - `handle_key_core` calls `handle_pre_control_shortcuts` (which handles plain Ctrl+K → `delete_input_to_end` for non-empty drafts, with an explicit Shift guard so Ctrl+Shift+K still scrolls) **before** the navigation/scroll handlers that claim Ctrl+K via `prompt_jump`.
  - Relevant code: `handle_pre_control_shortcuts` around line 2212, `delete_input_to_end` around line 1954.

- **Remote path** — `crates/jcode-tui/src/tui/app/remote/key_handling.rs`:
  - There is **no** `handle_pre_control_shortcuts` equivalent. `prompt_jump` is reached first (around line 558), and it claims Ctrl+K as "previous prompt" whenever the input has any navigation modifier (Ctrl included), so `delete_input_to_end` is never reached.
  - Confirmed identical in v0.68.0 and current master (v0.70.1).

The hint appears because `observe_known_hotkey` runs early in the remote handler and resolves Ctrl+K via the contextual lookup (`hotkey_feedback.rs`, `contextual_lookup`: "Plain Ctrl+K kills to end of line while a draft exists"), which reflects the *intended* behavior — so the UI advertises kill-to-end while the dispatch actually performs prompt jump.

## Suggested fix

Mirror the local path in `remote/key_handling.rs` before the `prompt_jump` check:

```rust
// Plain Ctrl+K kills to end of line (mirrors input.rs handle_pre_control_shortcuts).
// Must run before prompt_jump, which would otherwise claim Ctrl+K.
if modifiers.contains(KeyModifiers::CONTROL)
    && !modifiers.contains(KeyModifiers::SHIFT)
    && matches!(code, KeyCode::Char('k'))
    && !app.input.is_empty()
{
    input::delete_input_to_end(app);
    return Ok(());
}
```

Same for the disconnected path in `remote.rs` (`handle_disconnected_key_internal` calls `handle_navigation_shortcuts` first, which also claims Ctrl+K via `prompt_jump` before `handle_control_key` runs).

### LINKED ISSUE #838: Windows installer fails when $env:TEMP uses an 8.3 short path
## Description

The Windows installer aborts while cleaning up its temporary directory when `$env:TEMP` points to an 8.3 short path.

My Windows user profile directory contains a space:

```text
C:\Users\David X
```

In the affected PowerShell session, the temporary directory resolves through the corresponding short path:

```text
C:\Users\DAVIDX~1\...
```

The prebuilt binary downloads and passes SHA-256 verification successfully, but the installer then fails at the `Remove-Item` call in its `finally` block.

This is the normal prebuilt installation path. I did not use `-BuildFromSource`.

## Steps to reproduce

Run the standard installer in the affected environment:

```powershell
irm https://jcode.sh/install.ps1 | iex
```

Output:

```text
PS C:\Users\David X> irm https://jcode.sh/install.ps1 | iex
Fetching latest release...
Installing jcode v0.68.0
  launcher: C:\Users\David X\AppData\Local\jcode\bin\jcode.exe
Downloading jcode-windows-x86_64.exe from https://github.com/1jehuang/jcode/releases/download/v0.68.0...
Verified SHA256: jcode-windows-x86_64.exe
Remove-Item:
Line |
1094 |      Remove-Item -Path $TempDir -Recurse -Force -ErrorAction SilentlyC …
     |      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | An object at the specified path C:\Users\DAVIDX~1 does not exist.
```

## Expected behavior

The installer should complete successfully even when `$env:TEMP` uses an 8.3 short path.

A failure to clean up the temporary directory should also not abort an otherwise successful installation or prevent subsequent steps such as updating the user `PATH`.

## Cause

The installer creates its temporary directory from `$env:TEMP`:

```powershell
$TempDir = Join-Path $env:TEMP "jcode-install-$(Get-Random)"
New-Item -ItemType Directory -Path $TempDir -Force | Out-Null
```

It later removes that directory directly by path:

```powershell
finally {
    Remove-Item -Path $TempDir -Recurse -Force -ErrorAction SilentlyContinue
}
```

PowerShell has an existing FileSystemProvider bug where `Remove-Item -Path` can fail to normalize paths under `$env:TEMP` when an 8.3 short path is involved:

https://github.com/PowerShell/PowerShell/issues/21070

The upstream issue produces essentially the same error:

```text
An object at the specified path C:\Users\XXXXXX~1 does not exist.
```

## Confirmed workaround

Overriding `TEMP` and `TMP` with the long path in the current PowerShell process allows the installation to complete successfully:

```powershell
$oldTemp = $env:TEMP
$oldTmp  = $env:TMP

$longTemp = Join-Path $env:LOCALAPPDATA "Temp"
New-Item -ItemType Directory -LiteralPath $longTemp -Force | Out-Null

try {
    $env:TEMP = $longTemp
    $env:TMP  = $longTemp

    irm https://jcode.sh/install.ps1 | iex
}
finally {
    $env:TEMP = $oldTemp
    $env:TMP  = $oldTmp
}
```

With this workaround, jcode installs successfully.

## Possible fix

The workaround documented in the upstream PowerShell issue is to resolve the item first and pipe the resulting object to `Remove-Item`:

```powershell
Get-Item -LiteralPath $TempDir -ErrorAction SilentlyContinue |
    Remove-Item -Recurse -Force -ErrorAction SilentlyContinue
```

This could replace the current direct `Remove-Item -Path $TempDir` cleanup.

Alternatively, cleanup could use a .NET filesystem API, with cleanup errors handled separately so that they do not abort the remaining installation steps.

### LINKED ISSUE #840: Remote model picker X's out all slash-prefixed models as "OpenRouter auto" and blocks selection, even when a configured OpenAI-compatible profile serves them
**Type:** Bug
**Version:** v0.68.0
**Severity:** High (models become unselectable in `/model` for hours at a time)

## Summary

In remote sessions, models whose ids contain `/` (e.g. `vendouple/gpt-5.6-sol`, `MarcosFRG/deepseek-v4-pro`, `gggff123/glm-5.2`) get classified as OpenRouter "auto" routes and rendered with a **×** (unavailable) in the picker, which hard-blocks selection. This happens even when a configured OpenAI-compatible profile (pollinations, opencode zen, etc.) has the exact model id in its live catalog and the server can serve it fine.

## Repro context

- jcode running in **remote mode** (`remote=true` on every `model_picker_open` event in the log).
- Providers are configured as OpenAI-compatible profiles (`pollinations`, `opencode_zen`, `nvidia_nim`, ...) with live model catalogs.
- No `OPENROUTER_API_KEY` is set locally (the models come from the configured profiles, not OpenRouter).

## Observed behavior (from `~/.jcode/logs/jcode-2026-08-07.log`)

Same session, two picker states:

| State | `model_picker_open` event | Available |
|---|---|---|
| Full catalog (detailed routes attached) | `entries=751 entries_available=532` (13:00–14:53) | Mostly selectable |
| Names-only catalog (no route frame) | `entries=1225 entries_available=243`, `by_provider=auto:804` (12:43) | **804 slash-model rows X'd** |
| Names-only catalog (no route frame) | `entries=1528 entries_available=275`, `by_provider=auto:1074` (23:30) | **1074 slash-model rows X'd** |

The pattern "models get X'd after a few hours": the session runs on the good full catalog, then a names-only catalog update/refresh race drops the detailed route frame, the picker falls back to `remote_model_routes_fallback`, and every slash-prefixed model becomes unselectable.

## Root cause

`jcode-base/src/provider/catalog_routes.rs`, `remote_model_routes_fallback` (lines 841–1053):

1. **Any model id containing `/` is assumed to be an OpenRouter model** (lines 892–920): it's routed to `build_openrouter_auto_route(model, auth.openrouter != AuthState::NotConfigured, ...)`, so with no local `OPENROUTER_API_KEY` the route is `available: false` with no detail.
2. That branch `continue`s **before** `remote_openai_compatible_route_for_model(model)` (line 1006) is ever consulted, so a configured profile whose live catalog contains the exact id is never given a chance to produce an available route.
3. `remote_current_openai_compatible_route_for_model` (lines 1102–1127) also bails on `/`-containing ids (`if model.contains('/') { return None }`), so the "current provider" fallback can't help either.
4. The TUI then hard-blocks selection on any `!route.available` row: `jcode-tui/src/tui/app/inline_interactive.rs` lines 3348–3359 (`PickerAction::Model` → `unavailable_model_route_message` → "Model unavailable"), with the × rendered by `picker_row_marker` in `jcode-tui/src/tui/ui_inline_interactive.rs` lines 115–125.

The client mislabels working models as unconfigured OpenRouter purely because of the `/` heuristic; the server-side catalog (pollinations, opencode zen) contains them and serves them fine.

## Suggested fix

In `remote_model_routes_fallback`, before the `model.contains('/')` → `build_openrouter_auto_route` branch, consult `remote_openai_compatible_route_for_model(model)` first:

- If a configured profile's live cache (or static model list) contains the exact id → emit that profile's route with `available: true` and `continue`.
- Only fall back to the OpenRouter "auto" route when no profile matches (and mark unavailable only then, if OpenRouter is genuinely unconfigured).

Also relax `remote_current_openai_compatible_route_for_model` so slash-ids are matched when the current provider profile's catalog actually contains them (instead of blanket-skipping all `/` ids).

Optionally: consider not hard-blocking selection on `!available` rows — show the reason but still let the user attempt the switch when a route exists (the server is the real authority on whether a model is servable).

## Workarounds (no rebuild needed)

- Type `/model <name>` instead of using the picker — the typed path (`backend.rs:692`, `Request::SetModel`) sends the bare id to the server with no availability gate.
- Set a dummy `OPENROUTER_API_KEY` locally — the availability check is presence-only, so the picker rows flip to selectable; the dummy key is never used since selection still sends the bare model id.

1jehuang:
## Summary
- route slash-prefixed models through matching OpenAI-compatible profiles
- tolerate Windows 8.3 temporary paths during installer cleanup
- restore remote Ctrl+K draft editing and cancellable rate-limit retries
- meter API-key-backed compatible providers in the remote cost widget
- keep multiline tool errors out of focused memory queries

## Verification
- exact jcode-base catalog and memory regression tests pass
- exact jcode-tui billing, connected/disconnected Ctrl+K, and retry-cancellation tests pass
- `install.ps1` parses successfully in PowerShell
- clean branch diff and worktree checks pass

Fixes #840. Fixes #838. Fixes #832. Fixes #831. Fixes #826. Fixes #824.


1jehuang:
Superseded by #842, which merged these six verified issue fixes into `master`. Closing this now-redundant PR to avoid duplicate changes.

PR #856, opened 2026-08-09
fix: resolve safe issues from automated triage

### LINKED ISSUE #847: Named custom openai-compatible models without `input` are assumed to support images
## Summary

For named custom `open-ai-compatible` provider profiles, a model that does not declare an
`input` list is assumed to support image input. When a tool (e.g. `read` on an image file)
puts an image into conversation history, subsequent requests include `image_url` and
text-only endpoints reject them with HTTP 400.

The failure persists for as long as the config is unchanged: every following turn re-sends
the image from history, and `/compact` does not help, because the compaction request itself
carries the same image. There is no in-app way to drop the image. Recovery requires editing
the provider config to declare `input` and restarting jcode; after that the existing session
continues normally, because the image is stripped per request. The trigger is easy to hit
accidentally, since `read` accepts image paths.

## Environment

- jcode v0.71.1, linux-x86_64
- Provider type: `open-ai-compatible`, self-hosted internal gateway

## Reproduction

Config:

```toml
[providers.my-gateway]
type = "open-ai-compatible"
base_url = "https://internal-gateway.example.com/v1"
default_model = "my-text-model"

[[providers.my-gateway.models]]
id = "my-text-model"
```

1. Start a session and ask the agent to `read` an image file.
2. Send any follow-up message.
3. Request fails:

```
status: 400 Bad Request
response: {"object":"error","message":"Model only supports text input; received unsupported
content type 'image_url'.","type":"BadRequestError","param":null,"code":400}
```

## A/B verification

All runs on the same binary (v0.71.1) and the same two-turn flow
(turn 1: `read` an image; turn 2: send a trivial text message).
The only variable is the `input` key.

| Config | Turn 2 result |
| --- | --- |
| `input = ["text"]` | success |
| `input` omitted | 400 `Model only supports text input` |
| `input = ["banana"]` | success |
| `input` omitted, then `/compact` | still 400 |

Every turn in the affected session keeps failing until `input` is declared in config and
jcode is restarted; the same session then works, since stripping is applied per request.

## Cause

Two mechanisms combine.

**1. Models without `input` are absent from the capability map.**
`static_image_input_support` is populated only in `new_named_openai_compatible`
(`crates/jcode-provider-openrouter-runtime/src/lib.rs:1324-1338`), and models with an empty
`input` are skipped outright:

```rust
let id = model.id.trim();
if id.is_empty() || model.input.is_empty() {
    return None;
}
let supports_images = model
    .input
    .iter()
    .any(|input| input.eq_ignore_ascii_case("image"));
Some((id.to_ascii_lowercase(), supports_images))
```

A model with no `input` gets no entry at all, so the lookup misses rather than returning
`false`.

**2. The fallback then reports image support.**
`supports_image_input()` in
`crates/jcode-provider-openrouter-runtime/src/openrouter_provider_impl.rs:360-365`
falls back to `!self.supports_provider_features`. For a named profile whose
`provider_type` is not `OpenRouter` and which sets neither `provider_routing` nor
`allow_provider_pinning`, `supports_provider_features` is `false`
(`lib.rs:1348-1352`), so the negation yields `true` and the image is not stripped.

This explains all three rows of the table: `["text"]` and `["banana"]` both create an entry
resolving to `false` and the image is stripped, while an omitted `input` skips the map and
hits the permissive fallback.

The bug does not depend on the model family, only on whether `input` is declared.

## Note on the existing fallback

The fallback is deliberate and documented: the comment at
`openrouter_provider_impl.rs:360-365` explains that direct openai-compatible local
providers such as Ollama and LM Studio document image support on
`/v1/chat/completions`, and that the legacy OpenRouter aggregator behaviour is
intentionally left unchanged.

That reasoning holds for those cases. The problem is that named custom profiles always
evaluate `supports_provider_features` to `false`, so they inherit the local-provider
assumption even when the endpoint is a plain text-only gateway.

## Secondary finding: unknown `input` values are silently accepted

`eq_ignore_ascii_case("image")` at `lib.rs:1335` is the only place a declared modality
string is compared against `image`. (`route_builders.rs:62` matches `image` as a token in
model *names*, and `catalog_routes.rs:502` compares `input` against `text`; neither
validates declared modalities.) Anything that is not `image` is treated as "no image
support" with no warning or config error, so `input = ["banana"]` silently behaves like
`input = ["text"]`. A typo in a modality name is therefore invisible, which made the
original bug harder to isolate.

## Suggested fix

Preferred: for named custom openai-compatible profiles, default models with no declared
`input` to text-only, and make image support opt-in via `input = ["text", "image"]`. This
fails safe. An endpoint that does support images merely degrades to placeholder text,
whereas the current default breaks every turn of the session until the user discovers the
config key. Local-provider and OpenRouter paths can keep their present behaviour, since
they do not populate this map.

Additionally: validate `input` entries against a known modality allowlist and raise a
config error on unknown values.

If reversing the default is considered too breaking, an alternative is to classify this 400
as an image-capability error and retry once with images stripped, caching the result for
the rest of the session.

I am happy to open a PR once you indicate which direction you prefer, since changing the
default is potentially breaking for users who rely on the current implicit behaviour.

### LINKED ISSUE #849: ACP model switcher fails with `OPENROUTER_API_KEY not found` when the active provider is a catalog OpenAI-compatible profile (e.g. `opencode-go`)
## Summary

When jcode is used as a Zed external agent over ACP, the model selector exposed via
`configOptions` / `session/set_config_option` lists models from the active
OpenAI-compatible profile's catalog, but switching to any of them fails with
`OPENROUTER_API_KEY not found in environment or ~/.config/jcode/openrouter.env`.
The session silently stays on the fallback model (`deepseek-v4-pro`). The advertised
models are unselectable, and the host receives a JSON-RPC error instead of a model change.

## Environment

- jcode v0.71.1 (dd8755f7e), Linux x86_64
- Host: Zed via ACP (`client_instance_id=acp`)
- Active provider: `opencode-go` catalog profile → `https://opencode.ai/zen/go/v1`,
  auth `OPENCODE_GO_API_KEY` (in `~/.config/jcode/opencode-go.env`); `[providers]`
  empty in `config.toml`
- `OPENROUTER_API_KEY` is not configured anywhere (no env var, no
  `~/.config/jcode/openrouter.env`)

## Steps to reproduce

1. Configure only the `opencode-go` profile (`jcode login --provider opencode-go`)
   and no OpenRouter key.
2. Start a session in Zed with `jcode acp` as the external agent. The default model
   `deepseek-v4-pro` works (streams to `https://opencode.ai/zen/go/v1`).
3. Open the agent panel model selector (populated from ACP `configOptions`) and pick
   `deepseek-v4-flash`.
4. Zed sends `session/set_config_option` (`configId: "model"`, `value: "deepseek-v4-flash"`);
   jcode logs `server_model_change_failed` and the model stays `deepseek-v4-pro`.

## Expected behavior

Switching to a model that the active profile's own catalog already serves should
select it on that same profile (same endpoint and credentials). The selector should
only present models that can actually be activated.

## Actual behavior

Every switch to a non-default model fails:

```
[ERROR] EVENT event=server_model_change_failed error="OPENROUTER_API_KEY not found in environment or /home/xxx/.config/jcode/openrouter.env (retried after reloading auth from disk: OPENROUTER_API_KEY not fou..." fallback_model=deepseek-v4-pro
```

The ACP client receives a JSON-RPC error (`Failed to set model: ...`) and the session
keeps running `deepseek-v4-pro`.

## Root cause

The ACP layer sends bare model ids. `session_config_options`
(`src/cli/acp.rs:1015`) builds option values from `available_models` (deduped bare
ids from `listable_model_names_from_routes`), and `handle_set_config_option`
(`src/cli/acp.rs:506`) forwards `Request::SetModel { model: "deepseek-v4-flash" }`
verbatim.

In `MultiProvider::set_model` (`crates/jcode-base/src/provider/mod.rs`), the
unprefixed id first hits the "keep unprefixed names on that active endpoint" branch
(~L1922):

```rust
if self.active_provider() == ActiveProvider::OpenRouter
    && self.active_openrouter_execution_provider()
        .is_some_and(|provider| !provider.supports_provider_routing_features())
{
    return self.set_model_on_provider(ActiveProvider::OpenRouter, requested_model);
}
```

That branch intends to keep bare ids on the active OpenAI-compatible endpoint, but
the OpenRouter arm it calls then computes `needs_rebind`
(`provider/mod.rs:1138`) as `true` whenever the slot is an OpenAI-compatible
**catalog profile** runtime whose profile id is not the generic `openai-compatible`
profile (only `OPENAI_COMPAT_PROFILE.id` is exempt):

```rust
profile.id != crate::provider_catalog::OPENAI_COMPAT_PROFILE.id
```

The slot is therefore rebound to the real OpenRouter API-key runtime via
`new_openrouter_api_key_runtime()` (`crates/jcode-provider-openrouter-runtime/src/lib.rs:1582`),
which fails when `OPENROUTER_API_KEY` is absent, even though the requested model is
already served by the active profile's own catalog. `set_model_with_auth_refresh`
retries once after reloading auth from disk, fails again, and `send_model_changed_result`
logs `server_model_change_failed` with the previous model as fallback.

The TUI picker avoids this by emitting route-qualified specs (`<profile>:<model>`,
see issue #444); the ACP path, `/model <bare-id>` and session-resume paths send bare
ids and hit the rebind. Confirmed present on both master and v0.71.1.

## Suggested fix

1. **Resolver (recommended, B1):** in the OpenRouter arm of
   `set_model_on_provider_with_credential_modes`, skip the rebind when the active
   profile's own route catalog already serves the requested model
   (`route.model == model` with `api_method == "openai-compatible:<active profile>"`
   and `available`), and just call `set_model` on the existing slot. This matches the
   intent of the "keep unprefixed names on the active endpoint" branch and fixes every
   bare-id path (ACP, `/model <id>`, session resume), not just the ACP surface.
2. **ACP layer (B2):** build `configOptions` model values as profile-qualified specs
   (`<profile>:<model>`, e.g. `opencode-go:deepseek-v4-flash`) with the bare id as
   display name, deduplicating in favor of the active profile (same preference rule as
   `openai_compatible_profile_owning_model`). This is the server-side analog of the
   workaround used in #813.

## Related issues

- #813 documents the same underlying resolver problem as a side note of an ACP
  feature request for hermes-family hosts (`session/set_model`). Filed separately
  because this is a bug in the already-shipped `configOptions` path used by Zed, and
  the resolver fix (B1) is independent of the ACP contract decision.
- #765 (closed) introduced the `configOptions` model selector that Zed uses.

### LINKED ISSUE #852: `jcode` command fails on Linux when installed with Homebrew
I installed the jcode on Linux with Homebrew. The application fails when I run the `jcode` command. The wrapper script passes an empty subcommand to the binary file.

## Expected Behavior

The `jcode` command opens the tui.

## Actual Behavior

The command stops and prints an error.

```
~/ ❱ jcode
error: unrecognized subcommand ''

Usage: jcode-linux-x86_64.bin [OPTIONS] [COMMAND]

For more information, try '--help'.
```

## Steps to Reproduce

1. If you use Linux, install the jcode package with Homebrew.
2. Run the `jcode` command in a terminal.

## Environment

* Operating System: Fedora Linux 44 (Workstation Edition)
* Installation Method: Homebrew

### LINKED ISSUE #854: swarm tool surfaces raw 'No such file or directory (os error 2)' when the background server isn't running (v0.71.1)
When `jcode run` (headless) invokes the swarm tool with no background server running, every swarm call fails instantly (0ms) with the raw OS error:

```
Failed to list agents: No such file or directory (os error 2)
Failed to list channels: No such file or directory (os error 2)
Failed to read shared context: No such file or directory (os error 2)
```

This is just the daemon socket being absent, but the message gives the user no hint. With `jcode server start` first, the same calls produce the proper actionable errors ("Not in a swarm. Use a git repository to enable swarm features." / "No shared context found.").

Suggestion: detect the missing/unconnectable server socket in the swarm tool path and return something like "jcode server is not running — start it with `jcode server start`" instead of the raw ENOENT. In multi-agent setups where headless sessions are spawned by automation, the raw error sent three agents on a multi-hour goose chase (v0.71.1, dd8755f7e, macOS arm64).

### LINKED ISSUE #855: [bug] Copilot provider: built-in `swarm` tool makes every request fail with HTTP 400 Bad Request
## Summary

On the GitHub Copilot provider, exposing the built-in `swarm` tool makes **every** request fail with `HTTP 400 Bad Request` (bare body, no JSON error). Because `swarm` is in the default toolset, this bricks Copilot sessions entirely — no turn can complete.

Disabling one tool fixes it:

```toml
[tools]
disabled = ["swarm"]
```

## Environment

- jcode `v0.72.0` (also reproduced on `v0.64.2`)
- Windows 11, x86_64
- Provider: `copilot` (`format=chat_completions`, endpoint `api.enterprise.githubcopilot.com`)
- Models: reproduced on `claude-opus-4.6` and `claude-opus-5`

## Reproduction

```bash
jcode run --tools swarm "hi"
# Error: Copilot API error (HTTP 400 Bad Request): Bad Request

jcode run --tools memory "hi"    # works
jcode run --tools browser "hi"   # works
jcode run --disabled-tools swarm "hi"   # works, full toolset otherwise
```

## Isolation already done

I bisected this fairly thoroughly, so hopefully this saves time:

- **Not auth.** A hand-rolled `curl` to `/chat/completions` with the same bearer returns 200. `jcode provider-doctor copilot` passes every checkpoint.
- **Not MCP.** Reproduced with all MCP servers removed. MCP tools are namespaced `mcp__<server>__<tool>`, so they cannot collide with built-ins. Every MCP tool schema also passes individually and collectively.
- **Not size.** 29 built-ins alone fail; 9 (`--tool-profile minimal`) succeed. Tool-description lengths up to 32 KB return 200 on this endpoint.
- **It is specifically `swarm`.** Testing built-ins one at a time, `swarm` is the only one that 400s alone.

## What is different about the Copilot path

The same `swarm` tool works on the `openai-compatible` provider. I captured both requests:

| Provider path | `tools_json_chars` for the single `swarm` tool | Result |
|---|---|---|
| `openai-compatible` | 7691 | 200 |
| `copilot` | 8403 | **400** |

I replayed the captured `openai-compatible` payload verbatim against the Copilot endpoint (only swapping `model`) and it returned **200** — including with `stream: true`, `tool_choice`, and the full 8-message history.

So the schema itself is acceptable to Copilot. Something in the Copilot request builder adds roughly **712 characters** to this tool definition that the openai-compatible builder does not, and that addition is what the endpoint rejects.

For reference, the Copilot endpoint returns this same bare `Bad Request` (no JSON body) for: duplicate tool names, tool names containing `.`, tool names longer than 64 chars, and empty tool descriptions. A well-formed rejection returns JSON instead, e.g. `{"error":{"message":"model ... is not accessible via the /chat/completions endpoint"}}`. That suggests the added content trips schema validation at the edge rather than the model router.

`swarm` has 43 properties and is by far the largest built-in schema, so it may simply be the first tool to cross a threshold that the extra Copilot-side content pushes it over.

## Suggested fix

Diff the Copilot provider's tool serialization against the openai-compatible one for the same tool and check what the extra ~712 chars are.

## Workaround for anyone hitting this

```toml
[tools]
disabled = ["swarm"]
```

Copilot sessions then work normally with the rest of the toolset.

1jehuang:
## Summary

- normalize Copilot chat-completions tool schemas so `swarm` is accepted
- make missing swarm daemon errors actionable
- keep bare model switches on the active named catalog profile
- default unspecified custom model inputs to text-only
- preserve Homebrew launcher arguments during formula generation

## Verification

- integrated TUI build passed with `scripts/dev_cargo.sh build --profile selfdev -p jcode --bin jcode`
- Copilot runtime tests: 31 passed, including 2 new regressions
- provider resolver suite: 117 passed
- custom model capability tests: 8 passed
- swarm socket tests: 3 passed
- generated Homebrew formula line verified to retain literal `$@`

Fixes #855
Fixes #854
Fixes #852
Fixes #849
Fixes #847


1jehuang:
Quality Guardrails exposed a pre-existing Clippy `collapsible_if` failure in connection-message rendering. I fixed it in `9dcd6114a`; the exact `cargo fmt --all -- --check` and `cargo clippy --all-targets --all-features -- -D warnings` gates now pass locally.

1jehuang:
The rerun passes Clippy and reaches the oversized-file ratchet, which still fails. I reproduced that ratchet on an untouched detached `origin/master` worktree: the same nine files already exceed the checked-in baseline there. PR #856 adds 30 lines to one listed file (`provider/mod.rs`) but does not cause the other eight default-branch violations. Updating the global ratchet or refactoring nine unrelated files would be outside this triage PR, so this is now the only known hosted-check blocker.

PR #876, opened 2026-08-10
Fix remote session state after clear and skill activation

### LINKED ISSUE #873: Skill body never reaches the model for /skill invocations in client/server sessions (UI says "Activated skill", agent gets a bare prompt)
## Summary

In a client/server session (`jcode serve` + TUI client), invoking a skill as `/skill <prompt>`
shows **"Activated skill: …"** in the UI, but the skill body never reaches the model. The agent
receives only the trailing prompt, and the `/skill` token is stripped from the stored message too,
so nothing downstream can tell a skill invocation happened.

From the model's side, `/grill-with-docs do the thing` is indistinguishable from `do the thing`.

## Reproduction

jcode v0.75.0 (5ae238574), Linux x86_64, TUI client against a `jcode serve --socket …` daemon.

```
/grill-with-docs do you see a SKILL body for grill-with-docs in your system prompt? answer YES or NO and quote its first line
```

**Observed** — UI prints `Activated skill: grill-with-docs - A relentless interview to sharpen a
plan or design, which also creates docs (ADR's and glossary) as we go.`, and the model replies:

> NO.
>
> I only have the one-line catalog entry, not a skill body. Its first line reads:
> `/grill-with-docs  - A relentless interview to sharpen a plan or design, …`
>
> To load the actual body, I must run the Skill tool.

**Expected** — the model's system prompt contains `# Active Skill\n\n# Skill: grill-with-docs …`,
i.e. the `SKILL.md` content.

The one-line text the model *does* see is the `# Available Skills` catalog entry that
`build_system_prompt_split_with_capabilities` adds for every registered skill, so this is not
partial delivery. Nothing skill-specific arrives at all.

## Impact

Silent, and the failure mode is bad: the UI affirms the skill is active, so the operator assumes
the agent is following it. The agent, having received a bare prompt, does something else entirely.
In my case that cost roughly 200k tokens and half an hour of work in the wrong direction before
either of us noticed.

Every `/skill` invocation in a remote/daemon session is affected, which on this machine is all of
them.

## Cause

The skill body reaches the model only by being written into the **local** system prompt:

- `crates/jcode-tui/src/tui/app/turn_memory.rs:17-21` reads `self.active_skill` and passes
  `skill_prompt` to `crate::prompt::build_system_prompt_split`.
- `crates/jcode-base/src/prompt.rs:541-544` appends it as `# Active Skill\n\n{skill}` to the
  dynamic part.

That path is only taken by the **local** run loop (`turn.rs:83`). In a remote session the prompt
is built server-side, and the request that carries the turn has nowhere to put the skill:

```rust
// crates/jcode-protocol/src/wire.rs:39-49
Request::Message {
    id: u64,
    content: String,
    images: Vec<(String, String)>,
    system_reminder: Option<String>,
}
```

There is no `active_skill` field, and `active_skill` appears nowhere in `jcode-protocol`,
`jcode-message-types` or `jcode-session-types`. The client sets its own `App::active_skill`
(`crates/jcode-tui/src/tui/app/input.rs:3691`) and the server sets its own
`Agent::active_skill` (`crates/jcode-app-core/src/agent/turn_execution.rs:781`), but only from
its own CLI loop. The two are never connected.

Meanwhile the client has already consumed the marker:

```rust
// crates/jcode-tui/src/tui/app/input.rs:3691-3701
self.active_skill = Some(skill_name.clone());
self.push_display_message(/* "Activated skill: …" */);
if let Some(prompt) = trailing_prompt {
    input = prompt;          // <- the /name token is dropped here
} else {
    return;
}
```

`crates/jcode-tui/src/tui/app/remote/input_dispatch.rs:203-216` then deliberately re-derives just
the trailing prompt for the remote request. So the server is sent `do the thing`, with no way to
learn which skill was meant.

Confirmed in storage — `~/.jcode/sessions/session_<id>.json` holds the message as:

```json
{"type": "text", "text": "regarding this please /tmp/claudistan-harness-agnostic-handoff.md"}
```

for an input that was typed as `/grill-with-docs regarding this please /tmp/…`.

## Suggested fix

Carry the active skill on the wire. Add an optional field to `Request::Message` (and
`SoftInterrupt`), e.g. `active_skill: Option<String>`, and have the server resolve it against its
own registry when building the system prompt — the server already owns
`Agent::active_skill` and `prompting.rs` already consumes it, so only the plumbing is missing.

Sending the resolved body instead of the name would also work and would survive the client and
server disagreeing about which skills exist, at the cost of a larger payload on every turn.

Whichever is chosen, a regression test that asserts the rendered system prompt contains
`# Active Skill` after a **remote** `/skill x` invocation would pin it. The existing coverage
(`app/tests.rs:972`, `remote_tests.rs:504`) only asserts `active_skill` is set on the client,
which is true today and not sufficient.

## Secondary issue: bare `/skill` with no prompt

When there is no trailing prompt the client `return`s early
(`input.rs:3699-3700`), so no turn is sent at all. The skill is armed for the *next* message. That
is defensible, but combined with the above it means a bare `/skill` and a `/skill <prompt>` fail in
two different ways, neither visible.

## Note on a natural workaround that does not work

Putting the slash at the end — `do the thing /grill-with-docs` — silently does nothing at all, not
even the UI notice. `SkillRegistry::resolve_invocation` requires the slash at position 0:

```rust
// crates/jcode-base/src/skill/invocation.rs:16
let invocation = input.trim().strip_prefix('/')?;
```

Worth a mention in the docs, since it is the first thing an operator tries.

The working workaround today is to call the `Skill` **tool** rather than the slash command, or to
paste the `SKILL.md` body inline.

### LINKED ISSUE #874: swarm tool fails with 'Not in a swarm. Use a git repository to enable swarm features.' after /clear (member never re-registered)
## Summary

After `/clear`, every `swarm` tool action that needs membership (`spawn`, `list`, `start`, …) fails with:

```
Not in a swarm. Use a git repository to enable swarm features.
```

The message points at git, but the repo is a valid git worktree (`git rev-parse --is-inside-work-tree` → `true`) and `features.swarm = true`. The actual cause is that the session created by `/clear` is never registered as a `SwarmMember`.

Version: v0.75.0 (5ae238574), Windows x86_64.

## Root cause

In `crates/jcode-app-core/src/server/client_session.rs`, `handle_clear_session`:

1. Removes the old session's member entry (intentional, per the comment "Do not migrate the old session's swarm membership"):
   ```rust
   let swarm_id_for_update = {
       let mut members = swarm_members.write().await;
       members.remove(client_session_id).and_then(|member| member.swarm_id)
   };
   ```
2. Then calls `update_member_status(&new_id, "ready", …)` — but `update_member_status_with_report_tldr` (`swarm.rs`) only does `members.get_mut(session_id)` and **silently no-ops in the `else` branch when the member does not exist**. It never inserts.
3. The only function that *creates* a member, `ensure_client_swarm_member`, is called exclusively from `handle_subscribe`. `/clear` never re-runs it, and the connection stays subscribed, so no new subscribe happens.
4. The `clear_done` log then reports `swarm_id_updated=true`, which is misleading: it only means the *old* member had a swarm id to remove, not that the new session was registered.

Result: the new session has no `SwarmMember` entry, so every `swarm_id_for_session(id)` → `members.get(id).and_then(|m| m.swarm_id)` returns `None`, and all comm handlers answer "Not in a swarm."

## Evidence (from `~/.jcode/logs/jcode-2026-08-10.log`)

```
17:43:37 EVENT event=SESSION_LIFECYCLE … phase=clear_done old_session_id=session_guppy_… new_session_id=session_skunk_… swarm_id_updated=true
17:45:00 EVENT event=TOOL_LIFECYCLE … error="Not in a swarm." (swarm spawn)
17:45:08 EVENT event=TOOL_LIFECYCLE … error="Not in a swarm. Use a git repository to enable swarm features." (swarm list)
```

Every `phase=swarm_member_registered` event of the day belongs to sessions that went through `handle_subscribe`; none names the `/clear`-born session. Its state file `~/.jcode/state/swarm/session_session_skunk_….json` does not exist.

Checked `5ae238574..origin/master` on `client_session.rs`: no later commit fixes this. #816's fix (reset swarm plan state on clear, 5496ebb5) is an ancestor of v0.75.0 and doesn't cover member re-registration.

## Repro

1. Open a jcode TUI session in a git repo, use any `swarm` action (works).
2. `/clear`.
3. `swarm list` → "Not in a swarm. Use a git repository to enable swarm features."

Workaround: quitting the client and `jcode resume` on the same session re-registers the member via `handle_subscribe`.

## Suggested fix

In `handle_clear_session`, after swapping in the fresh agent and before `update_member_status`, re-register the replacement session, e.g. call `ensure_client_swarm_member(&new_id, client_connection_id, …, swarm_enabled, …)` (the old member's `swarm_enabled` can be captured before the `remove`). Alternatively, make the `update_member_status` no-op branch log a warning so the next silent drop of this kind is visible.

Also worth a tweak: the "Use a git repository to enable swarm features" wording sends users debugging git/config when membership is the actual gate.

1jehuang:
## Summary
- re-register the fresh session as a new swarm root after `/clear`, without restoring old plan participation
- carry the selected skill name on remote message requests and resolve it against the daemon registry before starting the turn
- reject client/server skill registry mismatches instead of silently running without the requested skill

## Verification
- `cargo test -p jcode-app-core handle_clear_session_replaces_runtime_handles_and_updates_shutdown_registration --lib`
- `cargo test -p jcode-protocol test_message_request_roundtrip_preserves_images_and_system_reminder --lib`
- `cargo test -p jcode-tui remote_skill_invocation_with_prompt_sends_remote_turn --lib`
- `cargo check -p jcode-protocol -p jcode-app-core -p jcode-tui --all-targets`

Fixes #874
Fixes #873

### DIFF
--- a/crates/jcode-app-core/src/agent.rs
+++ b/crates/jcode-app-core/src/agent.rs
@@ -570,6 +570,17 @@ impl Agent {
         self.rewind_undo_snapshot = None;
     }
 
+    /// Synchronize the remote client's selected skill, accepting only names
+    /// present in the daemon's own registry snapshot.
+    pub(super) fn set_remote_active_skill(&mut self, active_skill: Option<String>) -> bool {
+        let skills = self.current_skills_snapshot();
+        let recognized = active_skill
+            .as_ref()
+            .is_none_or(|name| skills.get(name).is_some());
+        self.active_skill = active_skill.filter(|name| skills.get(name).is_some());
+        recognized
+    }
+
     fn sync_session_compaction_state_from_manager(
         &mut self,
         manager: &crate::compaction::CompactionManager,
--- a/crates/jcode-app-core/src/server/client_api.rs
+++ b/crates/jcode-app-core/src/server/client_api.rs
@@ -51,6 +51,7 @@ impl Client {
             content: content.to_string(),
             images: vec![],
             system_reminder: None,
+            active_skill: None,
             no_reply: false,
         };
         let json = serde_json::to_string(&request)? + "\n";
--- a/crates/jcode-app-core/src/server/client_lifecycle.rs
+++ b/crates/jcode-app-core/src/server/client_lifecycle.rs
@@ -111,6 +111,7 @@ struct ProcessingMessage {
     content: String,
     images: Vec<(String, String)>,
     system_reminder: Option<String>,
+    active_skill: Option<String>,
 }
 
 struct ProcessingState<'a> {
@@ -1105,6 +1106,7 @@ pub(super) async fn handle_client(
                 content,
                 images,
                 system_reminder,
+                active_skill,
                 no_reply,
             } => {
                 if no_reply {
@@ -1133,6 +1135,7 @@ pub(super) async fn handle_client(
                         content,
                         images,
                         system_reminder,
+                        active_skill,
                     },
                     &client_session_id,
                     &mut ProcessingState {
@@ -2850,6 +2853,7 @@ async fn start_processing_message(
         content,
         images,
         system_reminder,
+        active_skill,
     } = message;
     if server_reload_starting() {
         crate::logging::info(&format!(
@@ -2864,6 +2868,20 @@ async fn start_processing_message(
         let _ = client_event_tx.send(ServerEvent::Error {
             id,
             message: "Already processing a message".to_string(),
+            retry_after_secs: None,
+        });
+        return;
+    }
+
+    if !agent
+        .lock()
+        .await
+        .set_remote_active_skill(active_skill.clone())
+    {
+        let skill_name = active_skill.as_deref().unwrap_or_default();
+        let _ = client_event_tx.send(ServerEvent::Error {
+            id,
+            message: format!("Skill '{skill_name}' is not installed on the server"),
             retry_after_secs: None,
         });
         return;
--- a/crates/jcode-app-core/src/server/client_lifecycle_tests.rs
+++ b/crates/jcode-app-core/src/server/client_lifecycle_tests.rs
@@ -909,6 +909,7 @@ fn reload_starting_rejects_new_turn_without_spawning_processing_task() {
                 content: "do not start during reload".to_string(),
                 images: Vec::new(),
                 system_reminder: None,
+                active_skill: None,
             },
             "session_guard",
             &mut ProcessingState {
@@ -1009,6 +1010,7 @@ async fn client_initiated_turn_fans_out_stream_and_terminal_events_to_live_attac
             content: "stream to every attachment".to_string(),
             images: Vec::new(),
             system_reminder: None,
+            active_skill: None,
         },
         session_id,
         &mut ProcessingState {
@@ -1133,6 +1135,7 @@ fn accepted_reload_recovery_continuation_marks_intent_delivered() -> anyhow::Res
                 content: "continue after reload".to_string(),
                 images: Vec::new(),
                 system_reminder: Some(continuation.to_string()),
+                active_skill: None,
             },
             session_id,
             &mut ProcessingState {
@@ -1232,6 +1235,7 @@ fn reload_starting_rejects_new_turns_for_multiple_sessions() {
                     content: format!("do not start {session_id} during reload"),
                     images: Vec::new(),
                     system_reminder: None,
+                    active_skill: None,
                 },
                 session_id,
                 &mut ProcessingState {
--- a/crates/jcode-app-core/src/server/client_session.rs
+++ b/crates/jcode-app-core/src/server/client_session.rs
@@ -229,11 +229,12 @@ pub(super) async fn handle_clear_session(
     // `/clear` creates a genuinely fresh session. Do not migrate the old
     // session's swarm membership or plan participation to the replacement:
     // doing so lets a subsequent plan snapshot repopulate the cleared UI.
-    let swarm_id_for_update = {
+    let (swarm_id_for_update, swarm_enabled, friendly_name) = {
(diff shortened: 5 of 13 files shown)


shaikahad-tech:
The `Windows Cross-Target Check (Linux)` job failed with a compile error because this PR adds a new `active_skill: Option<String>` field to `Request::Message` (in `crates/jcode-protocol/src/wire.rs`) but missed one call site:

```
error[E0063]: missing field `active_skill` in initializer of `jcode_tui::protocol::Request`
  --> src/cli/acp.rs:874:20
   |
874 | .send(&Request::Message {
   |                ^^^^^^^^^^^^^^^^ missing `active_skill`
```

**Fix:** add `active_skill: None,` to the `Request::Message` initializer at `src/cli/acp.rs:878`, matching what every other non-skill call site does (`client_api.rs`, `communicate_tests.rs`, `client_lifecycle_tests.rs`, `randomized.rs`):

```diff
--- a/src/cli/acp.rs
+++ b/src/cli/acp.rs
@@ -875,6 +875,7 @@
             .send(&Request::Message {
                 id: prompt_id,
                 content: text,
                 images,
                 system_reminder: None,
+                active_skill: None,
                 no_reply: false,
             })
```

The ACP (Agent Client Protocol) path has no skill concept, so `None` is the correct value.

I have a fix branch on my fork (`shaikahad-tech/jcode`, branch `fix/acp-missing-active-skill-field`) but couldn't open a PR due to the repo's `collaborators_only` PR policy. The branch is based on this PR's head commit (`fefae76`) with just this one-line addition.

PR #880, opened 2026-08-10
Fix pinned todos config-off test isolation

### LINKED ISSUE #877: CI failure: pinned_todos_payload_stays_empty_when_config_off (stale config cache leak from sibling test)
## Summary

The `Build & Test (ubuntu-latest)` job in run #31359160118 (for `chore(release): prepare v0.75.0` #2231) failed at step 13 "Run TUI library tests" with a single failing test:

```
test tui::app::tests::pinned_todos_payload_stays_empty_when_config_off ... FAILED

thread 'tui::app::tests::pinned_todos_payload_stays_empty_when_config_off' panicked at
crates/jcode-tui/src/tui/app/tests/todo_card.rs:227:5:
assertion failed: !app.refresh_pinned_todos_if_needed()

test result: FAILED. 2170 passed; 1 failed; 17 ignored; 0 measured; 2 filtered out
```

CI run: https://github.com/1jehuang/jcode/actions/runs/31359160118/job/93364458815

The "1 warning" in the same run is GitHub's Node 20 runner deprecation notice — infrastructure, not repo code.

## Root cause

This is a test-isolation bug, not a product bug.

`pinned_todos_payload_stays_empty_when_config_off` (in `crates/jcode-tui/src/tui/app/tests/todo_card.rs:219`) asserts that `refresh_pinned_todos_if_needed()` returns `false` when `display.pin_todos` is at its default (off). It fails because the jcode-base config cache throttles env re-checks for ~500ms — a behaviour already documented in this same file by `PinTodosEnvGuard::enable()` (lines 199-207):

```rust
// jcode-base's config cache throttles env re-checks (the zero
// interval under cfg!(test) applies only when jcode-base itself is
// the crate under test), so force a reload or a sibling test's
// JCODE_PIN_TODOS state leaks into this one for up to 500ms.
crate::config::invalidate_config_cache();
```

Every other pinned-todos test in the file flushes the cache via `PinTodosEnvGuard::enable()`. This test asserts the **config-off** path, so it deliberately does not enable the `JCODE_PIN_TODOS` override — but it also never invalidates the cache. When it runs shortly after a sibling test that set `JCODE_PIN_TODOS=1`, it reads the stale cached config where `pin_todos = true`, so `refresh_pinned_todos_if_needed()` (in `crates/jcode-tui/src/tui/app/todos_view.rs:84`) enters the enabled branch and returns `true`, tripping the assertion at `todo_card.rs:227`.

## Suggested fix

Add the same `crate::config::invalidate_config_cache();` call at the top of the test (before `create_test_app()`), mirroring `PinTodosEnvGuard::enable()`:

```rust
#[test]
fn pinned_todos_payload_stays_empty_when_config_off() {
    let _env_lock = crate::storage::lock_test_env();
    crate::config::invalidate_config_cache();   // <-- added
    ...
}
```

This is a one-line, focused test-isolation fix with no production behaviour changes. An alternative would be to make `PinTodosEnvGuard` always invalidate on construction and use it in a "disabled" mode here too, but that adds a mode flag for a single call site; a direct `invalidate_config_cache()` matches the existing pattern most closely.

## Reference patch

I pushed the fix to my fork (branch `fix/pinned-todos-config-cache-test-isolation`, commit https://github.com/shaikahad-tech/jcode/commit/03d135a1dffa76fb7a869e7ad830ca1bf4adc77e) but could not open a PR directly because this repo's PR creation policy is set to `collaborators_only`. Happy to open a PR if that's adjusted, or the maintainer is welcome to use the commit as a reference.

## Reproduction

The failure is intermittent (depends on test execution order and timing), but reliably surfaces in the full `cargo test -p jcode-tui --lib -- --test-threads=1` run as seen in the linked CI run.

## Verification

CI run: https://github.com/1jehuang/jcode/actions/runs/31359160118/job/93364458815
Failing test: `tui::app::tests::pinned_todos_payload_stays_empty_when_config_off` at `crates/jcode-tui/src/tui/app/tests/todo_card.rs:227`

Note: I could not run the full Rust test suite locally before filing this, so the diagnosis is based on reading the code and the in-file cache-throttle comment. Treating this as a proposal/reference per CONTRIBUTING.md.

1jehuang:
## Summary
- explicitly force `JCODE_PIN_TODOS=0` in the config-off test
- invalidate the configuration cache before creating the test app
- reuse the existing guard cleanup so later tests cannot inherit stale state

## Verification
- `cargo test -p jcode-tui --lib tui::app::tests::pinned_todos_payload_stays_empty_when_config_off -- --exact --test-threads=1`

Fixes #877

---

### DIFF
--- a/crates/jcode-desktop2/src/scene.rs
+++ b/crates/jcode-desktop2/src/scene.rs
@@ -1311,12 +1311,7 @@ fn draw_transcript(
                     Affine::scale(scale),
                     color,
                     None,
-                    &Rect::new(
-                        x0,
-                        y0,
-                        x1,
-                        y1,
-                    ),
+                    &Rect::new(x0, y0, x1, y1),
                 );
             }
             if let Some(selection) = model.selection.as_ref()
@@ -1941,18 +1936,10 @@ mod tests {
         for scale in [1.0, 1.25, 1.5, 1.75, 2.0, 2.5] {
             let hairline = 1.0 / scale;
             let origin = 13.37;
-            let (_, first_bottom) = diff_band_y(
-                Rect::new(0.0, 0.0, 100.0, 19.2),
-                origin,
-                hairline,
-                false,
-            );
-            let (second_top, _) = diff_band_y(
-                Rect::new(0.0, 19.2, 100.0, 38.4),
-                origin,
-                hairline,
-                false,
-            );
+            let (_, first_bottom) =
+                diff_band_y(Rect::new(0.0, 0.0, 100.0, 19.2), origin, hairline, false);
+            let (second_top, _) =
+                diff_band_y(Rect::new(0.0, 19.2, 100.0, 38.4), origin, hairline, false);
             let overlap_px = (first_bottom - second_top) * scale;
             assert!(
                 (overlap_px - 1.0).abs() < 1e-9,
--- a/crates/jcode-desktop2/src/tests/actions.rs
+++ b/crates/jcode-desktop2/src/tests/actions.rs
@@ -1590,10 +1590,7 @@ fn a_new_session_preserves_the_old_panel_until_the_new_one_attaches() {
     let (updates_tx, updates_rx) = channel();
     let (commands_tx, _commands_rx) = channel();
     let commands = crate::harness::CommandSender::for_test(commands_tx);
-    app.harness = Some((
-        updates_rx,
-        commands.clone(),
-    ));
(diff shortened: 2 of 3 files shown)


1jehuang:
CI follow-up: the formatting gate now passes. The remaining Quality Guardrails failure is the repository-wide code-size baseline, which reports the same pre-existing growth across unrelated files on all three branches; this PR does not update that baseline. Platform build/test jobs are still queued.

---

PR #882, opened 2026-08-10
Avoid duplicate Codex quota windows

### LINKED ISSUE #869: Codex quota widget shows two weekly bars when the hourly window is absent
## Summary

After Codex stopped returning an hourly quota, Jcode renders the remaining two quota windows as two separate **weekly** bars. This is misleading because neither returned window is an hourly limit.

## Reproduction

1. Sign in with a Codex/OpenAI OAuth account whose current usage response has no hourly window and contains two weekly windows.
2. Start Jcode and inspect the OpenAI limits in the info widget or compact header.
3. Jcode shows two quota lines, both labeled with the backend's weekly window label.

## Expected

Render the reported windows without assigning an hourly role when the response does not contain an hourly window. In particular, do not display two weekly windows as the legacy `five_hour` and `seven_day` slots. The UI could use a generic ordered list of returned windows, or correctly distinguish the two weekly quotas when the backend supplies an identifier.

## Actual

The first weekly window is stored in the `seven_day` field because its name matches the weekly detector. The generic fallback then stores that *same first window* in `five_hour`, because no hourly window was found. The widget renders both field labels verbatim, so users see the same weekly quota twice.

## Root cause

`crates/jcode-base/src/usage/openai_helpers.rs` still models OpenAI usage as fixed `five_hour`, `seven_day`, and `spark` slots. In `classify_openai_limits`:

```rust
if five_hour.is_none() {
    five_hour = generic_non_spark.first().cloned();
}
if seven_day.is_none() {
    seven_day = generic_non_spark.iter().find(/* a different item */).cloned();
}
```

This fallback causes the first generic window to fill the historical hourly slot whenever no limit matches `limit_mentions_five_hour`. A weekly limit has already filled `seven_day`, so that same weekly quota is rendered through both legacy slots. A second returned weekly window is not meaningfully represented.

`crates/jcode-tui/src/tui/app/tui_state.rs` then forwards `window.name` as each display label, and `info_widget_usage.rs` renders both bars. Therefore the issue is not merely cosmetic label text: the model cannot represent the changed Codex quota shape.

## Validation

- Inspected the v0.75.0 source at commit `5ae2385`.
- Ran `cargo test -p jcode-base test_classify_openai_limits --lib --quiet` successfully (2 tests passed). Existing coverage only tests the old 5h + 1w shape and does not cover two weekly windows.

## Suggested fix

Replace the fixed two-window fallback for OpenAI with a list of normalized windows, or only populate the hourly slot when a returned limit explicitly identifies an hourly duration. Add a regression test with two `7-day window` limits and no hourly limit. The regression should assert that the first weekly window is not duplicated into an hourly slot and that both returned windows remain distinguishable.

1jehuang:
## Summary
- exclude an already-classified weekly quota from the generic fallback
- preserve both distinct weekly windows when no hourly window is returned
- add regression coverage for the new response shape

## Verification
- `cargo test -p jcode-base --lib test_classify_openai_limits`

Fixes #869

---


1jehuang:
CI follow-up: the formatting gate now passes. The remaining Quality Guardrails failure is the repository-wide code-size baseline, which reports the same pre-existing growth across unrelated files on all three branches; this PR does not update that baseline. Platform build/test jobs are still queued.

---

1jehuang:
Ubuntu follow-up: the suite reached 2,170 passing TUI tests and failed only `pinned_todos_payload_stays_empty_when_config_off`, the independently reproduced cache-isolation defect fixed by #880. The focused tests for this PR pass, as do macOS, Windows, formatting, cross-target, SDK, installer, PowerShell, release, linked-issue, and Greptile checks.

---

1jehuang:
CI dependency update: this PR is now stacked on #880 so Ubuntu validates this change with the independently verified pinned-todos isolation fix present. After #880 merges, the base can return to `master` without changing this PR’s issue-specific commit.

---

1jehuang:
Integration validation update: the base is temporarily set to `master` so the repository’s CI workflow runs against the head containing both this fix and prerequisite #880. After terminal results are captured, the PR will be restored to the stacked base so its review diff remains issue-specific.

---

PR #918, opened 2026-08-13
Fix seven reproducible issues from automated triage

### LINKED ISSUE #908: Custom OpenAI-compatible provider profiles are silently routed through the OpenRouter runtime, inheriting its orphaned-tool-output drop bug
## Summary
Any provider configured in config.toml as `type = "open-ai-compatible"` (or a named profile) — e.g. a self-hosted LM Studio/Ollama/vLLM/SGLang endpoint — is routed through `ActiveProvider::OpenRouter` (`crates/jcode-base/src/provider/selection.rs:29-31`), not the native OpenAI provider. This means it inherits the OpenRouter runtime's `build_chat_messages` logic, including a bug where orphaned tool-call outputs (a `ToolResult` whose matching `ToolUse` was pruned, typically by compaction) are silently dropped instead of recovered.

## Routing chain (verified)
- `crates/jcode-base/src/provider/selection.rs:29-31`: `OpenAiCompatibleProfile(_) | NamedProfile(_) => ActiveProvider::OpenRouter`.
- `crates/jcode-base/src/provider/openrouter.rs:2-9,73`: OpenRouter runtime is documented as the generic transport that "Direct OpenAI-compatible endpoints reuse".
- `src/cli/startup.rs:164-190`: `register_openrouter_factory` maps compatible/named profiles to `OpenRouterProvider`.
- `crates/jcode-provider-openrouter-runtime/src/openrouter_provider_impl.rs:97`: calls `jcode_provider_openrouter::request::build_chat_messages`.
- `crates/jcode-base/src/provider_catalog.rs:573-632`: writes the custom base_url into `JCODE_OPENROUTER_API_BASE`.
- Only the native built-in `openai-api` provider (Responses API, `/responses`) uses the OpenAI runtime's orphan-recovery path (`crates/jcode-provider-openai/src/request.rs:338-357`).

## The underlying bug
In `crates/jcode-provider-openrouter/src/request.rs:303-305`, leftover `pending_tool_results` with no matching `ToolUse` are folded into `skipped_results` and discarded (log: `[openrouter] Filtered {} orphaned tool result(s)...` at line 315) — the content is never re-encoded into any message. By contrast, `crates/jcode-provider-openai/src/request.rs:74-98,338-357` recovers the same situation as a user message: `"[Recovered orphaned tool output: {}]\n{}"`.

## Reproduction
A minimal test against `build_chat_messages` in `jcode-provider-openrouter`, feeding a single orphaned `ToolResult{tool_use_id: "call_orphan", content: "orphan result"}`, produces an empty message list — the content exists nowhere in the output request. The equivalent OpenAI-provider test recovers it correctly.

## Impact
Because compaction (`jcode-base/src/compaction.rs`) can prune an assistant `ToolUse` while leaving its paired `ToolResult`, any long session against a self-hosted/custom OpenAI-compatible endpoint can silently lose tool-result content mid-task. The agent then appears to stall or end its turn prematurely with no memory of the tool result it just requested, requiring the user to manually say "continue." This affects every `open-ai-compatible`/named-profile user, not just people who deliberately choose openrouter.ai.

## Suggested fix
In `crates/jcode-provider-openrouter/src/request.rs`, at line ~303, re-encode leftover `pending_tool_results` as `{"role":"user","content":"[Recovered orphaned tool output: ...]"}` messages (OpenAI chat-completions format) instead of discarding them into `skipped_results`, mirroring `orphan_tool_output_to_user_message()` from the OpenAI provider. Leave the separate duplicate-output drop at lines 507-513 unchanged, as that path is already correct.

## Checked for duplicates
Searched issues/PRs for 'orphaned tool output', 'orphan', 're-ordering', 'dropped tool output', 'compatible profile' — no existing report covers this exact routing + drop combination.

### LINKED ISSUE #910: Enhancement: one-line clipped descriptions in the "Available Skills" system-prompt section
## Summary

The "Available Skills" section of the system prompt lists every loaded skill's full `description` verbatim. Long descriptions (200–400 chars, often multi-line with `[[wiki]]`-style references) make the section heavy, and since the section is part of the cache-stable prompt prefix, its size costs tokens on **every** turn regardless of whether the skill is ever invoked. Skills are invoked by `/name` — the full description is rarely needed up front.

## Gap

`build_system_prompt_full_with_capabilities` and `build_system_prompt_split_with_capabilities` (`crates/jcode-base/src/prompt.rs`) append `- \`/name\` — <full description>` with no clipping or whitespace normalization; a description containing newlines breaks the one-line-per-skill format.

## Proposed change (implemented and verified locally)

- `clip_skill_description`: collapse all whitespace to single spaces (one line per entry), clip to `SKILL_DESC_MAX_CHARS = 120` chars on a char boundary with a `…` suffix when truncated.
- Applied in both section builders; full bodies still load on demand via `/skillname`.

## Impact (measured, tiktoken o200k_base)

23 skills, host runtime set: Available Skills section **1 243 → 668 tokens (x1.86)**; descriptions part 5 179 → ~2.4 KB.

## Verification

- 4 unit tests (`skills_clip_tests`): long-description clipping, short-description whitespace collapse, Cyrillic char-boundary, section-builder integration.
- `cargo check -p jcode-base` + ACP smoke pass on the built image.

## Files

`crates/jcode-base/src/prompt.rs` (`clip_skill_description`, both section builders).

### LINKED ISSUE #911: Bug: AGENTS.md loaded twice when the working directory is $HOME
## Summary

`load_agents_md_files_from_dir` reads AGENTS.md from the **working directory** and from the **home directory** as two separate sources ("Project Instructions" + "Global Instructions"). When the working directory IS the home directory (very common for ACP hosts spawning `jcode` with `cwd=$HOME` — e.g. multica tasks whose workdir is the agent's home), the same file is loaded twice and both sections appear in the system prompt: duplicated instructions, duplicated tokens, and conflicting "which wins?" guidance for the model.

## Gap

No dedup between the two sources. `crates/jcode-base/src/prompt.rs` `load_agents_md_files_from_dir` unconditionally loads both.

## Fix (implemented and verified locally)

- Core logic extracted into `load_agents_md_files_from_dirs(project_dir, home_agents_path)` (testable, no I/O in the decision): if `canonicalize(project_dir) == canonicalize(home_dir)`, the home copy is skipped — the project source wins and only one "Project Instructions" section is emitted; the global section is not fabricated.
- Distinct project/home dirs keep the current behavior (both sections).

## Impact (measured)

With `cwd=$HOME`: ~2.9 KB of duplicated instructions removed from the system prompt (≈ 750 tokens, o200k_base).

## Verification

- 3 unit tests (`agents_md_dedup_tests`): project==home → loaded once (no "Global Instructions"), distinct dirs → both, missing home file → project only.
- `cargo check -p jcode-base` + ACP smoke pass.

## Files

`crates/jcode-base/src/prompt.rs` (`load_agents_md_files_from_dirs`, `load_agents_md_files_from_dir` shim).

### LINKED ISSUE #913: Idle animation blanks the rest of the screen: seeded partial repaint survives a swap_buffers reset
### Summary

On an idle screen the TUI blanks everything except the animation rows: transcript, header and composer disappear. Typing brings them back, so the reported symptom is **"the TUI blinks: I can see it while I write, it's gone when I stop"**.

Still present on `master` at v0.75.5 (`crates/jcode-tui/src/tui/app/run_shell.rs`, the `seeded_animation_area` path at ~line 402, and the `swap_buffers` site at ~line 429).

### Root cause

`draw_idle_animation_only` ends in `terminal.swap_buffers()`. ratatui's `swap_buffers` **resets the now-inactive buffer** (`Buffer::reset` blanks every cell) before flipping, and that inactive buffer is the one the *next* tick draws into.

The seeding optimization assumes the opposite. Once `seeded_animation_area == Some(area)` it copies only the animated rectangle and trusts every cell outside it:

```rust
if self.seeded_animation_area == Some(area) {
    copy_cells_in(previous_frame, current_buffer, area);
} else {
    current_buffer.clone_from(previous_frame);
    self.seeded_animation_area = Some(area);
}
```

Against a freshly reset buffer that trust is misplaced for the whole rest of the screen. The flushed diff therefore erases the transcript, header and composer and leaves the animation alone on screen. Typing repaints it only because a changed composer trips the `input_changed` guard and forces a full frame.

`invalidate()` clears the seed, but nothing clears it at the swap, which is the moment the seed stops being valid.

### Why the existing tests missed it

Every test in that module compares one rendered buffer against another. None model what ratatui does to its *own* buffers between frames, which is where the bug lives. It is a buffer-lifecycle mismatch, not a rendering error.

### Fix

Clear the seed at the swap, so a seed is only ever reused within a tick, never across one:

```rust
/// Record that a frame was handed to the terminal and the buffers swapped.
///
/// `Terminal::swap_buffers` resets the now-inactive buffer, which is the
/// one the next tick draws into, so no part of the previous frame survives
/// there. Any partial-repaint path that skips cells because they are
/// "already correct" must call this after swapping, or it will skip cells
/// that are in fact blank and erase the screen around its own rectangle.
fn note_animation_frame_presented(&mut self) {
    self.seeded_animation_area = None;
}
```

called immediately after `terminal.swap_buffers()` in `draw_idle_animation_only`.

I have this running locally as a single commit against v0.75.5 (+137/-6 in `run_shell.rs`, all but ~10 lines being comments and two regression tests). I do not have push access to open the PR; happy to attach the patch or have someone apply it.

The two tests it adds:

1. `the_idle_animation_repaint_does_not_blank_the_rest_of_the_screen` drives the two real production helpers through the real buffer lifecycle (seed, swap/reset, copy) for three consecutive ticks and asserts the cells outside the animated rectangle survive. It takes the seeding branch from the renderer's own state rather than assuming an answer, so the fix is what makes it pass.
2. `a_presented_animation_frame_requires_reseeding_the_next_tick` pins the invariant on the renderer itself, so a future edit cannot reinstate the optimization without keeping the seed valid across a swap.

### Environment

Windows 11 x86_64, jcode built from source. Originally hit on v0.65.0, verified by inspection that `master` (994b8d3dd, v0.75.5) still contains the unguarded path.

Related, though distinct: #289 "Flashing violently after auto updating" (closed).

### LINKED ISSUE #914: self-dev --build is broken on Windows: native path passed to bash -lc
### Summary

On Windows, `jcode self-dev --build` always fails immediately:

```
$ jcode self-dev --build
Building with scripts/dev_cargo.sh build --profile selfdev -p jcode --bin jcode...
/bin/bash: line 1: C:\Users\yw39i\.jcode\source\jcode\scripts\dev_cargo.sh: command not found
Error: Build failed: scripts/dev_cargo.sh build --profile selfdev -p jcode --bin jcode
```

The build command is assembled in `crates/jcode-build-support/src/paths.rs:196` (`selfdev_build_command_for_target`) and then run through `bash -lc`:

```rust
let wrapper = repo_dir.join("scripts").join("dev_cargo.sh");
if wrapper.is_file() {
    let script = wrapper.to_string_lossy();          // Windows: C:\Users\...\scripts\dev_cargo.sh
    ...
    format!("{} build --profile {} -p {} --bin {}{}", shell_escape(&script), ...)
```

`wrapper` is a native `PathBuf`, so on Windows `to_string_lossy()` yields a backslash path with a drive letter. `shell_escape` quotes it but does not translate it, and bash then treats `C:\Users\...` as a command name, where the backslashes are also escape characters. The same string works when converted:

```bash
$ bash -lc '/c/Users/yw39i/.jcode/source/jcode/scripts/dev_cargo.sh --help'
dev_cargo: sccache skipped for incremental build (...)     # runs
```

So the script itself is fine, only the path form is wrong.

Note `wrapper.is_file()` **succeeds** on Windows (it is a real file), which is what selects this branch. The fallback branch below it invokes `cargo` directly and would have worked.

### Impact

`self-dev --build` is the documented, supported way to publish a locally built binary to the `current` / `stable` / `shared-server` channels. With it broken on Windows there is no first-class path to publish a self-dev build:

- `selfdev reload` from a non-self-dev session deliberately does not publish (`tool/selfdev/setup.rs:281`, "this does not publish a freshly-built binary"), and answers "Already running the newest installed jcode build" because the fresh binary was never installed.
- the publish-and-reload path in `server/debug_command_exec.rs:608` requires the debug socket.

So a Windows self-dev user builds successfully, then cannot install what they built without hand-copying into `AppData/Local/jcode/builds/versions/<label>/` and recreating four symlinks by hand, which is exactly what `install_local_release` exists to prevent.

### Suggested fix

Convert to a POSIX path before handing it to bash, on Windows only. Something like:

```rust
let script = to_bash_path(&wrapper);   // C:\a\b -> /c/a/b
```

Alternatively, prefer the `cargo` fallback branch when `cfg!(windows)`, since `program: "bash"` is already questionable there. The invocation also assumes a login shell that has `cargo` on `PATH`; under Git Bash on Windows `cargo` was not found even when the script ran, so the same code path likely needs `~/.cargo/bin` handling as well.

### Environment

Windows 11 x86_64, Git Bash (`C:\Program Files\Git\bin\bash.exe`), jcode built from source at v0.75.5 + 1 local commit, rustup 1.29.0 / rustc 1.97.1 with `cargo` at `%USERPROFILE%\.cargo\bin` but not on the system `PATH`.

### LINKED ISSUE #916: tui: expand paste placeholder when the same text is pasted again
## Problem

When pasting 5+ lines of text into the input line, jcode collapses the paste into a `[pasted N lines]` placeholder (existing, intended behavior — avoids flooding the input line with a huge block of text).

Once collapsed, there's no way to review or edit that content from the input line. The only way to get the text back is to clear the input and re-paste from your clipboard/paste buffer, which is awkward mid-message (e.g. you collapsed a big block, then want to tweak one line of it before sending).

Other paste-collapsing composers solve this with a "paste to collapse, paste the same content again to expand" gesture: paste once and it collapses to a placeholder, paste that exact same content again while the placeholder is still there and it expands back into the full text in place.

## Proposed fix

I implemented this for jcode:

- Pasting text with 5+ lines still collapses into `[pasted N lines]` and stores the content in `pasted_contents`, as today.
- If you paste that **exact same text** again while its placeholder is still present, unedited, in the input, the placeholder is replaced with the full text in place (cursor moves to the end of the expanded text), and the entry is removed from `pasted_contents`.
- Pasting **different** text still collapses into a new placeholder, as before — no accidental expansion of an unrelated placeholder.

## Reproduction / test case

Added as unit tests in `crates/jcode-tui/src/tui/app/tests/remote_startup_input_03/part_01.rs`:

```rust
#[test]
fn test_paste_again_expands_the_placeholder_in_place() {
    let big = "a\nb\nc\nd\ne".to_string();

    // First paste collapses to a placeholder and remembers the content.
    app.handle_paste(big.clone());
    assert_eq!(app.input(), "[pasted 5 lines]");
    assert_eq!(app.pasted_contents.len(), 1);

    // Pasting the exact same text again expands the existing placeholder
    // instead of appending a second one.
    app.handle_paste(big.clone());
    assert_eq!(app.input(), big);
    assert!(app.pasted_contents.is_empty());
    assert_eq!(app.cursor_pos, big.len());
}

#[test]
fn test_paste_again_with_different_text_still_collapses() {
    app.handle_paste("a\nb\nc\nd\ne".to_string());
    assert_eq!(app.input(), "[pasted 5 lines]");

    // A different large paste does not match the stored content, so it
    // collapses into its own placeholder rather than expanding the first one.
    app.handle_paste("f\ng\nh\ni\nj".to_string());
    assert_eq!(app.input(), "[pasted 5 lines][pasted 5 lines]");
    assert_eq!(app.pasted_contents.len(), 2);
}
```

Ran locally on Windows:

```
cargo test -p jcode-tui paste_again
running 2 tests
test tui::app::tests::test_paste_again_with_different_text_still_collapses ... ok
test tui::app::tests::test_paste_again_expands_the_placeholder_in_place ... ok
test result: ok. 2 passed; 0 failed
```

Full existing `handle_text_paste` test suite (`test_handle_paste_large`, `test_paste_expansion_on_submit`, `test_multiple_pastes`, etc.) also passes unchanged.

## Implementation reference

I have a small, focused diff (`crates/jcode-tui/src/tui/app/input.rs` + the test file above, 57 insertions / 8 deletions) on my fork, in case it's useful as a reference per CONTRIBUTING.md:

https://github.com/plakunov/jcode/compare/master...plakunov:jcode:feature/paste-again-to-expand

(Repository interaction settings currently limit opening PRs to collaborators, so I'm filing this as an issue with the branch linked instead, per the "issues vs pull requests" guidance in CONTRIBUTING.md.)

## Edge cases / tradeoffs considered

- Match is on exact text equality against the most recently pushed matching entry in `pasted_contents` (`rposition`), and requires the placeholder to still be found verbatim in the current input (`rfind`) — so if the user edited the input around the placeholder text itself, or the placeholder is no longer present, it just collapses normally instead of expanding.
- Only the closest/most recent matching stored content is removed from `pasted_contents`, not all occurrences, to avoid clobbering an unrelated earlier paste of identical content.
- Didn't touch the &lt;5-line short-paste path, which inserts text directly and was already correct.

### LINKED ISSUE #917: A locally built binary can never reach the shared server: no caller for promote_version_to_shared_server
### Summary

A locally built binary can reach the client but never the daemon, and nothing in the tree can move it there.

The daemon resolves its reload and spawn target from the `shared-server` channel:

- `spawn_server` (`src/cli/dispatch.rs:1296`) -> `build::shared_server_update_candidate(...)`
- `reload_exec_target` (`crates/jcode-app-core/src/server/util.rs:102`) -> `newest_reload_candidate` -> same

`shared_server_update_candidate` (`crates/jcode-build-support/src/paths.rs:541`) tries `shared-server`, then `stable`, then `current_exe()`. It never consults `current`.

But `publish_local_current_build_for_source` (`crates/jcode-build-support/src/lib.rs:679`), which is what `self-dev --build` calls, only moves `current` and the launcher:

```rust
let current_link = update_current_symlink(&source.version_label)?;
let launcher_link = update_launcher_symlink_to_current()?;
```

And the two functions that *could* advance the daemon's channel both decline in this case:

- `advance_shared_server_if_tracking_stable` is only called from the update path.
- `repair_stale_shared_server_channel` drags `shared-server` to **stable**, and returns `AlreadyCurrent` when the existing marker is not a release marker.

So after `self-dev --build` the client runs the new binary and the daemon keeps the old one indefinitely.

### What that looks like

On this machine the client was on v0.75.8-dev while the server had been serving v0.65.0 since Aug 4, ten releases apart, over the same socket. Both of these are correct-but-useless:

```
$ jcode server reload --json
{"reloaded": false, "already_current": true,
 "detail": "jcode server is already running the newest binary; no reload needed."}

$ jcode server reload --force      # re-execs the SAME old binary
$ jcode selfdev status
Running: jcode v0.65.0 (d847d62e4)
Current: ee3b3f4a5
Shared server: 0.65.0
```

`reload` is truthful: relative to its own candidate (the stale channel) the daemon *is* current. `--force` bypasses `server_has_newer_binary` but still resolves the same candidate, so it cannot help either.

### There is already an API for this, with no caller

```rust
/// Promote an already installed immutable version onto the shared server channel.
pub fn promote_version_to_shared_server(version: &str) -> Result<Option<String>> {
```

`crates/jcode-build-support/src/lib.rs:726`, re-exported through `jcode-app-core/src/build.rs:14`. `grep -rn 'promote_version_to_shared_server('` matches only the definition and the re-export.

### Suggested fix

Wire it to a command. I have this working locally as `jcode server promote [VERSION] [--json]`, defaulting to whatever `current` points at:

```
$ jcode server promote --json
{
  "version": "ee3b3f4a5",
  "previous": "0.65.0",
  "binary": ".../builds/versions/ee3b3f4a5/jcode.exe",
  "promoted": true,
  "detail": "shared-server channel 0.65.0 -> ee3b3f4a5. Run `jcode server reload` to apply it ..."
}

$ jcode server reload --force
$ jcode selfdev status
Running: jcode v0.75.8-dev (ee3b3f4a5)
```

Kept as a separate step from `reload` rather than folded into it, since the pin exists precisely so an update never silently wipes out a self-dev build; promoting is the decision that pin protects. It reports `previous` so the move is reversible, refuses a version that is not installed (a dangling channel symlink would send the next server spawn down a silent fallback), and is idempotent.

An alternative worth considering: have `self-dev --build` offer to promote, or teach `server reload` to report *why* it considers itself current (naming the channel and version it resolved), so the state is at least discoverable. The current message actively misleads.

### Environment

Windows 11 x86_64, jcode built from source at v0.75.5 + local commits. Same machine as #912 and #914; this is the third link in the same chain: the updater would not update (#912), self-dev could not build (#914), and even after building, the daemon could not be moved (this one).

1jehuang:
## Summary
- add explicit shared-server build promotion and native Windows self-dev Cargo execution
- expand repeated paste placeholders and repair idle-animation buffer reseeding
- deduplicate home/project `AGENTS.md`, clip skill descriptions safely, and preserve orphaned OpenRouter tool outputs

Fixes #908
Fixes #910
Fixes #911
Fixes #913
Fixes #914
Fixes #916
Fixes #917

## Verification
- `cargo test -p jcode-build-support`
- `cargo test -p jcode-base prompt`
- `cargo test -p jcode-provider-openrouter`
- focused paste and idle-animation regression suites
- `cargo test -p jcode --lib server_promote_parses_default_and_explicit_version`
- `cargo check -p jcode-build-support --target x86_64-pc-windows-gnu`
- `git diff --check origin/master..HEAD`

### DIFF
--- a/crates/jcode-base/src/prompt.rs
+++ b/crates/jcode-base/src/prompt.rs
@@ -231,6 +231,41 @@ pub struct SkillInfo {
     pub description: String,
 }
 
+const SKILL_DESC_MAX_CHARS: usize = 120;
+
+fn clip_skill_description(description: &str) -> String {
+    let one_line = description.split_whitespace().collect::<Vec<_>>().join(" ");
+    if one_line.chars().count() <= SKILL_DESC_MAX_CHARS {
+        return one_line;
+    }
+
+    let mut clipped: String = one_line
+        .chars()
+        .take(SKILL_DESC_MAX_CHARS.saturating_sub(1))
+        .collect();
+    clipped.push('…');
+    clipped
+}
+
+fn build_available_skills_section(available_skills: &[SkillInfo]) -> Option<String> {
+    if available_skills.is_empty() {
+        return None;
+    }
+
+    let mut section = "# Available Skills\n\nYou have access to the following skills that the user can invoke with `/skillname`:\n".to_string();
+    for skill in available_skills {
+        section.push_str(&format!(
+            "\n- `/{} ` - {}",
+            skill.name,
+            clip_skill_description(&skill.description)
+        ));
+    }
+    section.push_str(
+        "\n\nWhen a user asks about available skills or capabilities, mention these skills.",
+    );
+    Some(section)
+}
+
 /// Information about what's loaded in the context window
 #[derive(Debug, Clone, Default)]
 pub struct ContextInfo {
@@ -453,14 +488,7 @@ pub fn build_system_prompt_full_with_capabilities(
     }
 
     // Add available skills list
-    if !available_skills.is_empty() {
-        let mut skills_section = "# Available Skills\n\nYou have access to the following skills that the user can invoke with `/skillname`:\n".to_string();
-        for skill in available_skills {
-            skills_section.push_str(&format!("\n- `/{} ` - {}", skill.name, skill.description));
-        }
-        skills_section.push_str(
-            "\n\nWhen a user asks about available skills or capabilities, mention these skills.",
-        );
+    if let Some(skills_section) = build_available_skills_section(available_skills) {
         info.skills_chars = skills_section.len();
         parts.push(skills_section);
     }
@@ -546,14 +574,7 @@ pub fn build_system_prompt_split_with_capabilities(
     }
 
     // Add available skills list (fairly static)
-    if !available_skills.is_empty() {
-        let mut skills_section = "# Available Skills\n\nYou have access to the following skills that the user can invoke with `/skillname`:\n".to_string();
-        for skill in available_skills {
-            skills_section.push_str(&format!("\n- `/{} ` - {}", skill.name, skill.description));
-        }
-        skills_section.push_str(
-            "\n\nWhen a user asks about available skills or capabilities, mention these skills.",
-        );
+    if let Some(skills_section) = build_available_skills_section(available_skills) {
         info.skills_chars = skills_section.len();
         static_parts.push(skills_section);
     }
@@ -840,8 +861,10 @@ fn gpu_summary() -> Option<String> {
     }
 }
 
-/// Load AGENTS.md files from a specific working directory
-pub fn load_agents_md_files_from_dir(working_dir: Option<&Path>) -> (Option<String>, ContextInfo) {
+fn load_agents_md_files_from_dirs(
+    project_dir: &Path,
+    global_agents_md: Option<&Path>,
+) -> (Option<String>, ContextInfo) {
     let mut contents = vec![];
     let mut info = ContextInfo::default();
 
@@ -858,21 +881,31 @@ pub fn load_agents_md_files_from_dir(working_dir: Option<&Path>) -> (Option<Stri
         }
     };
 
-    // Project-level files (from specified working directory or current directory)
-    let project_dir = working_dir.unwrap_or(Path::new("."));
-    if let Some((content, size)) = load_file(
-        &project_dir.join("AGENTS.md"),
-        "Project Instructions (AGENTS.md)",
-    ) {
+    let project_agents_md = project_dir.join("AGENTS.md");
+    if let Some((content, size)) = load_file(&project_agents_md, "Project Instructions (AGENTS.md)")
+    {
         info.has_project_agents_md = true;
         info.project_agents_md_chars = size;
         contents.push(content);
     }
 
-    // Home directory files
-    if let Ok(global_agents_md) = crate::storage::user_home_path("AGENTS.md")
+    // Canonical file identity handles cwd=$HOME as well as symlinked aliases.
+    // If either file is absent or cannot be resolved, loading below remains the
+    // source of truth and simply skips unreadable files.
+    let global_duplicates_project = global_agents_md.is_some_and(|global_agents_md| {
+        match (
+            std::fs::canonicalize(&project_agents_md),
+            std::fs::canonicalize(global_agents_md),
+        ) {
+            (Ok(project), Ok(global)) => project == global,
+            _ => false,
+        }
+    });
+
+    if !global_duplicates_project
+        && let Some(global_agents_md) = global_agents_md
         && let Some((content, size)) =
-            load_file(&global_agents_md, "Global Instructions (~/AGENTS.md)")
+            load_file(global_agents_md, "Global Instructions (~/AGENTS.md)")
     {
         info.has_global_agents_md = true;
         info.global_agents_md_chars = size;
@@ -884,6 +917,13 @@ pub fn load_agents_md_files_from_dir(working_dir: Option<&Path>) -> (Option<Stri
     } else {
         (Some(contents.join("\n\n")), info)
     }
+}
+
+/// Load AGENTS.md files from a specific working directory.
+pub fn load_agents_md_files_from_dir(working_dir: Option<&Path>) -> (Option<String>, ContextInfo) {
+    let project_dir = working_dir.unwrap_or(Path::new("."));
+    let global_agents_md = crate::storage::user_home_path("AGENTS.md").ok();
+    load_agents_md_files_from_dirs(project_dir, global_agents_md.as_deref())
 }
 
 /// Load optional prompt overlay markdown from ~/.jcode/ and ./.jcode/
--- a/crates/jcode-base/src/prompt_tests.rs
+++ b/crates/jcode-base/src/prompt_tests.rs
@@ -64,6 +64,61 @@ fn test_skill_prompt_integration() {
 }
 
 #[test]
+fn skill_description_collapses_whitespace_without_clipping_short_text() {
+    assert_eq!(
+        clip_skill_description("  Build\n\tand   test the project.  "),
+        "Build and test the project."
+    );
+}
+
+#[test]
+fn skill_description_clips_to_the_character_limit_with_ellipsis() {
+    let clipped = clip_skill_description(&"a".repeat(SKILL_DESC_MAX_CHARS + 20));
+
+    assert_eq!(clipped.chars().count(), SKILL_DESC_MAX_CHARS);
+    assert!(clipped.ends_with('…'));
+}
+
+#[test]
+fn skill_description_clipping_is_utf8_safe() {
+    let clipped = clip_skill_description(&"ж".repeat(SKILL_DESC_MAX_CHARS + 20));
+
+    assert_eq!(clipped.chars().count(), SKILL_DESC_MAX_CHARS);
+    assert_eq!(
+        clipped
+            .chars()
+            .filter(|character| *character == 'ж')
+            .count(),
+        SKILL_DESC_MAX_CHARS - 1
+    );
+    assert!(clipped.ends_with('…'));
+}
+
+#[test]
+fn full_and_split_prompt_builders_use_the_same_one_line_skill_descriptions() {
+    let skills = vec![SkillInfo {
+        name: "example".to_string(),
+        description: format!("First line\n\t{}", "д".repeat(SKILL_DESC_MAX_CHARS + 20)),
+    }];
+    let expected = build_available_skills_section(&skills).expect("skills section");
+
+    let (full, full_info) = build_system_prompt_full(None, &skills, false, None, None);
+    let (split, split_info) = build_system_prompt_split(None, &skills, false, None, None);
+
+    assert!(full.contains(&expected));
+    assert!(split.static_part.contains(&expected));
+    assert_eq!(full_info.skills_chars, expected.len());
+    assert_eq!(split_info.skills_chars, expected.len());
+    assert!(expected.contains("First line "));
+    assert!(!expected.contains("First line\n"));
+    let entry = expected
+        .lines()
+        .find(|line| line.starts_with("- `/example `"))
+        .expect("example skill entry");
+    assert!(entry.ends_with('…'));
+}
+
+#[test]
 fn test_load_agents_md_files_uses_sandboxed_global_files() {
     let _guard = crate::storage::lock_test_env();
     let prev_home = std::env::var_os("JCODE_HOME");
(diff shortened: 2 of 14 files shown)


pradfer123:
May I fix

1jehuang:
Closing as superseded by the equivalent fixes already landed on master. I also re-reviewed the remaining paste-repeat delta: Greptile reproduced a case where deleting the generated placeholder and typing the same literal still lets repeat-paste overwrite user-authored text, so this branch is not safe to merge as-is. The focused follow-up should track placeholder identity rather than infer it from rendered text.

pradfer123:
Ok

Issue #1073, opened 2026-08-26
Effort chip disappears after GetModelCatalog: partial History event clears remote_reasoning_effort


yumhum:
## Summary

The reasoning-effort chip next to the model name disappears on session open. It comes back only after switching the model.

Version: v0.81.1 (cae6d2a57), Linux x86_64, Anthropic provider, `anthropic_reasoning_effort = "medium"` in config.

## Repro

1. Open a new session, or resume an existing one.
2. The model chip shows `Opus 5` with no effort suffix.
3. Switch to a saved favourite model.
4. The effort appears and stays.

## Cause

The TUI renders the chip from `app.remote_reasoning_effort` only:

`crates/jcode-tui/src/tui/app/tui_state.rs:1307-1330`

That field is set from the `History` server event:

`crates/jcode-tui/src/tui/app/remote/server_events.rs:1701`

```rust
app.remote_reasoning_effort = reasoning_effort;
```

The assignment is unconditional. The three fields directly above it are guarded:

```rust
if session_changed || upstream_provider.is_some() { ... }
if session_changed || connection_type.is_some() { ... }
if session_changed || status_detail.is_some() { ... }
```

`handle_get_model_catalog` also emits a `History` event, and it hardcodes the field to `None`:

`crates/jcode-app-core/src/server/client_state.rs:254`

```rust
reasoning_effort: None,
```

So this catalog-only, partial `History` overwrites a good value with `None`, and the chip goes blank.

Recovery comes from `ReasoningEffortChanged`, which sets the field again:

`crates/jcode-tui/src/tui/app/remote/server_events.rs:2332`

That is why toggling a model brings the effort back.

## Suggested fix

Either:

- Guard the assignment like its neighbours: `if session_changed || reasoning_effort.is_some() { ... }`, or
- Populate the real effort in the catalog-only `History` payload, the way `send_history` does at `client_state.rs:643`.

The first option also protects any other partial `History` sender.

## Related

Closed issue #675 covered effort not persisting for Anthropic models. This is the same visible symptom from a different path.

1jehuang:
A focused fix with regression coverage is now in #1082. The integrated branch passes formatting, the relevant provider and Mermaid test suites, focused auth tests, and `cargo check -p jcode-tui`. This issue will close automatically when the PR merges.

--- *— Jcode agent (automated triage), on behalf of @1jehuang*

Issue #1072, opened 2026-08-26
Grok Build device login fails on Windows when HOME is unset: save_tokens never falls back to USERPROFILE


LamberlainMuli:
## Environment

1jehuang:
A focused fix with regression coverage is now in #1082. The integrated branch passes formatting, the relevant provider and Mermaid test suites, focused auth tests, and `cargo check -p jcode-tui`. This issue will close automatically when the PR merges.

--- *— Jcode agent (automated triage), on behalf of @1jehuang*

Issue #1071, opened 2026-08-26
RunningTool status skips the spinner fast path and forces a full-frame flush every tick


vsaakiancl:
## Summary

While a tool is executing (`ProcessingStatus::RunningTool`), every ~1s TUI tick falls through to the expensive full-frame `draw_full` redraw path instead of the existing single-cell spinner fast path that `Sending`/`Connecting`/`Thinking`/`Streaming` already use. On a normal terminal this is just wasted work; piped through a slow/embedded terminal consumer (in my case, a terminal pane hosted inside an Electron app rather than a native terminal), the backend flush for each of these redundant full-frame diffs took up to 3+ seconds, sustained for the full duration of a single long-running tool call, and drove system load to 32.8 on a 14-core Mac.

## Environment

- jcode v0.80.1 (d37ffd3dc), macOS (Darwin), 14 cores
- Terminal: jcode TUI running inside a terminal pane hosted by [Superset](https://superset.sh) (an Electron app) rather than a native terminal emulator
- Session status at the time: `RunningTool("bash")`, one tool call with `elapsed_ms=300248` (~5 minutes)

## Evidence (jcode's own `TUI_DRAW_CALL`/`TUI_SLOW_FRAME` telemetry, `~/.jcode/logs/jcode-*.log`)

For the ~5 minutes the bash tool was running — `is_processing:true`, `status:"RunningTool(\"bash\")"`, `streaming_text_len:0`, `display_messages` constant, i.e. nothing user-visible was actually changing — I still see repeated full-frame `TUI_DRAW_CALL` entries on plain `"input":{"event":"tick",...}`:

```
total_ms:1441.79  backend_flush_ms:1436.94  changed_cells:2141/4680
total_ms:1510.73  backend_flush_ms:1495.08  changed_cells:2124/4680
total_ms:1862.88  backend_flush_ms:1416.29  changed_cells:2632/4680  (after focus regain)
total_ms:2932.12  render_ms:2223.91                                  (resize)
total_ms:3043.27  backend_flush_ms:2728.98  changed_cells:3908/7680  (focus_gained)
```

Roughly half the buffer's cells are marked "changed" on plain `tick` events with no new content, and `backend_flush_ms` dominates `total_ms` in every case — this is write-side cost (flushing the diff to the terminal), not render computation.

Shortly after, when jcode restarted, its own perf-tier detection logged:

```
[...]

Issue #1068, opened 2026-08-26
Model picker hides Anthropic API-key route after switch and appears to remain on OAuth


wwillchen:
## Summary

In a remote/TUI session, selecting an Anthropic **API key** route in `/model` appears to leave the session on Claude OAuth. The picker sends and persists the correct API-key route, but the post-switch UI, server event, and environment snapshot collapse both Anthropic credential modes to the generic provider name `Claude`. This makes a successful API-key switch look like it stayed on OAuth.

## Environment

- jcode `v0.80.1 (d37ffd3dc)`
- macOS aarch64
- Both Anthropic auth modes configured: `claude` OAuth and `anthropic-api` direct API key
- Remote TUI session (`remote=true` in picker logs)

## Reproduction

1. Start a session currently using another provider or Claude OAuth.
2. Open `/model`.
3. Choose an Anthropic model route whose API method is `claude-api`.
4. Observe the provider/status after the switch.

## Observed evidence

The picker correctly identifies the direct API route:

```text
model_picker_select api_method=claude-api provider=Anthropic remote=true route_api_method=claude-api route_model=claude-fable-5 spec=claude-api:claude-fable-5
```

The server also receives the structured direct-API route:

```text
server_set_route_request current_provider=OpenAI requested_api_method=claude-api requested_model=claude-fable-5 requested_provider=Anthropic
```

The session snapshot correctly persists the route identity:

```json
{"provider_key":"anthropic-api-key","model":"claude-fable-5","route_api_method":"claude-api"}
```

However, the immediately following state only reports the generic provider:

```text
ENV_SNAPSHOT ... "provider":"Claude","model":"claude-fable-5" ...
server_model_changed model=claude-fable-5 provider=Claude
```

The same ambiguity appears in noninteractive output. An explicit direct-API smoke test:

```sh
jcode run --provider anthropic-api --json "Respond with exactly: ANTHROPIC_JCODE_OK"
```

returns `"provider": "Claude"` with `"text": "ANTHROPIC_JCODE_OK"`.

`jcode auth-test --provider anthropic-api` passes both provider and tool smoke tests, so the direct API credential itself works.

## Root cause

This appears to be an **active-route presentation/state propagation bug**, not a failure to apply the API-key credential:

1. `RouteSelection::routed_model_spec()` correctly maps `RuntimeKey::AnthropicApiKey` to `claude-api:<model>` in `crates/jcode-provider-core/src/lib.rs`.
[...]

PR #1082, opened 2026-08-27
fix: resolve safe issues from open-issue triage

### LINKED ISSUE #1090: Daemon idle-exit kills live headless swarm workers 5 minutes after the coordinator detaches
## Summary

The shared-server idle monitor exits the daemon after 5 minutes at zero connected clients. `client_count` only tracks **socket** clients. Headless and inline swarm workers run **in-process** and never open a client connection of their own, so they contribute **zero** occupancy.

A coordinator that detaches, or whose TUI is closed (SIGHUP), therefore leaves the daemon with live workers and no clients. Five minutes later the idle monitor calls `std::process::exit(EXIT_IDLE_TIMEOUT)` and every worker dies mid-turn, with no completion report and no visible error.

Because the survivors are then persisted as `crashed` / `stopped`, this presents as "swarm workers keep dying at random" rather than as a daemon lifecycle bug. The 5-minute delay between the terminal closing and the workers dying is what makes it look uncorrelated with anything.

## Code

`crates/jcode-app-core/src/server.rs`, idle monitor:

```rust
let count = *idle_client_count.read().await;
if count == 0 {
    // ... after IDLE_TIMEOUT_SECS
    std::process::exit(EXIT_IDLE_TIMEOUT);
}
```

`client_count` is incremented per client connection. `create_headless_session` registers a `SwarmMember` with `is_headless: true` and an internal `event_tx` whose receiver is simply drained; it never becomes a client connection. So live in-process work is invisible to this check.

## Evidence

From one machine's daemon log for 2026-08-28: **11 idle exits, 5 of which fired while the daemon was actively streaming provider responses and running tools** (counting `API call starting` / `Tool starting` in the final 60 seconds before exit):

```
2026-08-28 00:27:24: 3 live actions in last 60s
2026-08-28 00:47:34: 5 live actions in last 60s
2026-08-28 00:52:46: 10 live actions in last 60s
2026-08-28 01:11:36: 16 live actions in last 60s
2026-08-28 02:04:33: 38 live actions in last 60s
```

The 02:04:33 exit landed in the **same second** a worker began a `bash` tool call:

```
[02:04:33.386] Tool starting: bash
[02:04:33.386] EVENT event=TOOL_LIFECYCLE cwd=.../issue-321 execution_mode=AgentTurn
[02:04:33.756] Server idle for 5 minutes with no clients. Shutting down.
```

Across three days of logs, `Received SIGHUP in TUI process` is repeatedly followed by an idle exit at a gap of 5:00-5:20, exactly `IDLE_TIMEOUT_SECS`.

## Secondary symptom explained

This also accounts for `Session ... cleanup timed out waiting for agent lock (stuck task); skipping graceful shutdown`. The task is not stuck; the process is being torn down underneath it. Reading that warning as a hung worker sends debugging in the wrong direction.

## Expected behavior

A daemon with live in-process swarm workers is not idle. The idle timer should only run once those workers have actually reached a terminal or idle state.

## Proposed fix

Treat a headless member in a non-terminal, non-`ready` status as occupancy.

Attached (non-headless) members should be deliberately **excluded**: they already have a client connection that `client_count` tracks, and counting them here would keep the daemon alive forever after a client vanished uncleanly.

A patch with regression tests is ready and I will link it to this issue.

1jehuang:
## Summary
- fix clear, reproducible issues found during open-issue triage
- add Conifer as an approved built-in OpenAI-compatible provider
- add focused regression coverage for each changed behavior
- keep platform-sensitive and design-heavy reports out of this PR

## Issues
Fixes #1090, #1089, #1087, #1086, #1084, #1073, #1072, #1071, #1068, #1060, #1053, #1052, #1051, and #1050.

## Validation
- `cargo test -p jcode-app-core idle_monitor --lib`
- `cargo test -p jcode-provider-core zai --lib`
- `cargo test -p jcode-provider-openrouter-runtime --lib`
- `cargo test -p jcode-base conifer_static_fallback_contains_the_issue_catalog --lib`
- exact CLI provider-list regression test
- `cargo check -p jcode --bin jcode`
- selfdev packaged TUI build
- packaged `jcode provider list --json` returns Conifer with its API-key metadata
- packaged `jcode -p conifer provider current --json` resolves the requested provider to Conifer
- `git diff --check master...HEAD`

The packaged CLI check initially exposed that Conifer parsed as a provider but was omitted from `provider list`. Commit `e24dd976c` added that public discovery registration and a regression test; rebuilding and rerunning both public workflows then passed. The unrelated pre-existing `ui_input.rs` working-tree edit was preserved exactly and excluded.


manoj402kumar:
@1jehuang can you plese check #1029 and make the isse resloved

PR #1193, opened 2026-09-07
fix: safe fixes from September 7 issue triage

1jehuang:
## Status
Draft while targeted regression tests and a combined TUI build run. No contributor branch code was imported.

## Verified so far
- Merge duplicate top-level environment maps in CI, iOS TestFlight, and Windows Smoke. All nine workflow files pass a duplicate-rejecting YAML parse, and other workflow semantics are unchanged.
- Fix the two independently reproduced formatting failures. `cargo fmt --all -- --check` and module declaration checks pass.

Fixes #1191. Related duplicate: #1190. Additional isolated fixes will be added only after verification. Restored parsing does not yet establish that all CI gates pass.

---
*— Jcode agent (automated triage), on behalf of @1jehuang*

Issue #1191, opened 2026-09-07
CI broken since Aug 20: duplicate top-level env in ci.yml, ios-testflight.yml, windows-smoke.yml (Invalid workflow file)


ra-co88:
## Summary

Every GitHub Actions run on `master` (and all feature branches) fails instantly with:

```
Invalid workflow file
.github/workflows/ci.yml
(Line: 18, Col: 1): 'env' is already defined
```

The same error affects `ios-testflight.yml` (line 18) and `windows-smoke.yml` (line 23). All three workflows have **two top-level `env:` blocks**, which GitHub Actions rejects as a duplicate mapping key. Runs show `0s` duration — nothing executes.

## Root cause

Commit e644e42ba ("Mark CI telemetry explicitly", Aug 20) prepended a new top-level block:

```yaml
name: CI

env:
  JCODE_CI: "1"

on:
  ...
```

to `ci.yml`, `ios-testflight.yml`, and `windows-smoke.yml`, each of which already had a top-level `env:` block further down (e.g. `CARGO_TERM_COLOR` / `BUNDLE_ID` / `SCHEME` / `TEAM_ID`). Every push since Aug 20 has had zero CI.

Consequences beyond the parse failure itself: with CI unable to run, unformatted code has also landed on master — `cargo fmt --check` currently fails on `crates/jcode-tui/src/tui/app/input.rs:2931` and `crates/jcode-tui/src/tui/ui_header.rs:850` (from the recent SSH feature wave).

## Evidence

- `gh run list` on this repo: every run since e644e42ba completed in 0s with "Invalid workflow file"
- Current master still has both blocks: `grep -n '^env:' .github/workflows/ci.yml` → lines 3 and 16

## Fix

Fold `JCODE_CI: "1"` into each file's existing `env:` block (one block per file, semantics unchanged — top-level `env` is a flat map). A PR follows with the three-line fix. The formatting fallout can be a separate `cargo fmt` commit.

Found while syncing the ra-co88/jcode fork, whose CI hit the same parse failure.

ra-co88:
The fix is ready on the ra-co88/jcode fork, branched directly from current master:

- Branch: [`fix/ci-duplicate-env`](https://github.com/ra-co88/jcode/tree/fix/ci-duplicate-env) (1 commit ahead of master, cherry-picked cleanly)
- Diff: https://github.com/1jehuang/jcode/compare/master...ra-co88:fix/ci-duplicate-env
[...]

PR #1218, opened 2026-09-10
fix: address ten reproducible issues from September triage

### LINKED ISSUE #1184: Ghostty: focus-gained handler re-arms mode 1004, causing a 42k writes/s feedback loop in an idle session
## Summary

In Ghostty, an idle jcode session burns 66% CPU in the client and 90-106% in Ghostty's renderer thread, purely because the window is focused. Empty session, no prompt typed, nothing streaming.

jcode issues ~42,000 `write()` syscalls per second, 61 bytes each, 2.5 MB/s into a 114x32 terminal. Unfocus the window and it drops to exactly 0. Refocus and it comes straight back.

Two behaviours form a loop:

1. jcode calls `reapply_configured_terminal_modes()` inside its `Event::FocusGained` handler, which re-issues `ESC[?1004h` (`EnableFocusChange`).
2. Ghostty answers any set of mode 1004 by reporting the current focus state. While focused that reply is `CSI I`, which crossterm decodes as `Event::FocusGained`.

jcode's response to a focus event generates another focus event. No timer is involved, which is why no rendering setting affects it.

Suggested labels: `bug`, `priority: high`, `triage: reproducible`, `autonomous: clear`.

## Environment

- jcode v0.81.7 (358226c2a), prebuilt Linux x86_64 binary
- Ghostty 1.3.1
- Ubuntu, GNOME 46, Wayland (also reproduced on X11)
- crossterm 0.29.0 (`Cargo.lock:1426-1428`)
- Terminal 114x32

## Reproduction

Open a Ghostty tab, run `jcode`, type nothing. Sample the client:

```bash
P=<jcode client pid>
W1=$(awk '/^syscw/{print $2}' /proc/$P/io); sleep 5
W2=$(awk '/^syscw/{print $2}' /proc/$P/io); echo $(( (W2-W1)/5 )) writes/s
```

Sample unfocused, then focused, then unfocused again.

| window state | write() syscalls/s | jcode CPU |
|---|---|---|
| not focused | 0 | 0% |
| **focused** | **42,740 / 42,603 / 42,384** | **66-67%** |
| not focused again | 0 | 0% |

Ghostty threads while focused: renderer 104-106%, io-reader 29%, io 25%. A Ghostty window running `sleep 600` sits at 0%, so Ghostty alone does not spin.

## Root cause

`crates/jcode-tui/src/tui/app/local.rs:397`:

```rust
Some(Ok(Event::FocusGained)) => {
    crate::tui::reapply_configured_terminal_modes();
```

That reaches `reapply_terminal_modes_to` (`mod.rs:155-179`), which re-issues the mode set with no rate limit, no dedup, and no check that the modes were ever lost:

```rust
writer.queue(EnableBracketedPaste)?;
if focus_change { writer.queue(EnableFocusChange)?; }   // ESC[?1004h
```

Same call at `remote.rs:395` and `remote.rs:772`. The latter is in `handle_terminal_event_while_disconnected`, so a disconnected client spins too.

`Event::FocusLost` (`local.rs:403`) does not reapply. That asymmetry is why the bug is focus-gated.

Ghostty v1.3.1, `src/termio/stream_handler.zig:755`:

```zig
.focus_event => if (enabled) self.messageWriter(.{
    .focused = self.terminal.flags.focused,
}),
```

Setting mode 1004 emits the current focus state, with no check for whether the mode was already enabled. Re-arming an armed mode still produces a reply. This arrived in Ghostty PR #2196.

The cycle:

```
FocusGained (local.rs:397)
  -> reapply_configured_terminal_modes()   (local.rs:398)
  -> ESC[?1004h inside a 61-byte payload   (mod.rs:165)
  -> Ghostty replies CSI I                 (stream_handler.zig:755)
  -> crossterm decodes Event::FocusGained
  -> repeat
```

Unfocused, Ghostty replies `CSI O`, `FocusLost` does not reapply, and it stops after one hop.

The payload size matches:

```console
$ printf '\033[?2004h\033[?1004h\033[?1000h\033[?1002h\033[?1003h\033[?1015h\033[?1006h\033[=7u' | wc -c
61
```

42,740 writes/s x 61 B = 2.6 MB/s, matching the measured 2.5 MB/s.

## Reproduction without Ghostty

A `pty.fork()` harness running the real jcode binary, emulating one Ghostty behaviour and nothing else: on seeing `b"\x1b[?1004h"` from the slave, write back `b"\x1b[I"` for focused or `b"\x1b[O"` for unfocused. Dumb sink otherwise.

Same binary, same environment, 8 second runs, one byte of difference in the reply:

| simulated state | bytes from jcode | `ESC[?1004h` observed |
|---|---|---|
| focused (reply `ESC[I`) | 18,796,851 (2.35 MB/s) | **308,110** |
| unfocused (reply `ESC[O`) | 1,999 | **1** |

Reproduced three times at 33,860/s, 38,396/s and 41,447/s. Any terminal that answers a 1004 mode-set with the current focus state triggers this, so Ghostty is not a necessary condition.

<details>
<summary>Harness source</summary>

```python
#!/usr/bin/env python3
"""Usage: repro.py <focused:0|1> <seconds>"""
import os, pty, sys, time, select, signal, fcntl, termios, struct

SET_1004 = b"\x1b[?1004h"
FOCUSED = sys.argv[1] == "1"
DURATION = float(sys.argv[2])
REPLY = b"\x1b[I" if FOCUSED else b"\x1b[O"

pid, fd = pty.fork()
if pid == 0:
    env = dict(os.environ)
    env.update({
        "TERM": "xterm-ghostty",
        "TERM_PROGRAM": "ghostty",
        "GHOSTTY_RESOURCES_DIR": "/usr/share/ghostty",
        "COLUMNS": "114", "LINES": "32",
    })
    os.execvpe("jcode", ["jcode"], env)

fcntl.ioctl(fd, termios.TIOCSWINSZ, struct.pack("HHHH", 32, 114, 0, 0))

start = time.time()
total_bytes = reads = replies = 0
buf = b""
while time.time() - start < DURATION:
    r, _, _ = select.select([fd], [], [], 0.2)
    if not r:
        continue
    try:
        data = os.read(fd, 65536)
    except OSError:
        break
    if not data:
        break
    total_bytes += len(data)
    reads += 1
    buf = (buf + data)[-4096:]
    n = buf.count(SET_1004)
    if n:
        buf = buf.replace(SET_1004, b"")
        for _ in range(n):
            os.write(fd, REPLY)
            replies += 1

elapsed = time.time() - start
os.kill(pid, signal.SIGKILL)
os.waitpid(pid, 0)
print(f"focused={FOCUSED} elapsed={elapsed:.1f}s")
print(f"  bytes from jcode : {total_bytes:,} ({total_bytes/elapsed/1e6:.2f} MB/s)")
print(f"  ESC[?1004h seen  : {replies:,} ({replies/elapsed:.0f}/s)")
```
</details>

## Why GNOME Terminal is unaffected

VTE reports focus transitions only, never in response to the mode being set, so nothing answers and no loop forms. That was the original clue: the same session in GNOME Terminal does not spin.

jcode has no terminal-specific branch here. `perf.rs:208` sets `enable_focus_change = true` unconditionally, and only `perf.rs:234` (WSL plus Windows Terminal) disables it. `detect_terminal` does recognise Ghostty, but that only feeds `fragile_glyph_cache`.

## Ruled out by measurement

Each of these was changed on the live affected session with no effect:

| hypothesis | result |
|---|---|
| `display.redraw_fps` 60 -> 30 -> 10 | ~42k writes/s in all three |
| `display.performance` = `reduced`, `minimal` | unchanged |
| `display.latex_rendering` image -> text | unchanged |
| `display.mouse_capture` = false | 43,486 writes/s |
| `display.prompt_entry_animation` = false | 42,320 writes/s |
| `display.pin_todos` = false | 42,274 writes/s |
| `display.prompt_preview` = false | 42,680 writes/s |
| X11 versus Wayland | unchanged |
| disk logging | log grew 0 KB/s during the spin |

The redraw scheduler cannot cause this. `redraw_schedule.rs:16-20` floors intervals at 250 ms and 5 s, and `redraw_schedule.rs:428` returns `REDRAW_DEEP_IDLE` when `!client_focused`, so focus makes the schedule slower. The fastest scheduler path is 4 draws/s, three orders of magnitude below what I measured.

These writes never pass through ratatui. They go out through crossterm's `queue`/`flush` in `mod.rs:155-179`, so `client:draw-stats` cannot see them.

## Related issues

- **#392** (closed) diagnosed streaming-tail markdown re-render. Different cause: this session is empty and nothing streams. I misattributed this bug to #392 before measuring an empty session.
- **#1071** (closed, fixed) was `RunningTool` forcing a full-frame flush at roughly 1/s. Different mechanism and four orders of magnitude apart.
- **#540** (open) covers input latency from render-path work. Related area, different mechanism.

Nothing open covers focus-triggered mode re-arming.

## Fix

`reapply_terminal_modes_to` already takes a `focus_change` flag (`mod.rs:159`). The three focus-path callers should not re-arm the one mode whose reply re-triggers them. Non-focus callers keep re-arming it, since the defensive re-arm matters after an external program may have cleared the modes. Focus reporting is still enabled once at startup, so nothing changes functionally.

A patch doing this is on a branch, since this repo does not accept external pull requests: https://github.com/MatCer/jcode/commit/74ae3a8a2d98f9b3268ec6d1ec2fe24c88424d20

Three call sites pass `false`, plus a regression test on the emitted bytes. Measured on the same machine:

| binary | simulated state | `ESC[?1004h` loops in 8s |
|---|---|---|
| v0.81.7 | focused | **331,577** (41,447/s), 20.2 MB |
| patched | focused | **1**, 3.3 KB |
| patched | unfocused | 1, 2.2 KB |

The remaining occurrence is the normal startup mode-set.

## Impact

Any jcode user in Ghostty, or any terminal that reports focus state on mode-set, pays about one CPU core in jcode plus another in the terminal renderer for as long as the window is focused. I found this because the machine got hot whenever the jcode tab was in front. The workarounds are keeping the tab unfocused or using a different terminal.

### LINKED ISSUE #1186: Bug: slash autocomplete omits all installed skills despite '18 loaded' header (v0.83.0, Windows)
## Summary

In Jcode **v0.83.0 (`1301e382b`) on Windows x86_64**, the startup header reports **`Skills: 18 loaded`**, and `skill_manage action=list` confirms those 18 skills, but typing `/` presents a suggestion list containing **only the 111 built-in commands**. None of the installed skill commands are included.

This is not simply an eight-row viewport hiding skills further down the list. The screenshot's pagination count and final eight entries exactly match the complete built-in-only command registry in the installed revision.

## Environment

- Jcode: `jcode v0.83.0 (1301e382b)` (confirmed with `jcode --version`)
- OS / architecture: Windows / x86_64
- Observed: 2026-09-06
- Working directory: `~/dev`
- Skill location: `%USERPROFILE%\.agents\skills\<skill-name>\SKILL.md`
- 18 installed skills, confirmed by `skill_manage action=list`
- No claim yet that this is Windows-specific. The implicated client initialization and autocomplete code is shared.

## Observed reproduction

1. Have valid skills installed under `~/.agents/skills/`.
2. Open the TUI. In the affected session, the startup header displays `Skills: 18 loaded`.
3. Type a bare `/` and browse the command suggestions to the end.
4. Observe that only built-in commands appear. The final page has `↑103` on its first row and eight visible rows, ending in `/provider-test-coverage`.
5. Ask the agent to run `skill_manage action=list`. It reports `Loaded skills: 18`, including `/ponytail`, `/humaniser`, and the Google ADK and Matt Pocock skills, with their existing `SKILL.md` paths.

These steps describe the reported session. A fresh-session automated TUI reproduction has not yet been run, and the exact startup/event timing needed to trigger the omission is not established.

## Screenshot evidence, transcribed

The original screenshot was supplied in the debugging conversation. Its relevant content is transcribed here rather than claiming an image attachment:

```text
Skills: 18 loaded
~/dev

1> /
/thinking-display       Show/hide the model's thinking text (off/full/current)  ↑103
/tool-call-details      Show/toggle dimmed technical details on tool rows with an intent
/fast-macos-release     Publish a prepared macOS arm64 build immediately; CI adds other platforms
/onboarding-preview     Preview the first-run onboarding screen
/refresh-model-list     Refresh provider model catalogs
/compact-notifications  Show/toggle single-line swarm/file-activity notifications
/show-agentgrep-output  Show/toggle full agentgrep search output inline in chat
/provider-test-coverage Show live-test evidence for the current provider/model
```

There is no `+N more` indicator below the final row. The selected final command is `/provider-test-coverage`.

### Why this proves an omission rather than scrolling

The source of **the exact installed revision**, not just a different local checkout, was fetched and checked:

- Extracting `RegisteredCommand::public(...)` and `RegisteredCommand::remote(...)` entries gives **111 built-ins**.
- Sorting these by command length, then name, reproduces the screenshot's final eight commands exactly.
- The screenshot shows **103 preceding entries + 8 visible entries = 111 total**.
- With the 18 installed, distinct skill names included, the expected total is **129**. Several installed skill commands are longer than `/provider-test-coverage`, so they would also change the final page.

The installed skill names are:

```text
google-agents-cli-adk-code
google-agents-cli-deploy
google-agents-cli-eval
google-agents-cli-observability
google-agents-cli-publish
google-agents-cli-scaffold
google-agents-cli-workflow
humaniser
matt-pocock-grill-me
matt-pocock-handoff
matt-pocock-teach
matt-pocock-writing-great-skills
ponytail
ponytail-audit
ponytail-debt
ponytail-gain
ponytail-help
ponytail-review
```

## Expected behavior

Installed, available skills should be candidates in the same slash autocomplete list as built-in commands, both for a bare `/` and matching prefixes such as `/pony` or `/human`.

The header, skill listing, and autocomplete should agree on which skills are available for the active session. Users should not have to first submit a prompt, invoke a missing skill, restart, or manually refresh `/skills` to discover already-installed skills.

### 1. Skill commands are explicitly intended to appear in autocomplete

[`state_ui_input_helpers.rs` at `1301e382b`](https://github.com/1jehuang/jcode/blob/1301e382b/crates/jcode-tui/src/tui/app/state_ui_input_helpers.rs):

`command_candidates()`:

1. Returns the cached candidates if present.
2. Collects non-hidden registered commands.
3. Calls `current_skills_snapshot()` and adds `/<skill.name>` entries labeled `Activate skill`.
4. In remote/client mode, also adds names from `remote_skills`.
5. Caches the assembled vector.

`rank_suggestions()` ranks a bare `/` as a prefix match for every candidate, then sorts ties by command length and name. It does not intentionally exclude skills.

### 2. The minimal remote-client startup begins with empty skill data

These paths were also checked against the exact installed revision:

- [`tui_lifecycle.rs:1327-1338`](https://github.com/1jehuang/jcode/blob/1301e382b/crates/jcode-tui/src/tui/app/tui_lifecycle.rs#L1327-L1338): `new_for_remote_with_options()` constructs `Registry::empty()` and calls `new_minimal_with_session()`.
- [`tui_lifecycle.rs:350-355`](https://github.com/1jehuang/jcode/blob/1301e382b/crates/jcode-tui/src/tui/app/tui_lifecycle.rs#L350-L355): the minimal constructor initializes `App.skills` with `SkillRegistry::default()`.
- [`tool/mod.rs:302-307`](https://github.com/1jehuang/jcode/blob/1301e382b/crates/jcode-app-core/src/tool/mod.rs#L302-L307): `Registry::empty()` also contains an empty skill registry.

Thus neither of these client-local sources initially contains the installed global skills. Here, "remote" means the client/server runtime path, not necessarily an SSH connection.

### 3. Snapshot fallback does not load missing global skills

[`state_ui_runtime.rs`](https://github.com/1jehuang/jcode/blob/1301e382b/crates/jcode-tui/src/tui/app/state_ui_runtime.rs):

`current_skills_snapshot()` normally reads `self.registry.skills()`, falling back to `self.skills` only when the read lock cannot be acquired. A successfully read **empty** registry is accepted. It then composes the project-local overlay, not a fresh global registry load.

For SSH, it returns `self.skills.clone()` directly. Any repair must retain remote-host isolation rather than inadvertently enumerating local-machine skills for SSH sessions.

### 4. Server history is an existing skill synchronization point

[`remote/server_events.rs:1732-1733`](https://github.com/1jehuang/jcode/blob/1301e382b/crates/jcode-tui/src/tui/app/remote/server_events.rs#L1732-L1733):

```rust
app.remote_skills = skills;
app.invalidate_command_candidates_cache();
```

Therefore it would be inaccurate to claim there is no cache invalidation at all. The important next check is whether the affected fresh/blank client receives and applies a history payload containing its effective skills before autocomplete is first populated, and whether any other skill update path leaves the candidates cached as built-ins only.

`refresh_skills_snapshot()` explicitly reloads global skills and invalidates candidates, but that is a separate path from merely typing `/`.

## Diagnosis and remaining uncertainty

**Confirmed:** all 18 skills are absent from this screenshot's complete autocomplete candidate list despite being discoverable by the agent and advertised as loaded in the header.

**Leading explanation:** incomplete initialization/synchronization of the minimal client's skill data, potentially preserved by the candidate cache. The empty minimal registry path explains how a built-in-only list can be constructed without skill files themselves being invalid.

**Not yet confirmed:** which startup/history/refresh event was missing or delayed in this particular session. Live debug inspection was attempted with `jcode debug help`, but the running instance returned `Debug control is disabled`. No debug setting was changed, no running session was restarted, and no fix or end-to-end verification is claimed.

## Suggested investigation and regression coverage

- Reproduce using the real `new_for_remote_with_options()` / minimal-client path, not only a fully initialized standalone test app.
- Check a blank session before its first submitted message, then after history synchronization.
- Make the effective skill metadata available to autocomplete during client startup and invalidate cached candidates whenever that metadata changes. Avoid loading skills from disk on every rendered frame.
- Verify a bare `/` includes a fixture skill, and a matching prefix includes the same skill.
- Warm autocomplete while skill data is empty, deliver the real metadata/update event, and verify the fixture appears without restarting or running `/skills` first.
- Verify the header count and autocomplete's unique skill candidates agree for the active session.
- Preserve project-local isolation between simultaneous sessions in different working directories, and local/remote-host isolation for SSH.
- Cover installed global skills under `~/.agents/skills/`, not just project-local fixtures followed by an explicit refresh.

## Related issues

- #431: closed report about server-side `skill_manage reload_all` not syncing the TUI snapshot and autocomplete. This observation does **not** require a preceding reload operation and concerns already-installed skills in the bare-slash menu.
- #59: closed stale-registry report for direct slash invocation.
- #457: closed project-local/global skill scoping issue. Any fix here should preserve its session-isolation invariant.

An exact currently open duplicate was not found in searches for `skills slash` and `skills autocomplete`.

### LINKED ISSUE #1187: Desktop notification children still become zombies after the observer-hook fix (#935)
# Desktop notification children still become zombies after the observer-hook fix (#935)

## Summary

The observer-hook fix discussed in #935 does not cover the desktop notification helpers. They still discard `std::process::Child` immediately after spawning `notify-send` (Linux) or `osascript` (macOS).

Rust does not wait for a child when its `std::process::Child` handle is dropped. Once the notification command exits, it remains a zombie until its parent waits for it or exits. In a long-lived jcode process this accumulates across notifications.

## Environment and observed impact

- Linux x86_64, initially observed while using jcode v0.81.6.
- The same notification code is still present in upstream master `983b532702d2b9cb7766182d566d39081ad1abcb`, inspected on September 6, 2026.
- Live inspection found dozens of jcode-owned zombies, approximately 40 of them `notify-send`. Other child-lifecycle paths also contributed, so the overall count should not be attributed entirely to notifications.
- After applying local child-reaping fixes and reloading the actual daemon and clients, repeated `/proc` scans found zero jcode-owned zombies. This is a bounded observation, not proof that every possible child-lifecycle leak has been eliminated.
- Zombie processes consume process-table entries, not active CPU. Accumulation can eventually contribute to process/PID limits. Resource exhaustion was not reproduced in this investigation.

## Confirmed code paths

1. [`send_desktop_notification_rich`, app-core notifications.rs, lines 701–719](https://github.com/1jehuang/jcode/blob/983b532702d2b9cb7766182d566d39081ad1abcb/crates/jcode-app-core/src/notifications.rs#L701-L719)
2. [`send_desktop_notification`, setup-hints cli_launch_hints.rs](https://github.com/1jehuang/jcode/blob/983b532702d2b9cb7766182d566d39081ad1abcb/crates/jcode-setup-hints/src/cli_launch_hints.rs)

Both use the following pattern:

```rust
let _ = std::process::Command::new("notify-send")
    // arguments and stdio setup
    .spawn();
```

The macOS `osascript` branches use the same ownership pattern. Linux was tested. The macOS consequence is inferred from the code and Unix child semantics, not from a macOS runtime test.

## Reproduce through the production notification helper

For deterministic regression coverage, use a fake `notify-send` instead of relying on a desktop session or notification service:

1. In a temporary directory, create an executable named `notify-send`:

   ```sh
   #!/bin/sh
   printf '%s\n' "$$" >> "$JCODE_NOTIFY_PID_FILE"
   exit 0
   ```

2. Add a Linux-only test in `crates/jcode-app-core/src/notifications.rs` that calls `send_desktop_notification_rich("reap-test", None, "test", None)` five times.
3. Launch the test executable through Cargo with that temporary directory prepended to `PATH`, an empty PID file referenced by `JCODE_NOTIFY_PID_FILE`, and `--test-threads=1`. Set these variables in the parent shell, rather than modifying process-wide environment variables from a multithreaded Rust test.
4. Keep the test process alive. First wait with a deadline until all five child PIDs have been recorded. Then poll `/proc/<pid>/stat` for those exact children.
5. **Before the fix:** exited children remain with state `Z` and the test process as PPID. Fail if any of the recorded children still exists after a bounded wait.
6. **After the fix:** all five `/proc/<pid>` entries disappear while the parent test process remains alive.

Run the equivalent check for the setup-hints helper too.

Important: do not assert only that the initial zombie count is zero, or stop polling as soon as the count is zero. A child may not have exited yet, producing a false pass. Record exact PIDs, wait for the children to be launched, and require their disappearance. Avoid a process-wide `waitpid(-1, ...)` sweep in the test, which would hide the bug by reaping the leaked children itself.

A local notification regression using five real notifications failed with five zombies when the reaper was removed and passed with the reaper restored. The fake-command/PID approach above is the recommended more deterministic upstream test design, not an assertion that such a test is already in the fork.

## Measured deterministic reproduction

On September 6, 2026, an isolated Linux harness compiled the unchanged notification function bodies extracted from upstream `983b53270` and the fork fixes revision `21922c06c`. It supplied a fake `notify-send`, recorded every child's PID, and kept each parent alive during observation:

| Source | Helper | Children launched | Persistent zombies | Child PIDs disappeared |
| --- | --- | ---: | ---: | ---: |
| Upstream | app-core rich notification | 5 | 5 | 0 |
| Upstream | setup-hints notification | 5 | 5 | 0 |
| Fixed | app-core rich notification | 5 | 0 | 5 |
| Fixed | setup-hints notification | 5 | 0 | 5 |

Upstream children were still `Z` after a three-second observation window, with the test parent as PPID. Fixed children disappeared before the deadline while the parent remained alive. The harness reaped its intentional upstream zombies only after recording the result and exiting their parent. All recorded child PIDs were gone at cleanup.

**Scope:** this is a synthetic source-extraction harness, not a full jcode integration test or a macOS test. It demonstrates both notification ownership defects and the before/after behavior without a notification daemon or a live jcode session.

## Real-crate public API integration check

Also ran a Cargo integration test against the actual `jcode-app-core` crate at feature revision `a39df8e3f`, rather than copied function bodies. Result: **2 passed, 0 failed**, with one ignored helper test explicitly invoked as the subprocess entry point by the two parent tests.

The test exercised both exported functions, `send_desktop_notification_rich` and `send_desktop_notification`, through the real crate dependency graph:

- All five subprocesses received the expected `--app-name=jcode`, title, and body arguments.
- Both APIs returned while the notifier processes were deliberately held open behind a release-file gate. This verifies nonblocking behavior without relying on a tight timing threshold.
- After release, all five recorded child PIDs disappeared while their parent test process was still alive.
- With `notify-send` absent from the child process's `PATH`, both public APIs returned without failure.

This crosses the real app-core/platform/subprocess boundary using a controlled executable fixture. It does not claim that a desktop notification was visually displayed, that the setup-hints CLI workflow was integration-tested, or that macOS was tested.

The initial invocation lacked the scratch-directory environment variable and failed in test setup, before launching any notification children. Supplying that variable explicitly produced the successful result above.

## Suggested narrow fix

Transfer each successfully spawned `std::process::Child` to the existing `jcode_base::platform::reap_detached` helper, or the appropriate equivalent for the setup-hints dependency boundary. Waiting should be off the notification caller's critical path.

The relevant initial patch is available for reference in [gfx-labs/jcode commit e74371b7c](https://github.com/gfx-labs/jcode/commit/e74371b7c). That commit also touches other child-lifecycle paths. Please isolate the notification change and its regression tests rather than assuming the whole commit is required.

## Related lifecycle paths to investigate separately

Our broader local patch series is on [gfx-labs/jcode:fixes](https://github.com/gfx-labs/jcode/tree/fixes). It also contains unrelated test-lock fixes, so it is not a minimal patch for this report.

- `McpClient::drop` calls `start_kill()` without explicitly awaiting termination. Related existing report: #1036. Tokio child cleanup differs from `std::process::Child`, so this requires its own reproducer and should not be treated as identical to the confirmed notification bug.
- Exec-based daemon/client reloads preserve the parent PID but discard userspace cleanup state. Child shutdown and reaping must complete before exec. Shared and per-session MCP children, power inhibitors, and cancelled background commands warrant coverage.
- Awaiting an aborted task's `JoinHandle` establishes that task teardown finished, but does **not by itself prove** that every Tokio child has been reaped. Cancellation/reload tests should observe exact child PIDs disappearing rather than infer success from task status alone.
- A startup sweep can recover already-exited children left by earlier versions, but it is not a substitute for correct ownership and waiting. A broad reaper must not steal exit statuses from active child owners.

## Acceptance criteria

- Repeated desktop notifications do not leave persistent zombie children while the parent remains running.
- Both notification helpers are covered.
- A deterministic PID-based regression fails on affected upstream code and passes with the fix.
- Notifications remain nonblocking for the calling thread.
- Validation uses the newly built code. For end-to-end checks, confirm the long-lived daemon/client binary revision or use an isolated test daemon, rather than accidentally exercising the old installed server.

## Standalone reproduction script

Requirements: Linux with `/proc`, Python 3, Rust, Git, and a local clone of this repository. Save the script below as `repro_notifications.py` outside tracked source files. The upstream and fixed revisions must be available locally. If needed, fetch the fixes branch without checking it out or merging it:

```sh
git fetch https://github.com/gfx-labs/jcode.git fixes
export JCODE_SCRATCH_DIR="$(mktemp -d)"
python3 /path/to/repro_notifications.py
```

The script prints four result records and the location of `results.json`. It does not modify repository files or restart jcode. Its temporary binaries and source remain in the scratch directory for inspection.

<details>
<summary>Reproducer source</summary>

```python
#!/usr/bin/env python3
"""Isolated synthetic PID regression using unchanged extracted production helpers."""
import ctypes
import json
import os
from pathlib import Path
import subprocess
import tempfile
import time

BASE = '983b532702d2b9cb7766182d566d39081ad1abcb'
FIXED = '21922c06c811dae0cedb8675575134b0ff24640f'
# Adopt and reap only this harness's descendants when their test parent exits.
libc = ctypes.CDLL(None, use_errno=True)
if libc.prctl(36, 1, 0, 0, 0) != 0:  # PR_SET_CHILD_SUBREAPER
    raise OSError(ctypes.get_errno(), 'prctl subreaper failed')
root = Path(tempfile.mkdtemp(prefix='notification-repro-', dir=os.environ['JCODE_SCRATCH_DIR']))
fixture = root / 'bin'
fixture.mkdir()
notifier = fixture / 'notify-send'
notifier.write_text('#!/bin/sh\nprintf \'%s\\n\' "$$" >> "$JCODE_NOTIFY_PID_FILE"\nexit 0\n')
notifier.chmod(0o700)

def source(rev, path):
    return subprocess.check_output(['git', 'show', f'{rev}:{path}'], text=True)

def function(text, signature):
    start = text.index(signature)
    end = text.index('\n}', start) + 2
    return text[start:end]

def state(pid):
    try:
        text = Path(f'/proc/{pid}/stat').read_text()
        return text[text.rfind(')') + 2:].split()[:2]
    except FileNotFoundError:
        return None

results = []
for label, rev in [('upstream', BASE), ('fixed', FIXED)]:
    for helper in ['app_core', 'setup_hints']:
        name = f'{label}_{helper}'
        if helper == 'app_core':
            code = function(source(rev, 'crates/jcode-app-core/src/notifications.rs'), 'pub fn send_desktop_notification_rich(')
            platform = function(source(rev, 'crates/jcode-base/src/platform.rs'), 'pub fn reap_detached(')
            code = 'extern crate self as jcode_base;\npub mod platform {\n' + platform + '\n}\n' + code
            call = 'send_desktop_notification_rich("reap-test", None, "test", None);'
        else:
            text = source(rev, 'crates/jcode-setup-hints/src/cli_launch_hints.rs')
            code = function(text, 'fn send_desktop_notification(')
            if label == 'fixed':
                code = function(text, 'fn reap_detached(') + '\n' + code
            call = 'send_desktop_notification("reap-test", "test");'
        code += '\nfn main() { for _ in 0..5 { ' + call + ' } let mut s = String::new(); std::io::stdin().read_line(&mut s).unwrap(); }\n'
        src = root / f'{name}.rs'
        src.write_text(code)
        binary = root / name
        subprocess.run(['rustc', '--edition=2024', str(src), '-o', str(binary)], check=True)
        pidfile = root / f'{name}.pids'
        pidfile.write_text('')
        env = dict(os.environ, PATH=str(fixture) + os.pathsep + os.environ['PATH'], JCODE_NOTIFY_PID_FILE=str(pidfile))
        child = subprocess.Popen([str(binary)], stdin=subprocess.PIPE, text=True, env=env)
        pids = []
        try:
            deadline = time.monotonic() + 5
            while time.monotonic() < deadline:
                pids = [int(p) for p in pidfile.read_text().splitlines()]
                if len(pids) == 5:
                    break
                time.sleep(.02)
            assert len(pids) == 5, (name, pids)
            deadline = time.monotonic() + 3
            while time.monotonic() < deadline:
                states = [state(pid) for pid in pids]
                if label == 'fixed' and all(s is None for s in states):
                    break
                time.sleep(.02)
            states = [state(pid) for pid in pids]
            assert child.poll() is None, 'Parent exited before observation'
            if label == 'upstream':
                assert all(s == ['Z', str(child.pid)] for s in states), states
            else:
                assert all(s is None for s in states), states
            result = dict(case=name, children=5, zombies=sum(s is not None and s[0] == 'Z' for s in states), disappeared=sum(s is None for s in states), parent_alive=True)
            results.append(result)
            print(json.dumps(result), flush=True)
        finally:
            child.communicate('\n', timeout=5)
            # The upstream case intentionally creates zombies. Reap them here,
            # after the observation and test-parent exit, never during the test.
            for pid in pids:
                try:
                    os.waitpid(pid, 0)
                except ChildProcessError:
                    pass
            assert all(state(pid) is None for pid in pids), 'Reproducer left a child behind'

output = dict(upstream=BASE, fixed=FIXED, scope='synthetic harness, unchanged extracted functions, Linux only', results=results, cleanup='all recorded child PIDs gone')
(root / 'results.json').write_text(json.dumps(output, indent=2) + '\n')
print(f'ARTIFACT={root}/results.json')
```

</details>

## Cargo integration test source

Save the following as `crates/jcode-app-core/tests/notification_issue_acceptance.rs` in a test checkout. Then run:

```sh
export JCODE_SCRATCH_DIR="$(mktemp -d)"
cargo test --profile selfdev -p jcode-app-core --test notification_issue_acceptance -- --nocapture
```

The notifier fixture and outputs are created under that scratch directory. No process-wide environment changes are made by the multithreaded Rust test process.

<details>
<summary>Real-crate integration test</summary>

```rust
#![cfg(target_os = "linux")]

use std::io::Write;
use std::path::PathBuf;
use std::process::{Child, Command, Stdio};
use std::time::{Duration, Instant};

#[test]
#[ignore = "subprocess entry point for notification API acceptance tests"]
fn notification_api_child() {
    let dir = PathBuf::from(std::env::var_os("JCODE_NOTIFICATION_ACCEPTANCE_DIR").unwrap());
    for _ in 0..3 {
        jcode_app_core::notifications::send_desktop_notification_rich(
            "acceptance title", None, "acceptance body", None,
        );
    }
    for _ in 0..2 {
        jcode_app_core::notifications::send_desktop_notification("acceptance title", "acceptance body");
    }
    std::fs::write(dir.join("ready"), "returned").unwrap();
    let mut line = String::new();
    std::io::stdin().read_line(&mut line).unwrap();
}

struct Probe { child: Child, dir: PathBuf }
impl Drop for Probe {
    fn drop(&mut self) {
        let _ = std::fs::write(self.dir.join("release"), "release");
        if let Some(mut stdin) = self.child.stdin.take() { let _ = stdin.write_all(b"done\n"); }
        let _ = self.child.wait();
    }
}

fn wait_for(mut condition: impl FnMut() -> bool, description: &str) {
    let deadline = Instant::now() + Duration::from_secs(10);
    while !condition() {
        assert!(Instant::now() < deadline, "timed out: {description}");
        std::thread::sleep(Duration::from_millis(20));
    }
}

fn probe(mode: &str) -> Probe {
    use std::os::unix::fs::PermissionsExt;
    let root = PathBuf::from(std::env::var_os("JCODE_SCRATCH_DIR").unwrap());
    let dir = root.join(format!("notification-api-{}-{mode}", std::process::id()));
    std::fs::create_dir(&dir).unwrap();
    let bin = dir.join("bin");
    std::fs::create_dir(&bin).unwrap();
    if mode == "present" {
        let script = bin.join("notify-send");
        std::fs::write(&script, "#!/bin/sh\nprintf '%s|%s|%s|%s\\n' \"$$\" \"$1\" \"$2\" \"$3\" >> \"$JCODE_NOTIFICATION_ACCEPTANCE_DIR/pids\"\nwhile [ ! -e \"$JCODE_NOTIFICATION_ACCEPTANCE_DIR/release\" ]; do /usr/bin/sleep 0.02; done\n").unwrap();
        std::fs::set_permissions(script, std::fs::Permissions::from_mode(0o700)).unwrap();
    }
    let child = Command::new(std::env::current_exe().unwrap())
        .args(["--exact", "notification_api_child", "--ignored", "--nocapture"])
        .env("PATH", &bin)
        .env("JCODE_NOTIFICATION_ACCEPTANCE_DIR", &dir)
        .stdin(Stdio::piped()).stdout(Stdio::null()).stderr(Stdio::inherit())
        .spawn().unwrap();
    Probe { child, dir }
}

#[test]
fn public_notification_apis_are_nonblocking_and_reap_children() {
    let mut p = probe("present");
    wait_for(|| p.dir.join("ready").exists(), "both public notification APIs return while notifier is gated");
    wait_for(|| std::fs::read_to_string(p.dir.join("pids")).unwrap_or_default().lines().count() == 5, "five real subprocess launches");
    let records = std::fs::read_to_string(p.dir.join("pids")).unwrap();
    let pids: Vec<u32> = records.lines().map(|line| {
        let (pid, args) = line.split_once('|').unwrap();
        assert_eq!(args, "--app-name=jcode|acceptance title|acceptance body");
        pid.parse().unwrap()
    }).collect();
    assert!(pids.iter().all(|pid| PathBuf::from(format!("/proc/{pid}")).exists()));
    assert!(p.child.try_wait().unwrap().is_none());
    std::fs::write(p.dir.join("release"), "release").unwrap();
    wait_for(|| pids.iter().all(|pid| !PathBuf::from(format!("/proc/{pid}")).exists()), "all five children are reaped while their parent remains alive");
    assert!(p.child.try_wait().unwrap().is_none());
    println!("PUBLIC_API_ACCEPTANCE: 5 launches, arguments correct, nonblocking return, 5 child PIDs gone, parent alive");
}

#[test]
fn public_notification_apis_tolerate_missing_notifier() {
    let mut p = probe("missing");
    wait_for(|| p.dir.join("ready").exists(), "public API returns without notify-send installed");
    assert!(!p.dir.join("pids").exists());
    assert!(p.child.try_wait().unwrap().is_none());
    println!("PUBLIC_API_ACCEPTANCE: missing notify-send is nonfatal for both public notification APIs");
}
```

</details>

### LINKED ISSUE #1188: A hand-written `[sponsors] enabled = false` is undone after one config save and reload
**Source commit:** `983b53270` (upstream `master`, 2026-09-06; `repair_frozen_sponsors_optout` unchanged)
**Upstream:** https://github.com/1jehuang/jcode
**Binary identity:** `jcode v0.81.66-dev (0d008c720)`
**Platform:** macOS 26.6.2, arm64

## Summary

Writing `[sponsors]\nenabled = false` by hand into `config.toml` is respected on
the next load, but the opt-out disappears after any process saves config and a
later load reads it back. Observed twice on this host (2026-09-05 13:xx append,
gone by 22:26; re-appended, would have been lost on the next save).

## Cause (from source)

1. `Config::save` (`crates/jcode-base/src/config/config_file.rs`) serializes the
   whole struct. `sponsors_is_default` (`config.rs:837`) skips the table only when
   `enabled` is true, so an opt-out is written back as **two** keys:
   `enabled = false` plus the default `endpoint`.
2. `repair_frozen_sponsors_optout` (`config_file.rs:79`) treats exactly that
   shape, two keys with `enabled = false` and a default endpoint, as a
   machine-frozen legacy value and resets it to `SponsorsConfig::default()`
   (enabled) in memory.
3. The following save omits the now-default table. The user's opt-out is gone.

The heuristic cannot distinguish "hand-written opt-out round-tripped once through
save" from "frozen by the legacy default". Every hand-written opt-out becomes the
former after the first save.

## Reproduction

```text
printf '\n[sponsors]\nenabled = false\n' >> ~/.jcode/config.toml
grep -A2 '^\[sponsors\]' ~/.jcode/config.toml   # table absent
```

## Expected

An explicit `enabled = false` survives save and reload indefinitely.

## Proposed fix

Either serialize the opt-out as the single key the user wrote (skip `endpoint`
when it equals the default), or mark the repair with a one-time migration flag
instead of inferring it from table shape. A regression test: write
`[sponsors] enabled = false`, load, save, load, assert `enabled == false`.

## Workaround in use

Pair the opt-out with a non-default endpoint, which the heuristic respects:

```toml
[sponsors]
enabled = false
endpoint = "http://127.0.0.1:1"
```

### LINKED ISSUE #1189: `bg output` fails while a timeout-promoted command is still Running (output artifact created only at completion)
**Upstream:** https://github.com/1jehuang/jcode
**Source commit:** `1b99ae621` (upstream `master`, 2026-09-06)
**Binary identity:** `jcode v0.83.71-dev (10a4f385a)`, local branch on top of master
**Platform:** macOS 26.6.2, arm64

## Summary

When a foreground `bash` call exceeds its timeout and Jcode promotes it to a background task,
the task's `.output` artifact is not created until the command finishes. `bg output <id>` during
the Running state returns `Output not found for task`, so an agent cannot inspect partial output of
exactly the commands it most needs to watch (long builds, tests, uploads).

The status artifact is written immediately with `pid: null`; only the output artifact is missing.

## Reproduction (two live receipts, 2026-09-07)

Tool call: `bash` with `command: "echo start; sleep 20; echo done"`, `timeout: 2000`.

1. Jcode reports "exceeded the foreground timeout after 2.0s and is continuing in background",
   task `836444a36r`.
2. Immediately, `bg output 836444a36r` returns
   `Error: Output not found for task: 836444a36r. Task may not exist or output file was deleted.`
3. Filesystem at the same moment (`$TMPDIR/jcode-bg-tasks/`):
   - `836444a36r.status.json` exists, `"status": "running"`, `"pid": null`
   - `836444a36r.output` does not exist
4. After completion the output file appears with the full content.

An earlier receipt on 2026-08-31 with `v0.81.5-dev (422da2bc0)` showed the same behavior.

## Why the regression test misses it

`crates/jcode-app-core/src/tool/bash_tests.rs::test_foreground_timeout_promotes_and_command_keeps_running`
waits for `BackgroundTaskStatus::Completed` before calling `background::global().output(&task_id)`.
It never reads the output while the task is Running, so the gap is not covered.

## Expected

Promotion creates an empty retained output artifact before returning the task id, so
`bg output` succeeds (possibly empty) while Running, and the wrapper replaces it with the
complete output at termination. `pid: null` is fine for the in-process adoption path.

## Suggested test change

In the existing test, after promotion and before waiting for completion, assert
`background::global().output(&task_id).is_some()`. It is red today.

## Workaround in use

Start commands expected to outlive the timeout with `run_in_background: true` and consume them
through `bg`; do not rely on automatic promotion for anything whose partial output matters.

### LINKED ISSUE #1194: Scheduled tasks fall back to headless execution while the TUI is still open
When a scheduled task targets a session still open in the terminal, the task runs and changes files but the TUI stays on the previous response. The user cannot see its progress or completion.

Observed with Homebrew jcode 0.84.0 on macOS arm64 (`jcode version` reports `v0.84.0-dev (unknown)`). The same relevant code is present in master at 4e85d4423cfd248236044564aa8e4d48ce0ae473.

### Reproduction

1. Keep a jcode TUI session open and idle.
2. Schedule a task targeting that existing session.
3. Wait until it becomes due.
4. Compare the TUI with the session journal and files changed by the task.

Expected: the existing TUI receives the scheduled turn's text, tool activity and terminal completion event.

Actual: the scheduled work executes through headless resume; the TUI retains the previous response. The daemon log at the due time contains (session ID redacted):

```text
Ambient runner: live delivery for <session> fell back to headless resume: Client must Subscribe with a working_dir before sending stateful requests
```

### Cause and deterministic reproduction

`AmbientRunnerHandle::notify_live_session` opens a fresh `Client` and sends `NotifySession` without subscribing. `Request::is_lightweight_control_request()` excludes that request, so the server's initial-request gate rejects it before `handle_notify_session` can run. The scheduler catches this error and calls `resume_dead_session_with_reminder`, which uses a separate `Agent` and capture-mode output even though the original TUI is still connected.

A regression test using a real server on an isolated local socket and a fake streaming provider reproduces the problem without credentials or a real model: calling `deliver_scheduled_direct_item` completes, but the original subscribed client times out waiting for text and `Done { id: 0 }`.

I have a small proposed fix routing this explicitly session-targeted notification through the existing lightweight control handler, reusing `handle_notify_session`. The same regression then receives both output and completion; an unknown target still returns an error.

### Proposed patch and validation

PR creation was rejected because this repository currently has `pull_request_creation_policy: collaborators_only`. The proposed fix is available for maintainer review:

- Commit: https://github.com/eohjun/jcode/commit/6a945d3d88ed8b7188832aa12c929f1f52eef843
- Branch: https://github.com/eohjun/jcode/tree/fix/scheduled-live-delivery
- Base: upstream master `4e85d4423cfd248236044564aa8e4d48ce0ae473`

The patch changes three files: request classification, routing to the existing notification handler, and a regression test. Validation on that base:

- `cargo test -p jcode-app-core --no-default-features ambient::runner::runner_tests`: 5 passed
- `cargo test -p jcode-app-core --no-default-features notify_session_`: 2 passed
- `cargo test -p jcode-protocol`: 79 passed
- Changed-file rustfmt check and `git diff --check`: passed
- The new regression fails before the fix and passes afterward.

Repository-wide test-size, code-size, panic and swallowed-error ratchets already fail on the unmodified base; all four produce identical failures with this patch. No baseline files were changed. Full all-feature/workspace CI was not run locally.

### LINKED ISSUE #1201: jcode-tui test suite deadlocks on Windows above 2 threads (lock-order inversion, silently skips tests)
`cargo test -p jcode-tui --lib` does not finish on Windows at the default
thread count. It is not slow, it deadlocks: every worker blocks and the run
hangs indefinitely, so every test after the wedge is silently never executed.

This is on current `master` (verified at `e65e47c31`), not a fork-specific
problem. It matters beyond Windows because a hung run reports nothing, so the
suite has been quietly under-reporting rather than failing loudly.

## Reproduce

```bash
cargo test -p jcode-tui --lib     # hangs forever on Windows
cargo test -p jcode-tui --lib -- --test-threads=2   # completes
```

Exactly `--test-threads` many workers wedge. 1 and 2 are clean; 4, 8 and 16 all hang.

## Cause

Two process-global test mutexes:

- `jcode_base::storage::lock_test_env()` (guards `JCODE_HOME` / env vars)
- `jcode_tui::tui::ui::render_state_test_lock()` (guards global render state)

`create_test_app` takes the render lock internally, so the de-facto order is
**env then render**, and six tests follow it. Two take them in the opposite
order, which is a classic AB-BA cycle:

1. `crates/jcode-tui/src/tui/app/tests/scroll_copy_02/part_02.rs:86`
   `test_alt_shift_i_toggles_inline_images_and_persists` takes render, then env.

2. `crates/jcode-tui/src/tui/app/tests/smoothness_benchmark.rs:52`
   `smoothness_benchmark_simulated_streaming_turn_stays_within_budget` takes
   render, then env **indirectly** via `with_reasoning_current_home`.

The second is the one that makes this hard to spot: the env acquisition is
inside a helper, so the call site does not look like it touches a lock at all.

## A note on reading the thread dump

An earlier pass at this concluded a guard was being leaked from a dead thread,
because a gdb dump shows every worker blocked with *no thread holding either
lock*. That reading is wrong, and it is an easy trap: holding a mutex leaves no
stack frame, so a thread that owns one lock while blocking on the other is
indistinguishable from a thread that owns nothing. "No visible holder" is
exactly what an AB-BA cycle looks like.

Both locks also recover from poisoning, so a panicking test cannot leak them.

## Fix

Reducing the repro to just the two conflicting tests wedges within ~5 runs and
passes in 0.2s otherwise, which is why this never reproduced on demand.

Making both follow the env -> render order:

| | before | after |
|---|---|---|
| full suite, default 16 threads | never completes | ~50s, 5/5 runs |
| the 2-test repro at 2 threads | hangs within ~5 runs | 20/20 clean |
| serial results | 2255 passed / 15 failed | 2255 passed / 15 failed |

Identical pass/fail counts: this removes a hang, not a test.

The second test does not need the explicit render lock at all, since
`create_test_app` already takes it.

I also added a source-scanning test that asserts the ordering and names the
file, line and offending call, counting the three env-taking helpers
(`with_temp_jcode_home`, `with_reasoning_current_home`,
`with_ssh_remote_test_home`) as env acquisitions so the indirect form is
caught. It is static rather than runtime because the race is timing-dependent,
so a green run proves very little. Verified it fails by reintroducing the
original inversion.

## Availability

The fix is one commit on top of `master`, on a fork:
`YalmutairiAisc/jcode` @ `fix/windows-overload-and-tests`.

I could not open a PR ("An owner of this repository has limited the ability to
open a pull request to users that are collaborators on this repository"), so
filing here instead. Happy to send it as a standalone one-commit PR if you
would like to enable that, or feel free to just take the change directly, it is
a three-line reorder plus the guard test.

### LINKED ISSUE #1204: The exported `cargo` shim silently builds jcode when a Bash command cds into another Rust checkout
## Summary

The Bash tool exports a `cargo` shell function for the whole command, and `scripts/dev_cargo.sh` starts with an unconditional `cd "$repo_root"`. Together they mean that any cargo invocation made after a `cd` into an unrelated Rust checkout builds **jcode** instead of the project the agent is standing in.

It fails silently and with exit code 0, which is what makes it expensive.

Version: v0.83.73-dev (fdad9f09a), macOS/aarch64. Reproduces on master.

## Reproduction

From a jcode session whose working directory is the jcode source checkout, with any second Rust project on disk:

```bash
cd ~/some-other-rust-project && cargo metadata --no-deps --format-version 1 | jq -r '.packages[0].name'
```

### Observed

```
jcode
```

Exit code 0. `cargo test --lib` in the same position spends minutes compiling `jcode-tui-*` crates and reports jcode's test results as though they were the other project's.

### Expected

```
some-other-rust-project
```

`command cargo metadata` in the same directory already prints the correct name, which confirms the shim is the cause rather than PATH or mise.

## Mechanism

`wrap_repo_cargo_commands` (`crates/jcode-app-core/src/tool/bash.rs:48-69`) resolves the wrapper once from the session working directory and prepends:

```bash
export JCODE_DEV_CARGO_SCRIPT=/path/to/jcode/scripts/dev_cargo.sh
cargo() {
  if [[ "${JCODE_IN_DEV_CARGO:-0}" == "1" ]]; then
    command cargo "$@"
  else
    JCODE_IN_DEV_CARGO=1 "$JCODE_DEV_CARGO_SCRIPT" "$@"
  fi
}
export -f cargo
```

The function is defined for the entire command, so it survives any later `cd`. `scripts/dev_cargo.sh:5` then runs `cd "$repo_root"` with no check of where it was actually invoked from.

The wrapper is right to own cargo inside this repository: it applies memory-adaptive job sizing, linker policy, and the rust-actions.jsonl timing log. The defect is only that it also claims invocations made outside it.

## Impact

Any session that touches a second Rust project. That includes the dependency repos in this org: `mermaid-rs-renderer` and `agentgrep` are compiled into jcode, so an agent verifying a fix in either one gets a green cargo run that never compiled the crate under test. For `handterm`, whose open issue #4 is specifically about Darwin compilation, a hijacked `cargo check` would look exactly like "it builds fine on macOS".

It also inflates `target/debug`: two hijacked runs here added several GB to the jcode build cache.

## Proposed fix

Scope the wrapper to its own repository and hand everything else back to the real Cargo, in `scripts/dev_cargo.sh`:

```bash
repo_root=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)

export JCODE_IN_DEV_CARGO=1

case "$(pwd -P)" in
  "$repo_root" | "$repo_root"/*) cd "$repo_root" ;;
  *) exec cargo "$@" ;;
esac
```

`exec` runs a program rather than the exported shell function, so the handoff cannot recurse. Fixing it in the script rather than in `bash.rs` also covers cargo calls made from inside child scripts, which is the case the exported function exists to catch in the first place.

Branch, cherry-picks cleanly onto master: https://github.com/1jehuang/jcode/compare/master...ianalitis:jcode:jcode/dev-cargo-cwd-guard

## Validation

Ran the wrapper directly from three positions:

| Invocation directory | `cargo metadata` package before | after |
|---|---|---|
| jcode repo root | `jcode` | `jcode` |
| `jcode/crates/jcode-tui` | `jcode` | `jcode` |
| a sibling Rust checkout | `jcode` | that project |

`bash -n` clean; `shellcheck scripts/dev_cargo.sh` reports only the three pre-existing SC2034/SC1091 findings, none on the new lines.

I attempted a PR but `createPullRequest` returns FORBIDDEN for this account, so the compare link above is ready to cherry-pick.

### LINKED ISSUE #1206: Dragging a non-image file leaves a shell-escaped path in the composer that no file tool can open
## Summary

Dragging a non-image file into the TUI leaves the composer holding a *shell-escaped* path, which no file tool can open. Drop `/tmp/a b/c d.txt` and the composer gets `/tmp/a\ b/c\ d.txt`. That is correct shell quoting but it is not a path: `Read`/`Edit` resolve it literally, look for a directory component containing a backslash, and fail.

So the user drags a file in and the agent cannot read it. Images are unaffected (they become `[image N]` attachments); this only bites `.txt`/`.rs`/`.pdf`/etc.

## Environment

- macOS 15, aarch64
- jcode v0.84.0 and current `master`

## Reproduction

```console
$ mkdir -p "/tmp/jcode drop repro" && echo content > "/tmp/jcode drop repro/my notes.txt"
```

Drag that file from VS Code into the composer. It types:

```
/tmp/jcode\ drop\ repro/my\ notes.txt
```

```console
$ [ -f '/tmp/jcode\ drop\ repro/my\ notes.txt' ] && echo opens || echo "does not open"
does not open
$ [ -f '/tmp/jcode drop repro/my notes.txt' ] && echo opens
opens
```

## Root cause

`parse_dropped_paths` already does the hard part: it resolves shell escapes, `'...'`, `"..."`, and `file://` URLs, and verifies each token is an existing file. Those resolved paths are just never written back for non-image drops.

`promote_dropped_images` (`crates/jcode-tui/src/tui/app/input.rs`) returns early unless *every* dropped path is an image:

```rust
pub(super) fn promote_dropped_images(app: &mut App) -> bool {
    let Some(images) = dropped_image_files(&app.input) else {
        return false;   // <- a dropped .txt lands here, composer left escaped
    };
```

`dropped_image_files` maps over the paths and `?`-returns on the first non-image, so one text file discards the whole batch.

## Suggested fix

When the composer is *exactly* dropped paths, rewrite it to the verified paths `parse_dropped_paths` already produced. Since it confirmed each file exists, the rewrite is lossless. Guards that matter:

- only fire when the whole composer parses as dropped paths, so prose merely *containing* a path is never rewritten
- no-op when the path is already clean, so a typed path does not burn an undo entry
- keep multi-file drops quoted so they stay separable

Branch with a fix and tests, if useful as a reference: https://github.com/aphelion31/jcode/tree/fix/macos-clipboard-image-applescript

The tests assert `Path::is_file()` on the resulting composer rather than a string shape, so they pin the actual requirement (openable). Verified against real files whose directory *and* filename both contain spaces:

```
[PASS] R1 vscode shell-escaped   -> /tmp/.../my notes.txt
[PASS] R2 single-quoted          [PASS] R3 double-quoted
[PASS] R4 file:// URL            [PASS] R5 clean path untouched
[PASS] R6 prose not rewritten    [PASS] R7 image drop still attaches
[PASS] R8 multi-file separable   [PASS] R9 bracketed-paste route
```

Neutering the fix flips exactly R1-R4 to FAIL while the guard cases R5-R9 stay PASS, so the tests bind to the fix rather than passing incidentally. `cargo test --test e2e` is unaffected: 55 pass, and the 3 failures are pre-existing (`target/selfdev/jcode` missing in my checkout) and reproduce identically on the parent commit.

## Notes / non-goals

Deliberately not handled, since neither can be distinguished from a legitimate filename:

- `~/...` — the shell expands it before the terminal sends it, so a real drop never produces it
- percent-encoded paths *without* a `file://` scheme — indistinguishable from a filename that genuinely contains `%20`

Filing as an issue rather than a PR because `pull_request_creation_policy` is `collaborators_only`, so `POST /repos/1jehuang/jcode/pulls` returns `aphelion31 does not have the correct permissions to execute CreatePullRequest`. Happy to open one if that ever changes, or you're welcome to take the diff and rewrite it.

### LINKED ISSUE #1214: mcp_call arguments silently empty on schema-guided decoders (Kimi K3 via OpenRouter)
## Problem

On providers whose tool-call decoding is schema-guided (observed: Kimi K3 via OpenRouter), every `mcp_call` arrives at the MCP server with `arguments: {}` regardless of what the model emitted. Any MCP tool with required parameters then fails validation (`Required at pageId`, `missing field 'code'`), which makes the whole MCP surface unusable on that lane. Parameterless calls (e.g. chrome-devtools `list_pages`) work fine.

## Root cause

Three layers combine:

1. `mcp_call`'s `arguments` parameter is declared as a bare object schema (`{"type": "object", "description": ...}`), because the keys depend on the target MCP server's schema and cannot be declared at build time.
2. The OpenRouter dialect normalizer injects an empty `properties: {}` map into bare object schemas (`require_properties_on_objects`, #446), so the wire schema becomes `{"type": "object", "properties": {}}`.
3. Kimi K3's schema-guided decoding constrains generated tool-call JSON to the declared properties — zero of them — and silently drops whatever the model put into `arguments`.

Evidence from the session journal: `"mcp_call","input":{"arguments":{},"server":"chrome-devtools","tool":"select_page"}` — recorded at tool input, so the loss happens at/before argument parsing, not in dispatch. Even a JSON string payload arrived as `{}`. Meanwhile `batch` (whose sub-call items declare `additionalProperties: true`) carries arbitrary nested objects on the same lane without loss, which pointed at the schema declaration as the differentiator.

## Proposed fix

Reference implementation (with regression test) is on my fork: branch `fix/mcp-call-open-world-args`, commit `tobias-hui/jcode@557078e2b` (I could not open a PR — `CreatePullRequest` is denied for my account on this repo — so filing this issue instead, per the contributing guide).

The change declares `additionalProperties: true` on `arguments`, matching the existing open-world handling of `batch`'s sub-call payloads. The marker survives every dialect normalizer and tells schema-guided decoders that arbitrary keys are allowed. For OpenAI strict mode the tool was, and remains, ineligible (object with no declared properties), so behavior there is unchanged.

The included regression test (`mcp_call_arguments_stay_open_world_after_dialect_normalization`) builds the tool directly and asserts the marker survives OPENROUTER / OPENAI / ANTHROPIC dialect normalization.

## Verification

- New test plus the existing `tool::mcp` suite: 13 passed, 0 failed (`cargo test -p jcode-app-core --lib tool::mcp`).
- Live end-to-end on the Kimi K3/OpenRouter lane after the fix: `select_page {"pageId": 1}` reached chrome-devtools-mcp and switched the selected page; before the fix the identical call failed validation every time.

Reproduces with any MCP server/tool that has required parameters (e.g. node_repl `js` → `missing field 'code'`).

1jehuang:
## Summary

Ten scoped fixes from the newest-50 issue triage, with one issue-referencing commit each. Six are new fixes from this pass and four were already present in the local branch base and have been revalidated and explicitly published here. No contributor branch was imported. Unrelated local telemetry, embedding, provider, and workflow changes are excluded.

| Issue | Change | Observed focused verification |
|---|---|---|
| #1184 | Do not rearm terminal focus reporting from focus events | 4 TUI focus tests passed |
| #1186 | Refresh local-client global skills before initial History, after the SSH early return | 17 skill and 47 SSH tests passed, with overlap. Corrected negative control fails without the fix |
| #1187 | Reap notification children without blocking the caller | 4 process tests passed across both crates. 2 subprocess entry points are intentionally ignored by the outer runner |
| #1188 | Preserve explicit sponsor opt-outs across save/load | 8 sponsor configuration tests passed |
| #1189 | Create promoted background output artifact before returning | 20 background tests and 1 real Bash timeout-promotion test passed. Running output may still be empty, this is not live streaming |
| #1194 | Allow scheduled NotifySession through subscription-free dispatch to the existing handler | 5 runner, 2 busy/idle notification, and 80 protocol tests passed. Real isolated socket regression failed before the fix and passed six post-fix executions |
| #1201 | Acquire environment lock before rendering lock in both affected tests | 3 source-order regressions and both actual TUI tests passed. Original sites rejected by negative control |
| #1204 | Preserve Cargo caller working directory outside the Jcode checkout | 3 baseline failures became 5 passing real Cargo workspace tests. Existing job-sizing tests and shell syntax passed |
| #1206 | Normalize completed non-image file drops, not partial key streams | 7 targeted tests, 26 drop-filter tests and 15 input tests passed, with overlap. Initial 4 failures became green |
| #1214 | Preserve arbitrary MCP arguments in provider-facing schemas | 14 MCP tests passed, including schema dialect and strict-mode regressions |

## Review and validation limits

- **CI follow-through (September 10):** the published head has an actual [failed CI workflow-validation run](https://github.com/1jehuang/jcode/actions/runs/34513206384), not merely pending tests. GitHub reports `(Line: 16, Col: 1): env is already defined`. The [Windows smoke](https://github.com/1jehuang/jcode/actions/runs/34513208752) and [iOS workflow](https://github.com/1jehuang/jcode/actions/runs/34513207566) also failed before jobs. These failures are absent from the PR check rollup, where only Require Linked Issue passes. Draft status is not the cause, and marking ready would not repair invalid YAML. The separate #1191 / #1193 workflow repair remains the prerequisite. No branch was imported and no CI failure was bypassed.

- **Draft, not merge-ready.** This does not bypass failed or pending required checks. Existing draft #1193 owns the separate #1191 workflow/formatting repair. Its last checked result is 8 passed and 3 failed: Quality Guardrails, Ubuntu Build & Test, and macOS Build & Test. Those changes are not duplicated here. Resolve that baseline CI work and run this branch's full checks before merging.
- **Exact published snapshot now verified:** archived head `0e84b661a1c437b2376398a9be4409b5def6664b` was compiled independently through the host-wide Cargo gate. All 17 focused Rust cohorts listed above passed on the resulting executables, as did all five Cargo workspace tests, job-sizing tests, shell syntax and three source-order regressions. All 1,886 tracked archive files were checked against their published Git blob hashes after testing. This supersedes the initial working-branch-only validation limitation. It is not a full workspace CI run.
- `selfdev test` and the coordinated TUI build were blocked before compilation by a pre-existing missing telemetry Git blob. The repository's host-wide Cargo gate was used for local fallback. Remote build host was unavailable, and early high-memory compiler attempts were terminated by earlyoom. Successful focused reruns are the evidence above, not those failed attempts.
- The first CLI build timed out at ten minutes. A warmed continuation then completed in 2m59s. The exact CLI and a real isolated TUI/daemon both reported `v0.84.0-dev (0e84b661a)`. The acceptance checks below used that binary with private home/runtime/socket and no inherited API credentials. No shared daemon was reloaded or stopped.
- Full workspace CI, live Kimi decoding, native macOS/Windows runtime and real SSH host validation remain unverified. The real isolated TUI reached provider selection and reported every provider as not configured. Therefore no live-provider scheduler response is claimed. The scheduler regression uses real sockets with a fake provider, and notification lifecycle tests use instrumented notifier processes. These are narrower evidence, not replacements for the unavailable native/provider acceptance paths.
- No issue has been closed and no PR has been merged or released by this triage pass.

## Observed public-interface behavior

- **#1187:** A small caller linked the exact published public notification library. The API returned in 648µs, real `/usr/bin/notify-send` sent the expected notification over the desktop D-Bus, and process tracing confirmed its successful `wait4` before the still-alive parent exited. No notifier substitute was used in this check. This is public-library/desktop/OS lifecycle acceptance, not a TUI completion or macOS check.

- **#1184:** In native Ghostty 1.3.1-arch2 on Linux, three focus-away/back cycles reached the exact TUI as actual `focus_gained` events. The next 5-second focused interval measured 0.20% of one CPU core for Jcode, 1.20% for Ghostty and seven new draw samples. No focus storm recurred in this bounded run. Original focus was restored and the private window/daemon cleaned up. This is not a cross-version or long-duration benchmark.

- **#1214:** With deferred MCP enabled, the exact daemon catalog exposed `mcp_call.arguments.additionalProperties=true`. Actual MCP connect/call/disconnect against a controlled local JSON-RPC stdio peer preserved arbitrary keys, nested values, Unicode and escaped quotes in the wire request and result. Owned processes were cleaned up. This is real CLI/daemon/stdio integration with a fixture peer, not live Kimi decoding.

- **#1186:** In the real TUI, `/triage-v` plus Tab became `/triage-visible` before populated History. Captured frames showed the completed skill in the input area with no frame anomalies.
- **#1206:** Enter on a shell-escaped filename produced the unescaped path in real TUI history and its rendered transcript. Passing that exact path to the real daemon read tool returned the file contents.
- **#1188:** Normal `/account default-model ...` and `clear` commands saved the requested preferences while on-disk `[sponsors] enabled = false` survived both operations.
- **#1189:** The real daemon Bash tool promoted a 2-second command after 100ms. `bg status` said Running, `bg output` succeeded with an empty artifact, then completion returned exit 0 and the expected command output.
- **#1204:** Real Cargo metadata after an exported Bash shim, cd and child shell selected the foreign project rather than Jcode. Own workspace/member, symlink and error behavior also passed.

## Whole-result rerun after requirement mapping

Re-executed all 17 mapped Rust cohorts, five real Cargo workspace tests, job-sizing/shell checks and three lock-order checks against the complete published head. All passed again, and all 1,886 tracked source hashes still matched afterward. The 28-row requirement ledger records each observed output and its evidence boundary, including the still-failing baseline CI prerequisite rather than calling it passed.

Repeated the actual native TUI skill/frame, escaped-path/transcript/read, preference persistence, Bash/background lifecycle and deferred MCP stdio workflows successfully. Repeated native Ghostty focus checks observed actual focus events followed by 0.00% Jcode CPU, 0.40% Ghostty CPU and zero new draw samples over five seconds. Repeated real desktop notification delivery returned from the public API in 7,764µs and reaped the real child before parent exit. Private windows/processes were cleaned up. Test-driver corrections for onboarding dismissal and the correct TOML field were made before the successful complete rerun, without production edits.

Fresh GitHub reads over all 50 initially selected issues reconfirmed one bucket each, 30 newly owner-answered issues, 23 newly labeled issues, all 33 signed comments and ten fix links. All issues remain open. Native Windows/macOS, real SSH and live-provider/Kimi checks remain unverified, and full CI is still blocked. This rerun does not make the PR merge-ready.

Fixes #1184
Fixes #1186
Fixes #1187
Fixes #1188
Fixes #1189
Fixes #1194
Fixes #1201
Fixes #1204
Fixes #1206
Fixes #1214

---


1jehuang:
Scoped review complete. Integrating the ten contribution commits as ancestry-preserving `16b6badd0`, not the unrelated old branch diff. Conflict resolution retains the newer tmux extended-key lifecycle tests and existing MCP collision/identity policy.

Exact combined executable `7d25a5646` passed real isolated TUI/daemon checks: skill Tab completion before History, three focus cycles without focus-report rearming, escaped file-drop submission and actual file read, bracketed paste, account preference writes preserving sponsor opt-out, and foreground Bash timeout promotion with readable empty output while Running and final output after completion.

Fresh named regressions also pass for notification child reaping/missing executable (both paths), real subscribed-client scheduled delivery, idle/busy NotifySession, dynamic MCP schemas/strict-mode exclusion, five real Cargo-cwd cases, all seven file-drop cases, local/SSH skill isolation, tmux terminal recovery, and both actual lock-sensitive tests plus lock-order negative controls. These are Linux/local-fixture checks, not native Ghostty/macOS/Windows or live Kimi acceptance. Full hosted CI has unrelated baseline formatting failures documented in #1193.

PR #1237, opened 2026-09-13
fix: address three reproducible issues from September triage

### LINKED ISSUE #1221: Built-in deepseek route drops images for deepseek-flash, which accepts image input
## Environment

- jcode v0.84.0 (57d587899), macOS
- Built-in `deepseek` provider, `model = "deepseek-flash"` (`provider_key = "openai-compatible:deepseek"` in the session file)

## What happens

Reading an image in a session on the built-in deepseek route gives back the clamp marker instead of the image:

```
Image: /Users/…/codigo.png (1.4 KB)
Dimensions: 420x200
Image sent to model for vision analysis.

[Image omitted: this provider/model does not support image input; media_type=image/png]
```

The model never sees the pixels. `supports_image_input()` is false for this route, so the outbound image clamp replaces the block with that text.

## Why that is wrong now

`deepseek-flash` does accept image input. Same PNG, same endpoint, sent straight to the API as a base64 `image_url` block:

```
model: deepseek-flash
finish_reason: stop
content: "V6480G"
```

`V6480G` was a random 6-character code drawn inside a rectangle on the image, and it is not present anywhere in the prompt text. So the built-in profile is conservatively text-only (the policy from #755) for a model that stopped being text-only.

## Workaround

A named profile on the same endpoint works, because the OpenAI-compatible runtime defaults to reporting image support:

```toml
[providers.ds-vision]
type = "openai-compatible"
base_url = "https://api.deepseek.com"
auth = "bearer"
api_key_env = "DEEPSEEK_API_KEY"
default_model = "deepseek-flash"

[[providers.ds-vision.models]]
id = "deepseek-flash"
```

`jcode --provider-profile ds-vision` reads the same image correctly. So the behavior is reachable today, just not on the built-in route.

## Suggested fix

Mark `deepseek-flash` vision-capable in the built-in `deepseek` profile, or honour `models[].input` for it the way #772 asks.

On the aliases: `jcode model list -p deepseek` lists `deepseek-v4-flash`, `deepseek-flash` and `deepseek-v4-pro`; the API also accepts `deepseek-v4-flash` and `deepseek-v4-flash-vision-exp`, and all of them resolve server-side to `deepseek-flash`. I would only mark flash. In an earlier check on this machine `deepseek-v4-pro` did not answer on the image payload, so it should not be flipped without re-testing.

### LINKED ISSUE #1223: Agents systematically treat `timeout`  of bash tool as seconds (Anthropic OAuth)
For maybe a month now, I see agents executing bash commands, then immediately telling themselves that the timeout was actually supposed to be in milliseconds when their invocation dies some 100ms later.
Agent investigation and suggested fixes below:

## Summary

Agents repeatedly pass second-sized values to Bash `timeout`, which is in
milliseconds, silently killing commands in a fraction of a second. On the
Anthropic OAuth (subscription) path the advertised schema is a hand-written
copy that documents nothing, so the model has **no way** to learn the unit.

## Evidence from 330 local sessions

1529 Bash calls passed an explicit `timeout`. **54 were smaller than 5000ms**,
i.e. almost certainly intended as seconds:

```
300ms  x14     600ms  x10     900ms  x10
240ms  x4      400ms  x4      1800ms x2
...down to 15ms
```

The intent is unmistakable from the commands themselves — these are not
commands anyone expected to finish in 0.3s:

```
timeout=300ms   git clone --depth 50 https://github.com/1jehuang/jcode.git ...
timeout=40ms    sleep 20; aerospace list-apps; ...
```

That last one is decisive: `sleep 20` given a 40ms budget. The agent clearly
believed it was granting 40 seconds.

The healthy values cluster at 900000/300000/600000/120000 — agents that *do*
know the unit (or discover it after failure) use it correctly. The failures are
a distinct population, not noise. **~3.5% of all timed Bash calls are
self-sabotaging.**

## Why agents cannot learn the unit

The unit is inconsistent across this codebase's own tools:

| tool | param | unit |
|---|---|---|
| `bash` | `timeout` | **milliseconds** |
| `webfetch` | `timeout` | **seconds** |
| `debug_socket` | `timeout` | **seconds** |
| `bg` | `graceful_timeout_ms`, `max_wait_seconds` | stated in name |

Identical bare name, 1000x apart. `bash` is the outlier that omits the unit
suffix its own sibling `bg` already uses.

And on the OAuth path the model is told literally this:

```json
"timeout": {"type": "integer"}
```

No description. `format_tools()` in `jcode-provider-anthropic/src/lib.rs`
serves hand-written schemas for 8 builtins, bypassing the registry. The
registry copy *does* say MILLISECONDS — but OAuth users never receive it.

`bash.rs::timeout_message()` even carries a comment saying agents
"frequently mistake" this, and appends a corrective hint. Both the warning
and the hint live only on the path OAuth users never see, and both fire
**after** the command has already been killed.

## Secondary: the same curated copy also drops parameters

Registry advertises 8 Bash properties; the OAuth copy advertises 4.
Missing: `intent`, `notify`, `wake`, `stall_wake_seconds`.

This is the **third** occurrence of one root cause, not a new problem:

- **#706** — curated `ScheduleWakeup` advertised `delaySeconds/reason/prompt`
  while the real handler required `task`; every call failed.
- **#722** — curated `Bash` omitted `justification`, the destructive gate's
  escape hatch.
- **this issue** — curated `Bash` omits the `timeout` unit plus four
  parameters.

Commit 3e8cf17c7 fixed the first two and named the pattern exactly:
"Hand-maintained schemas drift from the real tools they stand in for, and the
failure is invisible until a model calls the tool."

Notably, #706 and #722 were fixed *differently*, and only one of the two
fixes held:

- `ScheduleWakeup` had its hand-written entry **deleted**, letting the real
  schema forward under the OAuth name — described in that commit as keeping
  them "in sync by construction". It has not drifted since.
- `Bash` was patched by hand-adding the single missing property, leaving the
  drift mechanism in place. It drifted again, which is this issue.

## Suggested fixes, in order of value

1. **Put the unit in the name.** Accept `timeout_ms` (serde alias, so existing
   `timeout` callers keep working). A name that carries its unit is the only
   fix that survives schema drift, prompt changes, and new providers.
2. **Document the unit on the curated OAuth schema** so the existing warning
   actually reaches subscription users.
3. **Apply the #706 treatment to Bash: delete the hand-written schema and
   forward the registry one under the OAuth name.** This is not a new
   proposal — it is the fix already chosen, shipped, and proven for
   `ScheduleWakeup` in 3e8cf17c7, and it is the reason that tool stopped
   drifting while Bash did not.

## Environment

jcode v0.84.23-dev, base commit 90c94adb0, macOS 26.6.2 arm64, Anthropic
OAuth path.

## Reproducing the measurement

```bash
python3 - <<'PY'
import json, glob, os, collections
rows = []
for f in glob.glob(os.path.expanduser("~/.jcode/sessions/*.json")):
    try:
        d = json.load(open(f))
    except Exception:
        continue
    def walk(o):
        if isinstance(o, dict):
            if o.get("name") in ("bash", "Bash") and isinstance(o.get("input"), dict):
                t = o["input"].get("timeout")
                if isinstance(t, int):
                    rows.append((t, str(o["input"].get("command", ""))[:70]))
            for v in o.values():
                walk(v)
        elif isinstance(o, list):
            for v in o:
                walk(v)
    walk(d)
susp = [r for r in rows if r[0] <= 5000]
print("bash calls with explicit timeout:", len(rows))
print("suspicious (<=5000ms):", len(susp))
print(collections.Counter(t for t, _ in susp).most_common())
PY
```

### LINKED ISSUE #1234: api-bridge: empty sessions/ directory panics stored_session_ids (chunk size must be non-zero)
## Observed

`jcode api-bridge` panics on `list_sessions` when `$JCODE_HOME/sessions` exists and is empty:

```
thread 'tokio-runtime-worker' panicked at crates/jcode-harness-api-server/src/translate.rs:1707:38:
chunk size must be non-zero
```

The API socket then resets. A subsequent `create_session` on the same client fails with a broken pipe / harness connection closed.

Measured on `jcode v0.81.5 (f46f9c354)` with an isolated `JCODE_HOME`.

## Controls

Isolated `JCODE_HOME`, `jcode api-bridge --api-socket ... --no-update`, handshake `hello` with `min_version: 1, max_version: 1`, then `list_sessions`:

| home | result |
|---|---|
| no `sessions/` directory | `{"ev":"sessions","sessions":[]}` then `create_session` succeeds |
| empty `sessions/` directory | panic at `translate.rs:1707:38`, socket reset, `create_session` BrokenPipe |

## Ask

`stored_session_ids` should treat a zero-length directory listing as an empty session list instead of dividing by a zero chunk count.

This is independently the root of Cotal-AI/Cotal#1293: a managed seat whose private home has an empty `sessions/` directory cannot start.

1jehuang:
## Summary

Three small fixes from the newest-50 issue triage, with one issue-referencing commit each. Unrelated local changes and local-only base commits are excluded. No contributor branch was imported.

| Issue | Fix | Observed verification |
|---|---|---|
| #1234 | Return an empty session list before zero-sized chunking | Regression reproduced the reported panic before the fix. All 113 bridge tests passed. The exact published bridge build passed real isolated socket `hello → list_sessions → ping → create_session` checks for missing, empty, and filtered-only directories. |
| #1223 | Forward the registered Bash definition through Anthropic OAuth instead of replacing it with a drifting curated schema | Regression failed before the fix. All 21 Anthropic crate tests passed. Tests preserve millisecond documentation, execution options, OAuth naming, one advertisement, and cache marking. No timeout alias or unit reinterpretation was added. |
| #1221 | Permit image input only for the three reported DeepSeek Flash aliases on the DeepSeek profile | Two new regressions failed before the fix. All 13 focused tests passed, including actual captured HTTP request payloads to a local fixture. Pro, unknown models, ZAI, and explicit per-model overrides remain conservative or retain their configured behavior. |

## Validation boundaries

- Exact published head `e2b9d7a1aeae7f0e951af6328ce1e58c13d27cff` passed all three affected crate suites: **268 passed, 0 failed, 1 ignored** (113 bridge, 21 Anthropic, 134 compatible-runtime). Its complete TUI/CLI selfdev build also passed. All 1,898 tracked source blobs still matched the published Git tree after testing. This is not a full workspace CI run.
- Repeated the real-socket check using the newly built exact-snapshot **CLI for both `serve` and `api-bridge`**. All three directory layouts passed `hello → list_sessions → ping → create_session` on one connection. Private homes/sockets and owned-process cleanup were used. No shared daemon was restarted/reloaded and no model inference was requested. The archive build reports `v0.84.0-dev (unknown)` because it has no `.git`, so identity was established by source-blob verification and binary SHA-256 `5fc1d981ca943f52ff22e2cadc9277b12381f38b7108cb02ef299843149c692a`, not its version string.
- DeepSeek server-side vision support relies on the reporter's direct-API evidence. No live DeepSeek or Anthropic OAuth inference was performed. The tests verify local capability/schema handling and emitted image payloads, not vendor behavior or future model choices.
- **CI is actually failing, not merely pending.** The [CI workflow](https://github.com/1jehuang/jcode/actions/runs/34731918900), [Windows smoke](https://github.com/1jehuang/jcode/actions/runs/34731918266), and [iOS workflow](https://github.com/1jehuang/jcode/actions/runs/34731918593) failed before jobs. The default branch retains duplicate top-level `env` maps tracked by #1191 / existing PR #1193. Only Require Linked Issue passes in the PR check rollup. This remains draft until that baseline workflow repair and full checks are resolved. Existing PR #1218 owns ten other triaged fixes and is not imported here.
- Existing formatting defects outside the new regression region remain in `translate_tests.rs`. Patch whitespace checks pass. No whole-workspace formatting pass is claimed.
- No issues were closed, and no PRs were merged.

Fixes #1234
Fixes #1223
Fixes #1221

---


1jehuang:
All three scoped fixes were already present on current master. Integrating ancestry-only commit `057a0c8d9` records the original contribution without changing or regressing newer code.

Fresh exact-candidate public bridge acceptance passed for missing, empty, and filtered session directories. Each returned hello_ok, sessions=[], pong, then attached with a nonempty session id. Checks ran in private homes/sockets and an isolated network namespace.

Fresh provider checks passed for OAuth Bash schema/timeout/justification preservation and DeepSeek image capability/explicit overrides, including captured local HTTP requests proving pixels are sent only on allowed routes. No live Anthropic or DeepSeek inference is claimed.

PR #1246, opened 2026-09-13
fix: report indeterminate source update checks (fixes #912)

### LINKED ISSUE #912: Update check reports "up to date" forever when the source branch has no upstream
### Summary

A source checkout whose current branch has **no upstream** silently skips every update check and reports **"up to date"**, no matter how far behind it is. My client sat 450 commits behind (v0.65.1-dev vs v0.75.5) for ten days showing a green, up-to-date status.

### Root cause

`check_for_updates()` in `src/cli/hot_exec.rs:245` decides "is an update available" with:

```rust
let behind = ProcessCommand::new("git")
    .args(["rev-list", "--count", "HEAD..@{u}"])
    .current_dir(&repo_dir)
    .output()
    .ok()?;

if behind.status.success() {
    let count: u32 = ...;
    Some(count > 0)
} else {
    None          // <-- no upstream lands here
}
```

When the branch has no tracking ref, `@{u}` fails with `fatal: no upstream configured for branch '<name>'`, so `behind.status.success()` is false and the function returns `None`.

The caller in `src/cli/startup.rs:331` treats `None` exactly like "no update":

```rust
if let Some(update_available) = hot_exec::check_for_updates()
    && update_available
{
    ...
} else {
    Bus::global().publish(BusEvent::UpdateStatus(UpdateStatus::UpToDate));   // <-- reached on None
}
```

`None` (could not determine) and `Some(false)` (definitely current) are collapsed into the same confident `UpToDate`.

The divergence guard added right below is aimed at exactly this developer case:

```rust
if hot_exec::local_commits_ahead_of_upstream() == Some(true) {
    logging::info("Auto-update skipped: local commits are ahead of upstream (diverged). ...");
```

but it uses `@{u}` too (`hot_exec.rs:284`), so it also returns `None` and never fires. A branch with no upstream therefore skips the update **and** skips the log line explaining why, leaving no trace anywhere in the UI or the logs.

This is easy to hit: `git checkout -b my-fix` in `~/.jcode/source/jcode` and never push. That is the normal shape of a local self-dev fix.

### Reproduce

```bash
cd ~/.jcode/source/jcode
git checkout -b local-fix        # no -u, so no upstream
git commit --allow-empty -m wip
# start jcode; status shows "up to date" regardless of how far master has moved
git rev-list --count HEAD..@{u}  # fatal: no upstream configured
```

### Suggested fix

Distinguish "cannot determine" from "current". Minimal version: when `@{u}` does not resolve, fall back to the remote default branch (`origin/HEAD`, or `origin/master`) for the behind-count, and surface an `UpdateStatus` that says the checkout is untracked rather than up to date. At the very least, log it, the way the diverged case does, so the state is discoverable:

```rust
// in check_for_updates, when @{u} fails
let fallback = ProcessCommand::new("git")
    .args(["rev-list", "--count", "HEAD..origin/HEAD"])
    ...
```

A `None` from `check_for_updates()` should probably never publish `UpToDate`.

### Environment

- jcode v0.65.1-dev (38f935ad2), source checkout, Windows 11 x86_64
- branch `fix/idle-animation-blank-v065`, no upstream, 450 commits behind `origin/master` (v0.75.5)
- `.git/FETCH_HEAD` was fresh, so the fetch ran; only the comparison failed

1jehuang:
## Summary
Closes #912.

- Stop reporting `UpToDate` when a source update comparison returns `None`, including an untracked branch, unavailable repository, or failed fetch.
- Publish the existing `UpdateStatus::Error` and log the diagnostic. Known-current and update-available paths keep their behavior.
- Do not select a fallback branch, change tracking configuration, or alter automatic pull/install policy.

The contributor's existing fallback proposal remains a separate policy option. This patch takes the conservative reporting-only path requested in the issue.

## Verification
- Negative control restoring the original `None -> UpToDate` behavior: exactly two regressions failed, while known-true/false controls passed.
- On final commit `dd717ce25`, `scripts/dev_cargo.sh test --profile selfdev -p jcode --lib cli::startup::tests`: **15 passed**. Coverage includes a real temporary Git repository with no upstream, an up-to-date local tracking branch, and a behind branch, without network access.
- On final commit, `scripts/dev_cargo.sh build --profile selfdev -p jcode --bin jcode` passed. Startup reports `v0.84.43-dev (dd717ce25)`.
- The same patched source was also exercised in a real PTY/TUI with a separate HOME, runtime directory, socket, and source fixture having no upstream or remote. Its rendered output showed `Update failed: Source update check failed: unable to compare the source checkout with...`, and the log recorded that diagnostic. No inference was requested. JSON frame capture was unavailable, so the actual rendered PTY output was inspected instead. The shared daemon was not reloaded.
- Changed-file formatting and `git diff --check` passed. Independent read-only review found no blocking concerns.

## Test prerequisite and limitations
A separate five-line test-only commit initializes the newly required optional `ModelRoute.usage` field in existing CLI fixtures. Without it, the unchanged CLI test target fails to compile before any regression can run. No production model-route behavior changed.

This draft does not claim a green full suite or native Windows/macOS acceptance. Hosted [CI run 34789313248](https://github.com/1jehuang/jcode/actions/runs/34789313248) failed before creating jobs, as did the Windows smoke and iOS workflows. These inherited duplicate top-level `env` mappings are tracked in #1191 / PR #1193. The linked-issue check passed, but that is not build/test validation. Please rerun hosted checks once that baseline is repaired. The fix was prepared in an isolated branch, preserving the active dirty worktree.

---


1jehuang:
Scoped review complete. Integrating original contribution ancestry as `0f3bc4dfa`, including the optional ModelRoute fixture fields.

All fifteen CLI startup tests pass, including the four exact source_update_check tests and real local Git upstream states. The exact combined executable's real private tmux TUI, with a repository lacking remote/upstream, displayed `Update failed: Source update check failed`, rather than claiming it is up to date. True/false outcomes remain covered as update-available/up-to-date. The acceptance test used only its fixture repository and did not update the live installation or shared daemon.

PR #1251, opened 2026-09-14
fix(tui): request extended keys from tmux so Shift+Enter arrives

### LINKED ISSUE #1250: Shift+Enter requires extended-keys always because Jcode does not opt in under tmux
## Bug

Shift+Enter submits instead of inserting a newline inside tmux with the documented `extended-keys on` configuration. Switching only `extended-keys` to `always` makes it work.

## Environment

- Jcode v0.84.0 (57d587899)
- macOS arm64
- Kitty
- tmux 3.7c

## Reproduction

Configure tmux:

```tmux
set -s extended-keys on
set -s extended-keys-format csi-u
set -as terminal-features ',xterm*:extkeys'
```

Launch Jcode inside tmux and press Shift+Enter. It submits rather than inserting a newline. Claude CLI and Neovim handle the chord correctly in the same environment. Kitty is explicitly configured to send `ESC[13;2u` for Shift+Enter, so the outer terminal already distinguishes it.

Running `tmux set -s extended-keys always` fixes Jcode, but forces extended-key reporting for other applications too.

## Investigation

`enable_keyboard_enhancement()` in `crates/jcode-tui/src/tui/mod.rs` only pushes Kitty keyboard flags. It never requests tmux's xterm `modifyOtherKeys` mode (`CSI > 4 ; 2 m`). With `extended-keys on`, tmux needs that application-side opt-in. Sending CSI-u from Kitty to tmux does not bypass tmux's per-pane output encoding.

An isolated tmux 3.7c server with `extended-keys on` and `extended-keys-format csi-u` confirms the behavior using a raw-mode pane and `send-keys S-Enter`:

| Application output | Input received |
| --- | --- |
| Kitty push only (`ESC[>7u`) | `\r` |
| `ESC[>4;2m` + Kitty push | `ESC[13;2u` |
| `ESC[>4;2m` + Kitty set | `ESC[13;2u` |
| Reset with `ESC[>4;0m` + Kitty pop | `\r` |

## Expected fix

Request extended keys when running under tmux, reassert the request when reapplying terminal modes, and reset it on cleanup. Preserve existing Kitty behavior outside tmux. No `always` setting or extra terminal key mappings should be required.

rafi:
Closes #1250

## Problem

Inside tmux with the settings `/terminal-setup` writes, Shift+Enter submits instead of inserting a newline:

```tmux
set -s extended-keys on
set -s extended-keys-format csi-u
set -as terminal-features ',xterm*:extkeys'
```

Only `extended-keys always` worked, which forces extended-key reporting on every other application in the server.

`enable_keyboard_enhancement()` pushed Kitty keyboard flags but never requested xterm `modifyOtherKeys` mode. tmux does not treat a Kitty push as a request for extended keys, so with `extended-keys on` it collapsed Shift+Enter to a bare CR.

Configuring the outer terminal does not help. Mapping Shift+Enter to `ESC[13;2u` in Kitty controls only what Kitty sends to tmux. tmux decodes keys and re-encodes them per pane, so the opt-in has to come from the application.

## Change

Send `ESC[>4;2m` when `TMUX` is set, reassert it in `reapply_terminal_modes_to` alongside the existing Kitty set form, and reset it with `ESC[>4;0m` on cleanup. Behavior outside tmux is byte-for-byte unchanged.

The sequences are extracted into small writer-taking functions so the whole lifecycle is testable without a terminal.

## Validation

**Protocol level.** An isolated tmux 3.7c server with `extended-keys on` and `extended-keys-format csi-u`, sending `S-Enter` to a raw-mode pane:

| Application output | Key received |
| --- | --- |
| Kitty push only (`ESC[>7u`) | `\r` |
| `ESC[>4;2m` + Kitty push | `ESC[13;2u` |
| `ESC[>4;2m` + Kitty set | `ESC[13;2u` |
| After `ESC[>4;0m` + Kitty pop | `\r` |

**End to end**, running the built binary in a tmux pane with `extended-keys on`, typing `alpha`, pressing Shift+Enter, then typing `beta`:

- Without the opt-in (same binary, `TMUX` unset): the prompt submitted and the model answered `alpha`.
- With the opt-in: `alpha` and `beta` sit on two lines of one input box, nothing submitted.

**Tests.** Four unit tests in `tui::terminal_mode_tests` cover the tmux lifecycle, the unchanged non-tmux lifecycle, reapplication without a second Kitty stack push, and no keyboard sequences when enhancement is disabled. The tmux test fails without the fix.

`cargo test -p jcode-tui --lib -- terminal_mode_tests terminal_setup shift_enter bare_carriage_return keyboard`: 41 passed.

Not from this change: `import_continue_reaches_ready_quality_first_openai_model` and `recent_project_review_falls_back_cleanly_when_no_repo_is_known` also fail on the unmodified base commit, and several `ssh_remote_*` tests hang in my environment.

## Notes

- Gated on the `TMUX` environment variable, matching how `terminal_setup::diagnose` already detects tmux.
- `ESC[>4;2m` is sent unconditionally under tmux rather than after a capability probe, since tmux consumes it and a terminal that does not understand it ignores it.
- Users keep `extended-keys on`; `always` is no longer needed.
- `docs/SHIFT_ENTER.md` records why terminal-side mappings alone cannot fix this.


rafi:
Two further checks since opening:

**Cleanup with the real binary.** Previously only the enable path was proven end to end. Running the built binary in a tmux pane with `extended-keys on`, quitting it, then handing the same pane to a raw-mode probe: `S-Enter` delivers `\r`. The `ESC[>4;0m` reset lands, so the mode does not leak to whatever runs next in that pane.

**Against a real user config** rather than a minimal `-f /dev/null` server. Using a config that sets `extended-keys on`, `extended-keys-format csi-u` and `terminal-features ,xterm*:extkeys`, Shift+Enter puts `alpha` and `beta` on two lines of one input box with nothing submitted.

`cargo clippy -p jcode-tui --lib` reports no warnings for the changed file.

rafi:
Re-ran the full check set against the exact pushed commit (`e0092557`, clean tree, byte-identical rebuild), and correcting one detail in the original description while doing so.

**Correction.** My first end-to-end runs classified submission by whether `alpha` and `beta` landed on the same captured line. That proxy is wrong: after a submission the transcript can still render them on separate lines, so a re-run reported the no-opt-in case as *not* submitting. Captured screens show what actually happens without the opt-in: `sending context…` and a new prompt ordinal `2…`, while the opt-in case stays in the editor at `1>`. Re-classifying on submission activity rather than line layout:

| check | expected | observed |
| --- | --- | --- |
| Shift+Enter with opt-in (`TMUX` set) | newline | newline |
| Shift+Enter without opt-in (`TMUX` unset) | submits | submits |
| pane after jcode exits | plain `\r` | plain `\r` |

The conclusion is unchanged, and the underlying claim is now backed by a discriminator I checked against real screens rather than an assumed one.

**Mutation check.** Deleting just the two `write_all` opt-in/reset lines makes `tmux_keyboard_lifecycle_requests_and_resets_extended_keys` fail and the other four pass; restoring them turns it green. The test fails for the intended reason.

Also on this commit: `terminal_mode_tests` 5/5, the wider keyboard/terminal-setup selection 40/40, and clippy clean on the changed file.

1jehuang:
Merged into master. Validated the built Jcode TUI inside an attached tmux 3.7c client PTY: physical CSI-u Shift+Enter produces a two-line composer, does not submit a message, and normal `/quit` resets translation back to CR. The pre-change binary reproduced the erroneous submission. Combined terminal-mode, KV-cache and Shift+Enter tests also passed.

Issue #518, opened 2026-07-20
Antigravity provider: tool calls fail with 400 "missing thought_signature in functionCall parts"


sigmaboysigmaboyyy:
Using the `antigravity` provider, plain chat works fine, but any request that triggers a tool call (e.g. bash) fails with:

X Antigravity generateContent failed (HTTP 400 Bad Request): {
  "error": {
    "code": 400,
    "message": "Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call [default_api:bash], position 7. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.",
    "status": "INVALID_ARGUMENT"
  }
}

Steps to reproduce:
1. jcode login --provider antigravity
2. Select antigravity as the active provider (default)
3. Send a prompt that triggers a tool call, e.g. "what time is it on my pc"
4. Request fails with the above 400 error; auto-poke stops with fallback prompt to switch to Copilot

Expected: tool calls should succeed the same way they do with the `gemini` (oauth) provider.

Environment:
- jcode v0.53.0, Windows 10/11
- Provider: antigravity (OAuth)
- Model: default

This looks related to Gemini's newer requirement that function call responses include a thought_signature field  — possibly the antigravity

sigmaboysigmaboyyy:
Per Google's own docs (https://ai.google.dev/gemini-api/docs/thought-signatures), when no real thought_signature is available, the backend accepts a placeholder value in that field instead of omitting it — either "skip_thought_signature_validator" or "context_engineering_is_the_way_to_go" bypasses validation cleanly. Other Antigravity/Gemini-CLI integrations (e.g. opencode-antigravity-auth) hit this same class of bug and fixed it by injecting that placeholder whenever a captured signature is missing/invalid, rather than trying to guarantee every call is always signed.

1jehuang:
Please retest on current v0.54.4 or current master and attach a redacted request/response trace for the first failing tool turn. Include the resolved Antigravity model, whether the response contained `thoughtSignature`, and whether the failure occurred on the initial call or a replayed follow-up.

--- *— Jcode agent (automated triage), on behalf of @1jehuang*

1jehuang:
Addressed in `e045ff0b3` (alongside #482).

[...]

PR #1256, opened 2026-09-14
fix(gemini): preserve thought_signature on stored tool calls; retry burst 429s; enable compaction

adminwat:
Three fixes on top of v0.84.0 for Gemini "works on turn 1, fails on turn 2":

1. **thought_signature dropped when storing tool calls** (`turn_streaming_mpsc.rs`, `turn.rs`). The signature was captured from `StreamEvent::ToolUseSignature` but the persisted `ContentBlock::ToolUse` was hand-built with `thought_signature: None` in the mpsc/TUI paths (only `turn_loops.rs` carried it). Turn 2 replayed an unsigned history and Gemini 3.x returned HTTP 400 `Function call is missing a thought_signature`. Construction now goes through `ToolCall::to_tool_use_block()`. Three regression tests in `jcode-message-types`; falsified by restoring `None` (2 of 3 fail).
2. **Burst 429 from Code Assist killed the turn.** Gemini was the only runtime with no 429 retry. Now retries up to 5x with backoff.
3. **Gemini/Cursor/Antigravity had no compaction path** (`supports_compaction() == false` skipped even the emergency hard-compact).

Verified on a daemon-hosted mpsc session: two sequential tool calls on `gemini-2.5-pro` (server-upgraded to `gemini-3.1-pro-preview`), persisted history carries `thought_signature` on every tool call, zero `stream_error`.

Refs #518 (shared thought-signature persistence). This reference does not claim live-provider acceptance or close that issue.


1jehuang:
Merged into master via reviewed integration 8567dfaf9. Preserved thought signatures and provider/session identity fixes, but corrected review findings: config no longer exports sticky process environment, auth state is keyed by route/project, compaction caps refresh safely, and HTTP retries classify transient versus persistent quota errors and respect bounded Retry-After. Cursor compaction remains disabled because its summary path truncates input. Combined config/compaction tests, 146 provider/message/schema tests, local HTTP 429->503->200 replay, actual cache replacement, and a two-turn agent signature/session-ID test passed. No live Google account acceptance was possible, so this does not certify upstream credentials/quota behavior.

Issue #972, opened 2026-08-16
Follow-ups from #941: two dispatch_name call sites, silent MCP name collisions, and complete_simple still dropping in-band errors


aayushbaluni:
Three loose ends from #941 (which closed #936 and #937). All three are still present on `master` at `c4cdc67`. Filing as one issue since both source threads are now closed.

Branches, since PR creation is restricted to collaborators here:

- `fix/mcp-dispatch-name-call-sites` — [compare](https://github.com/1jehuang/jcode/compare/master...aayushbaluni:jcode:fix/mcp-dispatch-name-call-sites) (items 1 and 2)
- `fix/complete-simple-inband-errors` — [compare](https://github.com/1jehuang/jcode/compare/master...aayushbaluni:jcode:fix/complete-simple-inband-errors) (item 3)

---

## 1. Two call sites still build the dispatch name by hand

`dispatch_name` normalizes `-` to `_` in **both** segments, which the test in #941 pins (`hyphenated-server` -> `mcp__hyphenated_server__`). Two places still use the old `format!` and so no longer match what is registered.

**`src/cli/commands.rs:2549`** — the cold-cache MCP startup wait:

```rust
let prefix = format!("mcp__{}__", server);
names.iter().any(|name| name.starts_with(&prefix))
```

For a server named `my-server`, registered tools are `mcp__my_server__*`, so `covered` never becomes true. The loop cannot detect readiness, burns the full budget, and then logs

> jcode run: timed out waiting for cold-cache MCP server(s); their tools may be missing from this run

for tools that did in fact register. Silent latency plus a misleading warning.

**`crates/jcode-app-core/src/server/debug_command_exec.rs:327`** — `mcp:call:<server>:<tool>`:

```rust
let tool_name = format!("mcp__{}__{}", server, tool);
```

Misses for a hyphen in either segment, which is the exact case #941 set out to fix.

Both are one-line swaps to `dispatch_name`. I also added a test pinning `dispatch_name(server, "")` against the full name, so a prefix scan cannot drift from the registered form again.

## 2. Dispatch-name collisions are silent

Because normalization is many-to-one, two distinct remote tools can collapse onto one dispatch name:

- one server exposing both `query-docs` and `query_docs`
- two servers named `a-b` and `a_b`

`Registry::register` is `tools.insert(name, tool)`, so the second registration silently replaces the first. The losing tool stays listed by `mcp list` but is not the tool that runs — a wrong-tool-executes failure with no diagnostic.

[...]

PR #1259, opened 2026-09-14
fix(mcp): prevent normalized tool collisions

yansigit:
## Summary
Prevent normalized MCP tool names from colliding at provider boundaries while preserving the original server-qualified dispatch identity.

## Changes
- Generate deterministic unique aliases when normalization produces duplicates.
- Apply the same mapping to eager, cached, and live registration paths.
- Keep original names available for dispatch back to the MCP registry.
- Cover collision ordering, aliases, and dispatch behavior with focused tests.

## Validation
- `git diff --check upstream/master...HEAD`
- Rebasing onto current `upstream/master` completed successfully.
- Focused jcode-base MCP tests passed.
- The broader app-core suite still has unrelated baseline failures in communication, session-search, tool-schema, and selfdev tests.

Refs #972


1jehuang:
Merged into master with review follow-ups in bcb6686b9 and real stdio coverage in c01aab726. Beyond collision-resistant names, integration now reconciles aliases by original MCP identity, removes stale live entries while retaining failed-server cached proxies, keeps legacy/alias deny policy consistent across eager/deferred routes and membership changes, honors legacy allowlists, and disconnects exact servers. Manual connect/list output uses actual aliases. The combined candidate passed 58 base MCP and 27 app-core MCP tests plus actual subprocess stdio coverage for eager/cache/reversed/partial discovery, dispatch identity, and management lifecycle.

Issue #1274, opened 2026-09-15
Conifer static fallback lacks context limits for 25 models and fails catalog completeness test


ianalitis:
## Summary

The Conifer static fallback added in #1089 contains 25 model IDs for which the existing context-limit completeness test cannot find a known limit. This makes `every_static_profile_model_has_a_known_context_limit` fail before any provider login or network request.

## Revision and reproduction

Upstream `master` was `74e7a4be54ae1db736bbd0e32ae1e6f59b475044` when checked on 2026-09-15. Run:

```sh
scripts/dev_cargo.sh test --offline --locked -p jcode-base --lib every_static_profile_model_has_a_known_context_limit -- --test-threads=1
```

Observed on macOS/aarch64: 0 passed, 1 failed, exit 101. The test reports:

```text
static profile models without a known context limit (would fall back to the generic default)
```

All 25 reported entries belong to Conifer:

```text
grok-4.6, grok-4.5, grok-4.3
seed-2.0-pro, seed-2.0-code, seed-2.0-mini
step-3.7-flash, step-3.7-flash-novita
hy3, hy3-tencent, hy3-novita
ling-3.0-flash, inkling, inkling-small
nemotron-3-ultra, nemotron-3-ultra-together
nemotron-3-super-120b, nemotron-3.5-lightning
mistral-large-latest, mistral-medium-latest, mistral-small-latest
command-a-cohere, llama-4-maverick, llama-4-scout, gemma-4-31b
```

## Attribution and limits of this report

This was encountered during a local structural cleanup on a dirty checkout at that revision, not a pristine whole-workspace run. I repeated the failing test after restoring the exact pre-extraction `provider_catalog.rs`, and it failed with the identical 25 IDs. That original file is byte-identical to the file fetched from upstream master (SHA256 `df1b92150a42f0b53076a12508c388ea67bd41fceac22bfb9deda6971e429864`). The Conifer static catalog shape test passed. No model limits or assertions were changed, and no vendor requests or credentials were used.

## Expected

The static fallback and context-limit resolver should agree for every advertised model, using verified provider/model metadata. Mutable `*-latest` aliases may need a deliberate provider-specific policy rather than guessing fixed limits. The completeness assertion should remain enabled. This report does not establish a live provider failure, only a deterministic static metadata/test gap.

Reported with Jcode assistance on behalf of the fork maintainer.

ianalitis:
Follow-up with authoritative, unauthenticated metadata for the missing-context report.

[...]

PR #1276, opened 2026-09-15
fix: honor ambient hot-reload and platform-aware badge tests

### LINKED ISSUE #1269: [ambient] Enabling ambient in config.toml has no effect until restart, while `ambient:status` reports `enabled: true`
### Summary

Ambient is enabled with `[ambient] enabled = true` in `config.toml`. If the daemon is already running, the change has no effect: no cycle ever runs and no wake is scheduled. `jcode ambient status` reports `enabled: true` in the same period, because it reads live config while the ambient loop's gate does not.

### Why this should work without a restart

Config hot-reload is an existing, intentional contract in this codebase, not something ambient is expected to opt out of:

- `config()` (`crates/jcode-base/src/config.rs:263`) fingerprints the config file (path, mtime, size) and reloads it when the fingerprint changes, on a 500ms check interval.
- On reload it calls `notify_config_reloaded()`, which bumps a monotonic generation counter documented as existing so callers "can poll this cheaply and re-derive their snapshot when the generation changes, giving instant hot-reload of config edits without a restart" (`config.rs:429`).
- Upstream already tests this: `global_config_cache_reloads_after_manual_file_edit` (`crates/jcode-base/src/config_tests.rs:708`) asserts that a manual on-disk edit is observed, with a comment noting the fingerprint is deliberately built "so the metadata fingerprint notices the manual edit."

### Root cause

`run_loop` snapshots the value once, before entering the loop, then gates on that frozen copy forever:

```rust
// crates/jcode-app-core/src/ambient/runner.rs:551
let ambient_enabled = config().ambient.enabled;

loop {
    // ...
    // runner.rs:592
    let ambient_allowed =
        ambient_enabled && !matches!(state.status, AmbientStatus::Disabled);
```

Meanwhile the observable status reads live config on every call:

```rust
// runner.rs:252 — the `ambient:status` payload
"enabled": config().ambient.enabled,
```

and `AmbientManager::is_enabled()` (`manager.rs:31`) also reads live. So after an edit the status endpoint reports the new value while the loop gate keeps the old one. The two disagree for the remaining lifetime of the process.

The rest of the loop is inconsistent with the frozen gate, which suggests this is an oversight rather than intended:

- `runner.rs:575` reads `config().ambient` for scheduler config
- `runner.rs:895` reads `config().ambient.visible` live on every cycle
- `runner.rs:252` reads live

Only `enabled` is frozen. If a restart were the intended contract for ambient, `visible` would be frozen as well, since both live in the same `[ambient]` section.

### Reproduction (no daemon required)

Against `74e7a4be5`, using the isolation pattern already used in `config_tests.rs` (`JCODE_HOME` + `Config::invalidate_cache()`):

```rust
let _guard = crate::storage::lock_test_env();
let temp = tempfile::tempdir().expect("tempdir");
let _home = EnvVarGuard::set_path("JCODE_HOME", temp.path());

let path = Config::path().expect("config path");
std::fs::create_dir_all(path.parent().expect("parent")).expect("mkdir");

// Daemon boots with ambient disabled.
std::fs::write(&path, "[ambient]\nenabled = false\n").expect("write");
Config::invalidate_cache();

// runner.rs:551 — the loop's boot-time snapshot.
let boot_snapshot = config().ambient.enabled;
assert!(!boot_snapshot);

// User edits config.toml while the daemon keeps running.
std::fs::write(&path, "[ambient]\nenabled = true\n# edited\n").expect("edit");
Config::invalidate_cache();

// runner.rs:252 — what `ambient:status` reports.
assert!(config().ambient.enabled);   // passes: the edit is observed
assert_eq!(boot_snapshot, config().ambient.enabled);  // fails
```

Observed output:

```
live config after edit      : true     <- status endpoint reports enabled=true
loop snapshot after edit    : false    <- gate stays false
```

The status endpoint and the loop gate disagree at the same instant, and they stay that way until restart.

### Observed vs expected

**Observed:** after enabling ambient in a running daemon's config, `jcode ambient status` shows `enabled: true` with `status: idle`, `total_cycles: 0`, and no `state.json` is written. No cycle runs. Only a restart activates ambient.

**Expected:** ambient becomes active within the ~500ms config reload interval, consistent with what `ambient:status` already reports and with how other config-derived state behaves.

### Impact

Enabling ambient silently does nothing for anyone who edits the config while a daemon is running, which is the normal path for a user who already has `jcode serve` or a TUI session up. Because the status endpoint confirms `enabled: true`, the failure looks like ambient is broken rather than like a required restart.

### Suggested fix

Evaluate per iteration instead of once:

```rust
let ambient_allowed =
    config().ambient.enabled && !matches!(state.status, AmbientStatus::Disabled);
```

Two related points, offered as questions rather than assertions:

1. `runner.rs:556` gates the one-shot reply-poller spawns on the same startup variable. Since those are spawn-once, keying them to startup config may well be intended; I have not assumed either way.
2. The scheduler is built once at `runner.rs:575` from `config().ambient` (interval, `pause_on_active_session`). Whether interval changes are meant to apply mid-loop is a separate question from the enable gate, and I would not fold it into this fix.

### Availability

Patch plus a regression test on my fork: https://github.com/marko-k0/jcode/compare/master...fix/ambient-enable-and-stuck-running

The test asserts on the gate decision itself (extracted into a small helper, since `run_loop` cannot easily be asserted against). On pristine upstream it fails with:

```
gate must observe the live config edit, not the boot-time snapshot (false);
otherwise ambient:status reports enabled=true while the loop never runs a cycle
```

Happy to open a PR if that's useful.

### LINKED ISSUE #1275: macOS: copy-badge margin test assumes the non-macOS Alt-label width
## Summary

On macOS/aarch64, `test_copy_badge_reserves_right_margin_for_info_widgets` fails in an isolated, exact, single-threaded run because its expected margin assumes the three-column `Alt` label. The renderer defaults to the one-column `⌥` label on macOS.

This is a platform-sensitive test expectation, not evidence that the rendered macOS badge should become wider. Unlike the parallel-state failures tracked in #596, this failure occurs with one test running.

## Reproduction

Source HEAD: `74e7a4be54ae1db736bbd0e32ae1e6f59b475044`.
Environment: macOS, aarch64.

```sh
scripts/dev_cargo.sh test --offline --locked -p jcode-tui --lib \
  tui::ui::tests::basic::test_copy_badge_reserves_right_margin_for_info_widgets \
  -- --exact --test-threads=1
```

Observed:

```text
running 1 test
test tui::ui::tests::basic::test_copy_badge_reserves_right_margin_for_info_widgets ... FAILED

  left: 18
 right: 16

test result: FAILED. 0 passed; 1 failed; 0 ignored
```

The run was against a checkout containing other in-progress changes, not an isolated clean worktree. However, both files responsible for this test and width calculation were byte-identical to HEAD, and the failure was observed before applying the proposed structural cleanup:

- `crates/jcode-tui/src/tui/ui_tests/basic/input_layout.rs`
- `crates/jcode-tui/src/tui/ui_viewport.rs`

## Source-level explanation

- `ui_viewport.rs:5-8` selects `⌥` for macOS and `Alt` elsewhere. A configured display label can also override that default.
- `copy_badge_shortcut_width` measures the rendered shortcut using Unicode display width.
- `reserve_copy_badge_margins` subtracts that measured width from the affected row.
- `input_layout.rs:372` hardcodes a remaining width of `16` from an initial width of `30`.

The default shortcut including its separator is 12 columns with `⌥`, versus 14 with `Alt`, explaining `30 - 12 = 18` rather than the hardcoded `30 - 14 = 16`.

A neighboring fixture, `test_copy_badge_truncates_full_width_line_before_appending_shortcut`, also appends a literal `[Alt] [⇧] [A]` after calculating reservation through the platform/config-aware helper. That neighboring test has not yet been executed in this investigation, so this is a source-level follow-up, not a claimed second reproduced failure.

## Expected behavior and narrow repair

Preserve the platform/config-aware production rendering. Make the fixtures compare reserved space with the independently constructed rendered badge width, while retaining assertions that unaffected rows stay unchanged and appended shortcuts fit the viewport. Do not merely replace `16` with `18`, skip the test, or weaken it to a nonzero-width assertion.

The existing default-label and trimmed-override tests should remain part of verification. Any fix should be validated with the exact regression and adjacent badge/layout tests on macOS, while retaining the non-macOS `Alt` contract. No fix or full-suite success is claimed here.

1jehuang:
## Summary

Fixes #1269. Fixes #1275.

Two independently reviewable, narrow fixes from issue triage:

- **#1269:** Re-read `ambient.enabled` at the loop gate instead of retaining the startup value. Preserve explicit `AmbientStatus::Disabled`, startup-only reply pollers, and the existing scheduler/wake policy. Add a disk-config regression covering false → true → false and the explicit-stop override.
- **#1275:** Construct expected shortcut text with the configured/platform modifier label and Unicode display width. Keep unaffected-row and exact-fit assertions, and fix the neighboring fixture's literal `Alt`. No production badge rendering changes.

Commits: 2200140f5 (#1269), 180e95db0 (#1275).

## Verification

Performed on Linux x86_64 through the coordinated self-dev test/build queue:

- Reproduced the original badge assertion with an isolated `JCODE_HOME` and `[display] copy_badge_alt_label = "⌥"`: **18 != 16** before the fixture fix.
- `scripts/dev_cargo.sh test --offline --locked -p jcode-tui --lib copy_badge -- --test-threads=1`: **16 passed** with default `Alt`, and **16 passed** with configured `⌥`, including platform-default and trimmed-override tests.
- Ran each of the two changed badge fixtures separately with `--exact --test-threads=1` under each label: **4 exact invocations passed** on the final formatted tree.
- `scripts/dev_cargo.sh test --offline --locked -p jcode-app-core --lib ambient::runner:: -- --test-threads=1`: **6 passed**, including the new toggle regression and real running-loop hot-reload regression and existing disabled-ambient scheduled-task coverage.
- `scripts/dev_cargo.sh build --profile selfdev -p jcode --bin jcode`: **passed** via `selfdev build`.
- `git diff --check` and `rustfmt --check --edition 2024` on all three changed files: **passed**.

## Measured behavior follow-up

Additional regression coverage in c515991f2 exercises the actual `run_loop`, not just the extracted gate. It starts disabled, edits `config.toml` to enable ambient without invalidating the cache, and observes the loop enter `Paused` for an active user session after nudging it. This exercises file fingerprint reload and the production gate without model inference.

Restoring the old startup-snapshot gate made this exact test fail after its 3-second observation deadline. Restoring the fix made it pass, along with all six ambient runner tests. This establishes a behavioral improvement at the real loop boundary, while not claiming a full CLI/daemon or native macOS acceptance run.

## Limits and non-goals

- Native macOS execution is still needed. Linux `⌥` coverage exercises the reported width difference, not the macOS OS/runtime.
- Ambient edits take effect at the next loop wake. This does not promise a 500ms wake or interrupt an already-running cycle. Reply-poller and scheduler hot-reconfiguration remain out of scope.
- The shared daemon was not reloaded. Runtime assertions above are from the compiled test targets, not an end-to-end live-daemon hot-reload run.
- No full-workspace green claim: independently reproduced the pre-existing Conifer completeness failure in #1274 (the same 25 missing IDs), which requires a separate metadata/alias policy decision.
- Existing contributor fixes for #1266/#1264/#1252/#1250 remain in their own PRs. No duplicate implementation or unrelated branch is included.

## Hosted CI follow-up

This PR is draft pending hosted checks and native macOS validation. The [CI run](https://github.com/1jehuang/jcode/actions/runs/35024226179), [Windows smoke](https://github.com/1jehuang/jcode/actions/runs/35024227957), and [iOS workflow](https://github.com/1jehuang/jcode/actions/runs/35024226999) all failed before creating any jobs. `gh run view` reports a likely workflow-file issue, and these unchanged base workflows each contain duplicate top-level `env` maps. The existing repair is tracked by #1191 / PR #1193 and is not duplicated here. Require Linked Issue passed. Local test/build results above remain valid, but this is not green hosted CI.

---
*Jcode agent (automated triage), on behalf of @1jehuang*


1jehuang:
All three scoped commits were already present byte-for-byte on current master. Integrating ancestry-only `7d25a5646` records the contribution without replacing newer code.

Fresh exact executable acceptance with a private daemon/local provider fixture observed disabled startup with zero requests, config=true on next wake producing one cycle, config=false stopping further requests/cycles, and explicit ambient stop still winning after config=true. The real-loop, config-gate, and schedule-service regressions pass.

Both full copy-badge cohorts passed: sixteen rendering tests with default config and sixteen with Unicode Alt label `⌥`, including right-margin reservation and full-width-line truncation. No native macOS run is claimed.

PR #1285, opened 2026-09-16
fix(tui): consume focus/paste events and prevent panic-induced exits

KooshaPari:
## Summary

Three interacting TUI bugs allowed the foreground TUI to exit (or the
backgrounded server to record the session as `Crashed`) without restoring
the terminal. This PR fixes all three by adding explicit event arms,
making the panic hook synchronously restore the terminal, and switching
the runtime to `UnhandledPanic::Task`.

This is the upstream contribution derived from the `KooshaPari/kcode`
fork (k1.1.1), where the same fixes were committed as
`c6d1200b39ff3ead6f28334bac81a32addce063c` after passing the local CI
loop. The kcode fork additionally includes a Windows-only compile
workaround in `crates/jcode-herdr/src/socket.rs` and a version bump in
`Cargo.toml`; those are intentionally omitted from this upstream PR
because they are not behaviour changes for `1jehuang/jcode`.

## Fix A — consume focus and paste events in turn loops

The three `tokio::select!` loops in
`crates/jcode-tui/src/tui/app/turn.rs` (the API-wait, streaming, and
tool-execution loops) used a catch-all `_ => {}` arm that silently
swallowed `Event::FocusGained`, `Event::FocusLost`, and on the
tool-execution path `Event::Paste`. The primary event handlers in
`local::apply_terminal_event` and `remote::apply_terminal_event`
already consume these events, but during a long turn the secondary
loops fell behind — a focus-in byte leaked into the catch-all, focus
state stayed stale, and animated UI elements (status spinner interval,
copy-mode badges, feature unlocks gated on focus) misbehaved for the
remainder of the turn.

Each of the three `select!` sites now has explicit arms:

```rust
Some(Ok(Event::FocusGained)) => {
    crate::tui::reapply_configured_terminal_modes();
    self.note_client_focus(true);
    let _ = self.set_client_focused(true);
}
Some(Ok(Event::FocusLost)) => {
    self.set_client_focused(false);
}
Some(Ok(Event::Paste(text))) => {
    self.handle_paste(text);
    status_spinner_renderer.draw_full(self, terminal)?;
}
Some(Err(error)) => {
    crate::logging::warn(&format!(
        "tui: transient event-stream error: {error}"
    ));
}
```

The new `Some(Err(error))` arm replaces the previous silent drop. A
transient `crossterm::EventStream` error (typically a ConPTY sync
hiccup during a focus event) was previously invisible in logs; it now
warns and continues rather than propagating out of the tokio task,
which would have shut the runtime down.

## Fix B — synchronous panic hook with rate guard

`install_panic_hook` in `src/cli/terminal.rs` previously took the
default hook and ran, leaving the terminal in raw mode with focus
events and bracketed paste still enabled. The hook now:

1. Disables raw mode, focus events, bracketed paste, the alternate
   screen, and re-shows the cursor (best-effort, errors swallowed so
   the panic handler cannot itself panic) **before** delegating to the
   previous hook.
2. Appends `PANIC at <RFC3339>: <payload>\n  at <file>:<line>:<col>\n`
   to `<session>.panic.log` next to the session file so a crashed
   backgrounded server can be diagnosed after the fact.
3. Rate-guards itself: if five panics arrive within a 12 s window we
   `exit(101)` after the next one, so a runaway panic loop is
   distinguishable from a hot loop.

The cleanup runs on every panic, not only on the terminating one,
because a single panic in a tokio task no longer implies runtime
shutdown once Fix D lands.

## Fix D — `UnhandledPanic::Task` in `src/main.rs`

`tokio::runtime::Builder` is now configured with
`unhandled_panic(UnhandledPanic::Task)` so a panic in a spawned task
is killed in isolation rather than tearing the whole runtime down.
This is cfg-gated on `tokio_unstable` (Tokio 1.49); build with
`RUSTFLAGS=--cfg tokio_unstable` to enable. Without that cfg the call
is a no-op and the prior default (`Shutdown`) is preserved, so
existing release builds are unaffected.

## Files changed

```
crates/jcode-tui/src/tui/app/turn.rs | 69 ++++++++++++++++++++++++++++++++
src/cli/terminal.rs                  | 76 ++++++++++++++++++++++++++++++++++++
src/main.rs                          | 28 +++++++++++--
3 files changed, 170 insertions(+), 3 deletions(-)
```

## Validation performed locally

- `cargo check --bin jcode` — clean
- `cargo clippy --bin jcode` — clean
- `cargo build --bin jcode` — clean
- `cargo test -p jcode-tui --lib terminal_setup` — 24/24 passing
- `cargo test -p jcode-tui --lib focus` — 32/32 passing
- `cargo test -p jcode-tui --lib paste` — 22/22 passing

## Reproduction

1. Launch `jcode` and click out of the terminal window for > 30 s
   during an API stream. Without this PR, focus state stays stale
   and the spinner / copy-mode badge never reactivate when you click
   back in.
2. Trigger any panic inside a tokio task (e.g. an unwrap on a
   transient API error). Without this PR, the runtime shuts down,
   the session is recorded as `Crashed`, and the terminal stays in
   raw mode with bracketed paste / focus events still enabled.

With this PR, both scenarios restore the terminal cleanly, persist
the panic detail to `<session>.panic.log`, and keep the foreground
TUI responsive after a focus-in.

## Related

- Investigation report: `jcode-tui-investigation-2026-09-16.md`
- Change report: `jcode-fixes-applied-2026-09-16.md`
- kcode fork commit: `KooshaPari/kcode@c6d1200b3`


1jehuang:
Rebased onto current master and merged the turn-loop focus/error handling (switched to `reapply_configured_terminal_modes_after_focus`, which master introduced to avoid a Ghostty focus-report loop). I dropped the panic-hook and tokio runtime changes: the `.panic.log` path was relative to cwd, and tearing down the terminal in the global hook would also fire for panics caught by `catch_unwind` while the TUI keeps running. Happy to look at those separately.

Issue #1116, opened 2026-08-31
Switching Claude between OAuth and API key leaks JCODE_RUNTIME_PROVIDER process-wide, mislabeling and breaking restore for other sessions


wwillchen:
## Summary

Switching a Claude session between **OAuth** and **API key** in `/model` mutates the process-global env var `JCODE_RUNTIME_PROVIDER`. Every *other* session hosted by the same server process then derives its `provider_key` from that leaked value instead of from its own active provider. Sessions running an OpenAI model get persisted as `provider_key: "claude-api"`, and on resume the restore path rebuilds the model request as `claude-api:gpt-5.6-sol`, which hard-fails.

This is the cross-session sibling of #1068 (which fixed the *presentation* of the active route). The credential switch is applied correctly for the switching session, but the switch has global side effects on unrelated sessions.

## Environment

- jcode `v0.81.4 (896f866eb)`, macOS aarch64
- Both Anthropic auth modes configured: `claude` OAuth + `ANTHROPIC_API_KEY`
- Multiple sessions served by one `jcode serve` process (normal multi-tab usage)

## Reproduction

1. Have both Claude OAuth and an Anthropic API key configured.
2. In session A, `/model` -> pick a Claude model on the **API key** route (`claude-api:claude-fable-5`).
3. In the same server process, open session B and select an **OpenAI** model (e.g. `gpt-5.6-sol`).
4. Inspect `~/.jcode/sessions/<session-B>.json`.
5. Resume session B.

**Expected:** session B persists `provider_key: "openai"` and resumes on `gpt-5.6-sol`.

**Actual:** session B persists `provider_key: "claude-api"` with `model: "gpt-5.6-sol"`, and resume logs:

```text
[ERROR] Failed to restore session model 'gpt-5.6-sol' via 'claude-api:gpt-5.6-sol':
Model gpt-5.6-sol not supported by Anthropic provider
(retried after reloading auth from disk: Model gpt-5.6-sol not supported by Anthropic provider)
```

The session silently falls back to whatever model the provider defaults to, so the user's model selection is lost on resume.

## Observed evidence

Live process state right after toggling Claude OAuth -> API key, while the active provider is OpenAI:

```text
JCODE_RUNTIME_PROVIDER=claude-api
JCODE_ACTIVE_PROVIDER=openai
```

Route toggling works at the request level (both directions are accepted):

```text
event=model_picker_select api_method=claude-api  route_api_method=claude-api  spec=claude-api:claude-fable-5
[...]

Issue #691, opened 2026-07-31
fix(provider): OpenRouterProvider::name() should return profile_id for named OpenAI-compatible profiles


hoa1982rp:
## Problem

`OpenRouterProvider::name()` hardcodes the return value to `"openrouter"` regardless of the profile. When a named OpenAI-compatible profile is configured (e.g., `[providers.my-gateway]` with `type = "open-ai-compatible"`), the provider is still identified as `"openrouter"` instead of using the actual profile ID.

This breaks several integration points for named profiles:

1. **Auth refresh identity** — the auth lifecycle cannot distinguish between different OpenAI-compatible profiles because they all report the same name.
2. **Model catalog cache namespace** — cached catalogs from different profiles collide under the same `"openrouter"` key.
3. **Provider identification in logs** — all named profiles log as `provider=openrouter` instead of their actual profile name, making debugging multi-gateway setups difficult.
4. **Profile-specific behavior gates** — checks like `profile_id.as_deref() != Some("fpt")` exist elsewhere in the same impl, confirming that profile-aware logic is expected; `name()` should follow the same pattern.

## Affected code

**File:** `crates/jcode-provider-openrouter-runtime/src/openrouter_provider_impl.rs`  
**Line:** ~324

**Current (v0.64.2):**
```rust
fn name(&self) -> &str {
    "openrouter"
}
```

## Proposed fix

The `profile_id` field already exists on the struct and is used extensively throughout the same implementation for profile-specific logic. Return it when available:

```rust
fn name(&self) -> &str {
    self.profile_id.as_deref().unwrap_or("openrouter")
}
```

This preserves backward compatibility — when `profile_id` is `None` (vanilla OpenRouter), it still returns `"openrouter"`. When a named profile is active, it returns the actual profile name.

## Additional note

I noticed the upstream repo contains a companion test `named_openai_compatible_provider_reports_profile_name` (in `openrouter_tests.rs` as of a local hotfix on v0.32.0) that validates this exact behavior — but it was never merged upstream. That test should be included alongside this fix:

```rust
#[test]
fn named_openai_compatible_provider_reports_profile_name() {
    let _lock = ENV_LOCK.lock();
    let _key = EnvVarGuard::set("TEST_NAMED_COMPAT_KEY", "test-key");

    let profile = crate::config::NamedProviderConfig {
        base_url: "https://llm.example.com/v1".to_string(),
        api_key_env: Some("TEST_NAMED_COMPAT_KEY".to_string()),
[...]

Issue #380, opened 2026-06-15
Provider/model runtime identity can diverge across config reload, session restore, UI, model refresh, and reload recovery


wtry:
Provider/model runtime identity can diverge across config reload, session restore, UI, model refresh, and reload recovery
Summary
Jcode currently appears to maintain provider/model state across multiple independent sources without one authoritative runtime identity.

Observed state sources include:

• active Agent.provider
• connection-level/shared provider
• session.provider_key
• session.model
• session.route_api_method
• config defaults from config.toml
• provider memory caches
• persisted provider model catalog caches
• TUI remote model catalog cache
• remote history/bootstrap payloads
• reload recovery continuation state
• UI header / welcome / model picker state

These sources can diverge. When they do, different code paths may use different provider/model identities, causing confusing behavior such as:

• config edits not affecting the active runtime until killall jcode
• header/welcome showing the wrong provider
• session restore using stale provider/model/route metadata
• model refresh targeting one provider while the UI snapshot comes from another
• model picker showing stale catalog entries
• reload recovery continuing with an unintended provider/model
• custom OpenAI-compatible profiles being displayed or routed incorrectly

This is especially visible with custom OpenAI-compatible providers, where the provider slot, profile id, display name, API base, auth env key, route API method, selected model, and catalog namespace are all separate concepts.

──────────────────────────────────────────────────────────────────

Confirmed code observations
1. Config reload does not rebuild the active provider runtime
config() reloads the config cache when the config fingerprint changes.

Startup currently registers lightweight reload listeners such as:

• AuthStatus::invalidate_cache
• Bus::global().publish_models_updated()

There does not appear to be a config reload listener that rebuilds or reconciles the current Agent.provider.

So editing config.toml may update future config reads while leaving the active provider runtime unchanged.

2. Model refresh uses a connection-level provider, while the update event is generated from the agent
In client_lifecycle.rs:

┌─ rust 
│ Request::RefreshModels { id } => {
│     handle_refresh_models(id, &provider, &agent, &client_event_tx).await;
│ }
└─

In handle_refresh_models:

┌─ rust 
│ let refresh_future = provider_clone.refresh_model_catalog();
└─

After refresh succeeds, the UI event is generated from the agent:

┌─ rust 
│ let event = available_models_updated_event(&agent_clone).await;
└─

So if the connection-level provider and the active agent provider diverge, refresh may target provider A while the UI snapshot is generated from provider B.

3. History fallback can display provider/model from persisted session metadata
[...]

PR #1288, opened 2026-09-17
fix(provider): name the active profile in logs, History and /usage, not the openrouter slot

### LINKED ISSUE #1286: Direct OpenAI-compatible providers are reported as `openrouter` in logs and in `/usage` spend
## Summary

The `openrouter` provider is a multiplexing slot: it serves the public OpenRouter aggregator **and** every direct OpenAI-compatible profile (`deepseek`, `nvidia-nim`, `zai`, ...). The slot's transport name is `openrouter` / `OpenRouter`, while the profile-aware name is `display_name()` (`runtime_display_name()`).

Several code paths use the transport name instead of the profile-aware name, so sessions that talk to DeepSeek through a direct profile are reported as `openrouter`:

1. **Log prefix** – every line of a DeepSeek session is tagged `prv:openrouter|mod:deepseek`.
2. **`/usage` spend ledger** – spend is attributed to an `openrouter` bucket on hosts where OpenRouter is not configured and was never used.
3. **KV cache telemetry** – `provider=OpenRouter` while `model=deepseek-flash`.

## Environment

- jcode `v0.84.0 (57d587899)` — release build, Ubuntu 20.04, x86_64
- Provider: DeepSeek via API key (`~/.config/jcode/deepseek.env`), `openrouter` reports `not_configured`
- Also reproduced on `master` (`5f33d6239`); the offending lines are unchanged there

## Steps to reproduce

```sh
mkdir /tmp/jcode-repro
cat > /tmp/jcode-repro/config.toml <<'EOF'
[[providers.deepseek.models]]
id = "deepseek-flash"
context_window = 1048576
EOF

# API key: export DEEPSEEK_API_KEY=... or place it at
# $JCODE_HOME/config/jcode/deepseek.env
JCODE_HOME=/tmp/jcode-repro jcode run -p deepseek -m deepseek-flash "hi"
grep 'prv:' /tmp/jcode-repro/logs/*.log
```

## Actual

A single run of the release binary in a clean `JCODE_HOME`:

```
[2026-09-17 10:18:33.787] [INFO] [ses:session_macaque_1789|prv:openrouter|mod:deepseek] TURN_CANCEL_REGISTERED ...
[2026-09-17 10:18:33.788] [INFO] [ses:session_macaque_1789|prv:openrouter|mod:deepseek] Locking tool list at 30 tools ...
```

The same host's existing release-version logs contain `prv:OpenRouter|mod:deepseek` 503,047 times (2026-09-09) and 115,349 times (2026-09-10). The sessions' own metadata says the provider is DeepSeek (`"provider_key": "deepseek"` / `"openai-compatible:deepseek"`).

KV cache telemetry on the same host, same release:

```
[2026-09-16 09:51:35] [INFO] KV_CACHE_USAGE: turn=65 call=1 provider=OpenRouter upstream=None model=deepseek-flash input=799739 cache_read=2432 ...
[2026-09-16 09:50:41] [AUTH] AUTH event=auth_status_check_fast provider=all ... openrouter=not_configured ...
```

`~/.jcode/provider_activity.json` on that host (what `/usage` renders):

```json
"openrouter":                 { "month_usd": 629.99, "all_time_usd": 774.80 },
"openai-compatible:deepseek": { "month_usd": 155.73, "all_time_usd": 192.34 }
```

while `jcode auth status` reports `openrouter  not_configured`, there is no `auth.json` and no OpenRouter API key anywhere, and no OpenRouter model was ever selected (the model picker usage contains only DeepSeek, Copilot, NVIDIA NIM and generic OpenAI-compatible entries). The `openrouter` bucket therefore cannot be genuine OpenRouter spend.

## Expected

- A session using the `deepseek` profile is reported as `deepseek` / `DeepSeek`, with the session's provider key (`openai-compatible:deepseek`) used wherever a stable identity is needed.
- `/usage` attributes spend to the provider that actually served the request.

## Root cause

- `crates/jcode-base/src/provider/mod.rs:1776-1800` – `name()` returns the transport slot label (`ActiveProvider::OpenRouter => "OpenRouter"`), while `display_name()` resolves the active profile through `runtime_display_name()`. Callers that need "which provider is billing me" must use the latter (or the session provider key), but several use `name()`.
- `crates/jcode-app-core/src/agent/environment.rs:35` – `logging::set_provider_info(self.provider.name(), &self.provider.model())` puts the slot name into every log line.
- `crates/jcode-app-core/src/server/client_state.rs:500` – the `History` event's provider name falls back to `provider.name()` (the slot) when the session has no `provider_key`; the TUI stores that verbatim and reuses it as the billing identity.
- `crates/jcode-tui/src/tui/app/misc_ui.rs:327-341` – `record_api_key_spend` derives the ledger key from the TUI's display name plus `JCODE_RUNTIME_PROVIDER`; a name of `OpenRouter` with the env var unset maps straight to the `openrouter` bucket (`source_key_for_provider_label`, `crates/jcode-base/src/provider_activity.rs:310`).
- `crates/jcode-base/src/session.rs:282` – `derive_session_provider_key` falls back to env vars and finally to the raw name, so a direct profile can be recorded as `openrouter`.

## Suggested fix

Per the contract clarified in #691, `Provider::name()` should stay the stable machine-facing id. The fix belongs in the surfaces that read it while needing a profile identity:

- `Agent::set_log_context` – log `display_name()` (profile-aware) instead of `name()`.
- the `History` provider-name fallback in `client_state.rs` – same.
- `record_api_key_spend` / `/usage` ledger attribution – use an authoritative provider key rather than re-deriving it from a display label plus process env.
- `derive_session_provider_key` – fall back to the profile identity, not the raw transport name.

## Related issues

- **#691** – this is the concrete surface report that issue asked for. `name()` stays the stable machine id; these are the surfaces still reading it where the profile identity is needed.
- **#380** – a specific case of the runtime-identity divergence described there (the History fallback is quoted in that issue).
- **#1116** – same family: provider identity re-derived from a process-global env var, different trigger.
- **#795 / PR #863** – fixed the auth/route half of "OpenRouter reported while not configured". The log prefix and `/usage` ledger half is still reproducible on v0.84.0 and on `master`.

## Additional context

Related but not yet released: provider-keyed `[pricing.providers."<profile>"]` rate cards (in development after v0.84.0) are looked up with the same derived provider key. Because that key can degrade to `openrouter`, the user's card can be silently skipped and the call priced by the generic USD fallback instead. Details available on request.

I am happy to prepare a PR for the surfaces above if that helps.

ningfangbin:
Closes #1286

## Problem

The `openrouter` provider is a multiplexing slot: it serves the public OpenRouter aggregator **and** every direct OpenAI-compatible profile (`deepseek`, `nvidia-nim`, `zai`, ...). `Provider::name()` is the stable machine id for the *class* (`OpenRouter` for the slot, `openrouter` for a concrete runtime instance), while `display_name()` resolves the active profile (`DeepSeek`). Surfaces that read the class id where a profile identity belongs reported DeepSeek sessions as OpenRouter:

- every log line was tagged `prv:OpenRouter` / `prv:openrouter`
- both `History` fallback paths could report the class name
- `/usage` attributed spend to an `openrouter` bucket on hosts where OpenRouter is not configured and was never used (the ledger key is derived from that provider name)
- KV-cache telemetry printed `provider=OpenRouter` for `model=deepseek-flash`

Per the contract clarified in #691 and the `Provider` trait docs, `name()` stays the machine id; the fix belongs in the surfaces that read it.

## Changes

| File | Change |
|---|---|
| `crates/jcode-app-core/src/agent/environment.rs` | log context (`prv:`) uses `display_name()` |
| `crates/jcode-app-core/src/server/client_state.rs` | `History` provider name: the persisted-session path now maps both persisted shapes of a direct OpenAI-compatible profile (`deepseek` and `openai-compatible:deepseek`) to the profile's display name and falls back to `display_name()`; the `handle_get_model_catalog` busy-agent fallback (a second emission path into the same field) now does too. Import-source codes (`opencode`, `claude-code`, `openai-codex`) share the field but are not profile routes, so they stay verbatim |
| `crates/jcode-base/src/provider_activity.rs` | `source_key_for_provider_label` keeps an already-prefixed `openai-compatible:<id>` label instead of falling through `contains("openai")` to `openai:api-key` |
| `crates/jcode-tui/src/tui/app.rs` | KV-cache telemetry keeps the canonical provider family in `provider=` (the baseline identity the TTL/expiry classifier keys on) and adds a `label=` field with the profile, instead of overwriting the identity with a label |

`Provider::name()` is untouched, so routing, auth and restore decisions are unchanged.

## Evidence

Real-binary A/B on one machine, throwaway sandboxes per path.

Local run (`jcode run -p deepseek -m deepseek-flash "hi"`):

| binary | version | log prefix |
|---|---|---|
| pristine `master` | `v0.84.75-dev (168358c3d)` | `[ses:…\|prv:openrouter\|mod:deepseek]` |
| this branch | `v0.84.57-dev (419b4fc97)` | `[ses:…\|prv:DeepSeek\|mod:deepseek]` |

Daemon path (`serve` + a debug session, i.e. the multiplexing slot path that produced the real logs):

| binary | version | log prefix |
|---|---|---|
| pristine `master` | `v0.84.75-dev (168358c3d)` | `prv:OpenRouter\|mod:deepseek` ×14 |
| this branch | `v0.84.57-dev (419b4fc97)` | `prv:DeepSeek\|mod:deepseek` ×14 |

Both runs use two throwaway `JCODE_HOME` sandboxes and the same command (`jcode run -p deepseek -m deepseek-flash "hi"` for the first table; `serve` plus a debug session for the second), so the two prefixes never mix in one log file.

Release-binary evidence for the ledger half: on a host where `jcode auth status` reports `openrouter not_configured`, no OpenRouter credential exists and no OpenRouter model was ever selected, `provider_activity.json` still carried an `openrouter` bucket with real spend while the same host's logs held 503k+ `prv:OpenRouter|mod:deepseek` lines.

## Tests

Every changed surface is pinned by a test that fails when its fix is reverted (verified by reverting each hunk and re-running):

- `agent::environment::tests::log_context_names_the_profile_not_the_transport_slot` — new
- `server::client_state::tests::history_provider_name_maps_a_compatible_profile_key_to_its_label` — new
- `server::client_state::tests::history_provider_name_maps_a_bare_compatible_profile_key` — new (reverting the bare-key mapping fails it with `left: Some("deepseek") / right: Some("DeepSeek")`)
- `server::client_state::tests::history_provider_name_keeps_import_source_codes_verbatim` — new
- `server::client_state::tests::history_provider_name_falls_back_to_the_profile_label` — new
- `server::client_state::tests::source_key_mapping_keeps_an_already_prefixed_profile_key` — new
- `tui::app::tests::kv_cache_baseline_identity_stays_canonical_while_the_label_is_the_profile` — new (fails with `left: "DeepSeek" / right: "openrouter"` if the baseline stores the display label)
- the existing `handle_get_model_catalog_*` tests now guard the busy fallback: the shared mock's `display_name()` differs from its `name()`, so reaching for `name()` fails the assertion
- the existing `history_provider_name_preserves_unknown_runtime_profile` test moved from `opencode-go` (itself a catalog profile, i.e. no longer "unknown") to `remote-catalog`

Validation:

- focused suites: `jcode-base` provider_activity 13/13, `jcode-app-core` history + log-context 5/5, `jcode-tui` kv_cache 13/13
- full `jcode-app-core --lib`: 1225 passed / 24 failed on the branch vs 1223 passed / 23 failed on pristine `master`; the failure-name sets are identical apart from `tool::bash::tests::test_background_command_respects_timeout`, which passes when run alone (load-sensitive flake, unrelated)

An independent reviewer with clean context re-ran the suites, re-derived every revert failure, and reported no blocker; the two should-fix items it raised (a second `History` emission path, and overstated wording about `name()`'s casing) are addressed above.

Greptile's P2 was valid and is fixed: storing the rendered profile label in the KV-cache baseline disabled `KvCacheMissReason::Expired` for direct profiles, because `cache_ttl_for_provider_model` recognises provider families (`openrouter`), not labels (`NVIDIA NIM`). The baseline now keeps the canonical identity and the log line gained the separate `label=` field.

## Tooling note

Developed with **jcode** running **deepseek-v4-flash**: the fix, the regression tests and the A/B evidence above come from that session, plus a second clean-context jcode subagent that reviewed the diff independently and found the second `History` emission path now fixed here.

## CI notes

Two pre-existing guardrails are red on `master` independent of this PR, in files it does not touch:

- `cargo fmt --all --check` also wants to reformat 9 files none of which are in this diff (`jcode-harness-api-server/src/translate.rs`, `translate_tests.rs`, `jcode-protocol/src/wire.rs`, `jcode-provider-core/src/lib.rs`, `jcode-sdk/src/lib.rs`, `jcode-tui-markdown/src/lib.rs`, `jcode-tui/src/tui/app/remote/server_events.rs`, `jcode-tui/src/tui/backend.rs`, `src/cli/acp.rs`). Only the two files this PR changes were formatted.
- `scripts/check_code_size_budget.py` and `scripts/check_test_size_budget.py` already exit non-zero on pristine `master` (verified in a clean checkout of `origin/master`). This PR adds 2 lines to `crates/jcode-tui/src/tui/app.rs`, which is already over its recorded budget on `master` (2563 → 2565 LOC); the failure set is otherwise byte-identical.
- `clippy -D warnings` also trips on a crate this PR does not touch: `crates/jcode-render-core/src/markdown.rs:426` (`clippy::collapsible_if`), so clippy cannot currently be used as a signal for this diff.

## Known limitations (deliberate scope)

- **Import-source codes are excluded from the mapping.** `opencode`, `claude-code` and `openai-codex` are stored in the same field as profile keys, and `opencode` collides with the `OpenCode Zen` profile id. They stay verbatim rather than guessing. Unifying the session vocabulary with the ledger vocabulary would remove this collision properly (#380).
- **`@provider`-pinned models** still persist `provider_key = "openrouter"` (`agent/status.rs` `reconcile_explicit_provider_pin_route`). That is route/restore identity rather than a label, and it overlaps #1116; left untouched.
- **Structured telemetry keeps the class id** in its payload fields (e.g. `provider=openrouter` inside internal diagnostics) per the `Provider::name()` contract. Only the human-facing surfaces changed.
- **No end-to-end label-to-ledger test**: the ledger write happens in the TUI client and spawns a writer thread, so a unit test would have to intercept process-global I/O; the derivation is covered by the mapping tests instead.


ningfangbin:
Reviewed the Greptile P2 and fixed it in 1b07253e6.

**KV-cache TTL identity (valid):** `kv_cache_provider_name()` had started returning the rendered profile label (`DeepSeek`, `NVIDIA NIM`), which is the value stored in `KvCacheBaseline.provider` and later passed to `cache_ttl_for_provider_model`. That classifier matches provider *families* (`openrouter`, `openai`, `anthropic`, `gemini`, ...), so direct OpenAI-compatible profiles silently lost `KvCacheMissReason::Expired`.

Fix: the baseline identity is canonical again (`provider.name()` / the server-provided provider name), and the human-facing label moved to a separate field in the log line:

```
KV_CACHE_USAGE: turn=… call=… provider=openrouter label=DeepSeek upstream=None model=deepseek-flash …
```

`provider=` keeps what the TTL/expiry classifier and the baseline comparison key on; `label=` is what a person reading the diagnostic wants (issue #1286 symptom 3). New regression test `tui::app::tests::kv_cache_baseline_identity_stays_canonical_while_the_label_is_the_profile` pins both halves and fails with `left: "DeepSeek" / right: "openrouter"` if the baseline goes back to storing a label.

1jehuang:
Merged into master. The combined candidate passed history-provider, model-catalog, log-context, provider-activity and KV-cache identity tests. Real isolated CLI calls using the OpenAI-compatible profile returned History provider `OpenAI-compatible`, produced profile-aware log prefixes, and produced zero incorrect OpenRouter labels. Direct-profile mappings are covered by the focused tests.

PR #1290, opened 2026-09-17
fix(grok-build): make Grok Build usable in Jcode

### LINKED ISSUE #1289: Grok Build /model switch fails: cannot serialize RuntimeKey::Other
## Summary

After logging into Grok Build via the managed ACP provider, selecting a Grok Build model in `/model` fails before the switch reaches the server:

```
Failed to request model switch: cannot serialize tagged newtype variant RuntimeKey::Other containing a string
```

## Reproduction

1. Authenticate Grok Build (`jcode login --provider grok-build`).
2. Open `/model`.
3. Select a Grok Build route such as `grok-build:grok-4.6` (api_method `grok-build-acp`).

## Expected

The picker sends a structured `SetRoute` request and the session switches to Grok Build.

## Actual

The client cannot serialize `RuntimeKey::Other("grok-build-acp")`. `RuntimeKey` is internally tagged (`#[serde(tag = "kind")]`), and serde cannot serialize a newtype variant that contains a string.

Grok Build routes use `api_method = "grok-build-acp"`, which currently falls through to `RuntimeKey::Other`.

## Environment

- Jcode with Grok Build ACP (`grok-build-acp`)
- Model picker route: `grok-build:grok-4.6 via Grok Build (grok-build-acp)`

### LINKED ISSUE #1296: Grok Build file edits do not show diffs in the TUI
## Summary

When Jcode uses the Grok Build ACP provider, the agent edits files but the TUI does not show a file diff the way Claude does. The user only sees a status title such as `Editing src/lib.rs`, with no `+/-` view.

## Reproduction

1. Log in to Grok Build (`jcode login --provider grok-build`).
2. `/model grok-build:grok-4.6`.
3. Ask Grok to edit a source file.

## Expected

The TUI renders an `edit`/`write` tool row with a unified diff, matching Claude.

## Actual

Grok Build applies the edit itself over ACP and notifies the client with `session/update` `tool_call` (title, locations, optional `content: diff`). The Grok Build runtime previously mapped that to `StreamEvent::StatusDetail` only, so `ToolUseStart` / `ToolInput` / `ToolResult` never reached the TUI.

## Environment

- Provider: Grok Build ACP (`grok-build-acp`)
- Model: `grok-build:grok-4.6`

### LINKED ISSUE #1297: grok-4.6 context window is reported as 200k instead of 500k
## Summary

xAI documents Grok 4.6 / 4.5 as a 500,000-token context window. Jcode has no family entry for those ids, so `context_limit_for_model("grok-4.6")` falls through to `DEFAULT_CONTEXT_LIMIT` (200,000). The TUI meter and compaction budget then under-report the real window.

## Expected

`grok-4.6` and `grok-4.5` resolve to 500k. `grok-code-fast` stays 256k.

## Actual

200k default.

## Notes

This is a client-side catalog/heuristic issue only. The model still cannot take 1M; 500k is the official cap.

### LINKED ISSUE #1298: Grok Build ACP session/resume fails with FS_NOT_FOUND after cwd drift
## Summary

After a Grok Build ACP process restarts (daemon reload, dropped stream, next turn), Jcode resumes with:

```
Grok CLI ACP session/resume failed: Path not found.: {
  "code": "FS_NOT_FOUND",
  "detail": "No such file or directory (os error 2)"
}
```

The previous Grok conversation is still on disk. Resume looks in the wrong workspace folder.

## Reproduction

1. Start a Grok Build session while the Grok CLI workspace is `/mnt/c/Users/HASAKI` (or any cwd other than the Jcode daemon cwd).
2. Reload the shared daemon or otherwise drop the ACP subprocess.
3. Send another message in the same Jcode session.

## Expected

`session/resume` uses the cwd Grok originally stored the session under (`~/.grok/sessions/<percent-encoded-cwd>/<session-id>`).

## Actual

Jcode always passes `std::env::current_dir()` (often the daemon's `$HOME`). Grok looks under `~/.grok/sessions/%2Fhome%2Fviethx/<id>` instead of the original folder, e.g. `~/.grok/sessions/%2Fmnt%2Fc%2FUsers%2FHASAKI/<id>`.

## Notes

If the session folder is genuinely gone, resume should fall back to `session/new` instead of failing the turn.

### LINKED ISSUE #1300: Grok Build harness loops the last task after ACP tools
## Summary

Grok Build through Jcode can re-run the same user task every ~25–30s after a tool call. Example: `curl localhost:1424/api/v2/basic` to print degrees, then the same print again and again.

## Cause

1. ACP tool results are stored as `role: user` `tool_result` messages.
2. `build_prompt` used the latest user message including tool_result content as the next Grok prompt.
3. On resume it also re-sent Jcode's identity/system dump, which Grok wraps as a new `<user_query>`.

The Grok session then sees a new user query that restates the previous task.

## Expected

After a tool finishes, the turn ends (or continues with real user text). It must not feed the tool result / last task description back as a new prompt.

## Actual

Journal shows the same `model_usage_turn_id` with injected user `tool_result` content like `Print degrees from local BE /api/v2/basic`, then another identical curl.

vietgs03:
Fixes #1289
Fixes #1296
Fixes #1297
Fixes #1298
Fixes #1300

Make Grok 4.6 usable **inside Jcode** with the Grok CLI subscription (OIDC), not `XAI_API_KEY`.

When a scoped CLI OIDC token is present, chat goes to `https://cli-chat-proxy.grok.com/v1` and **Jcode owns tools** (same loop as Opus/Claude: native `edit`/`write`/`read`/`bash`, inline diffs, MCP, skills). If there is no token, the runtime falls back to `grok agent stdio` (ACP).

## Auth (review P1s)

- Load only the xAI issuer/client OIDC entry (`https://auth.x.ai::<client_id>`), never the first nonempty `key` in `auth.json`
- Credential store is `GROK_HOME` or `~/.grok/auth.json` for both save and load
- Honour `expires_at`, refresh with `refresh_token`, rebuild HTTP per turn (no snapshot Arc token across forks/sessions), retry on 401

## ACP fallback

- Preview in-progress edit diffs, but emit `ToolUseEnd` / `ToolResult` only on `Completed` or `Failed`
- Workspace `.mcp.json` overrides `~/.jcode/mcp.json`; a later `disabled` entry removes the server
- Jcode coordinator text is ACP `_meta.rules` on `session/new`, not a fake user query

## Also

- `/model` SetRoute serde (`RuntimeKey::Other` / `GrokBuild`)
- grok-4.6 context 500k + Jcode auto-compact on the HTTP path
- TUI inline diffs keep `old_string` / `new_string` / `content` / `patch_text`
- `session/resume` cwd lookup under `~/.grok/sessions/<encoded-cwd>/`
- Do not replay `tool_result` or `[auto] Continue the work…` as the next user prompt


1jehuang (changes requested):
Reviewed the actual 17-commit contribution (`0735c7531..a013b5d58`), separately from the divergent upstream history in the GitHub comparison. I am holding this PR rather than merging it with the other approved contributions because of these correctness/security blockers:

1. **P1: Credential scope is not checked.** [`bearer_token`](https://github.com/1jehuang/jcode/blob/a013b5d587eb9635d198a4bb79d6d8c4b35ddeda/crates/jcode-base/src/auth/grok_build.rs#L265-L284) selects the first nonempty `key` from any auth-store entry and forwards it to the fixed Grok endpoint. An isolated exact-source probe with an unrelated issuer/client entry and a valid xAI entry selected the unrelated credential. Select the expected issuer/client and supported auth mode explicitly.
2. **P1: The HTTP token lifecycle is missing.** The loader ignores expiry and refresh metadata, uses `$HOME/.grok` rather than the configured Grok home, and [provider construction](https://github.com/1jehuang/jcode/blob/a013b5d587eb9635d198a4bb79d6d8c4b35ddeda/crates/jcode-provider-grok-build-runtime/src/lib.rs#L76-L95) snapshots the bearer token for the provider lifetime. A probe returned an expired default-home token despite a valid synthetic store under `GROK_HOME`. Use one scoped store for save/load and implement coordinated refresh, external credential reload, and 401 recovery before taking authentication ownership away from the CLI.
3. **P1: An in-progress edit can be reported as successful.** [`events_for_acp_tool`](https://github.com/1jehuang/jcode/blob/a013b5d587eb9635d198a4bb79d6d8c4b35ddeda/crates/jcode-provider-grok-build-runtime/src/lib.rs#L1038-L1082) finalizes when a diff is available even with `InProgress`. A later `Failed` update is discarded because the tool is marked finished. An exact-source probe reproduced a successful ToolResult followed by zero events for the failure. Preserve the preview but emit the final result only on terminal status.

Please also address the HTTP fork sharing mutable `Arc<OpenRouterProvider>` model state across sessions, and the new MCP loader's global-first precedence overriding workspace overrides/disables. The PR/module description still describes ACP-only operation although the new HTTP path changes tool ownership.

Validation here consists of scoped review and 13 isolated source-harness tests using synthetic scratch credentials, including probes confirming the failures above. It is not a full-package or live-provider acceptance pass. No real credentials were inspected or sent. Before merging a revision, exercise the actual provider interfaces for credential refresh/scope, concurrent forks, and ACP in-progress-to-failed transitions, plus a dedicated-socket CLI/TUI workflow. This environment currently has no configured Grok subscription or CLI.

1jehuang:
Thanks @vietgs03, this PR was the reference for the direct-HTTP path. Master now ships it in a33bb7e65: Grok Build talks to `cli-chat-proxy.grok.com` directly using the Grok CLI OIDC credentials with scoped-entry selection, expiry and refresh, and a 401 retry. It presents the official Grok CLI identity headers, logs in natively via the xAI device flow, and Jcode owns tool execution. We dropped the ACP subprocess fallback entirely, since Jcode's direction is to talk to provider APIs directly rather than wrap vendor CLIs. The SetRoute serde fix had already landed separately. With that in place, and given the diff is on a very stale base (~174k lines), I'm closing this as superseded. If anything from your version is missing (e.g. the grok-4.6 context-window details or MCP handling), a focused PR on top of master would be very welcome.

PR #1319, opened 2026-09-19
fix(tui): keep drag-to-copy aligned on wrapped rows

### LINKED ISSUE #1318: Drag-to-copy returns shifted text on wrapped list, quote, and code rows
**What happens**

Highlight a single continuation row of a wrapped markdown list item (any row below the first), release, paste. The pasted text does not start at the highlighted row; it starts N characters early, where N is the marker width (`1. ` = 3, `• ` = 2) times the number of rows above it, so the highlighted row's own start is cut off.

Selecting the whole item is fine, and output that does not wrap is fine, which makes it look like a one-off.

**Repro**

1. Have an assistant message contain one long ordered-list item that wraps over roughly 5 rows, for example:

   `1. copyrepro alpha bravo charlie delta echo foxtrot golf hotel india juliet kilo lima mike november oscar papa quebec romeo sierra tango uniform victor whiskey xray yankee zulu REPRO-MIDDLE alpha bravo charlie delta echo foxtrot golf hotel india juliet kilo lima mike november oscar papa quebec romeo sierra tango uniform victor whiskey xray yankee zulu REPRO-END copythisstringandpastetotest final words here`

2. Drag-select only a continuation row of that item, left edge to right edge. Release.
3. Paste.

Expected: the pasted text starts with the highlighted row's first characters.
Actual: it starts with characters from an earlier row. Measured drift for a 3-column marker: 6 characters at two rows down, 12 at four rows down, 21 at seven.

**Cause**

`crates/jcode-tui/src/tui/ui_prepare.rs`, in `wrap_lines` and `wrap_lines_with_map`.

Wrapping re-seeds the gutter / hanging-indent prefix on every continuation row, so a row's display width is `prefix + source text width`. The map advanced the source segment by the full display width and applied `wrapped_copy_offsets` only to row 0 (`remaining_copy_offset` saturates to 0 after that). Every continuation row therefore moved the source column forward by the prefix width too much.

Same class affects wrapped code and quote rows (gutter width) and wrapped multi-line user prompts (marker width).

**Environment**

jcode v0.85.0 (01c8cdf48), Linux x86_64, foot 1.28.0, Wayland.

zipadoodlez:
Wrapped rows re-seed a display-only prefix (list hanging indent, `│ ` gutter). The
copy map counted those columns as source text and offset only row 0, so
drag-to-copy returned shifted text on any continuation row. Both wrap loops now
recover the prefix from `sum(row widths) == first_copy_offset + source_span +
prefix * (rows - 1)` and use it as each row's copy offset.

Closes #1318

Not compiled or tested locally (could not afford the `cargo test -p jcode-tui`
build); CI is the first real build. Repro and numbers are in #1318.


Akumar96:
wow zipadoodoo goodjob

PR #1333, opened 2026-09-20
fix(tui): give drag-edge autoscroll its own rate; replace wheel momentum with discrete scrolling

### LINKED ISSUE #1332: TUI: drag-to-copy edge autoscroll scrolls ~3x too fast and glides after release
## Summary

While drag-selecting, holding the cursor at the top or
bottom edge of the chat pane scrolls far faster than one line per tick, and keeps
drifting after the mouse button is released.

## Repro

1. Run the TUI with a transcript taller than the viewport.
2. Press and drag below the last visible line (or above the first) and hold.

Expected: browser-like continuous scroll, roughly one line per tick, stopping on
release.
Actual: ~3 lines per frame (about 180 lines/s at 60fps), plus a short momentum
glide after release.

## Root cause

`progress_copy_selection_edge_autoscroll` (`app/copy_selection.rs`) routes through
`scroll_copy_selection_pane` -> `enqueue_mouse_scroll` (`app/navigation.rs`), which
is the *mouse-wheel momentum* path:

- velocity multiplier: gap <= 30ms means 2x. A 60fps tick is ~16ms, so every
  autoscroll tick was classified as a hard flick.
- intent becomes `min(3 * 2, 5) = 5` lines, added to a queue capped at 30.
- each call drains up to 3 lines, and `mouse_scroll_drain_amount` drains 3 more
  per frame while the queue is >= 6.

The queue therefore saturates at 30 and drains a flat 3 lines per frame. On
release the leftover queue keeps draining at 3/frame for ~10 frames, which is the
glide.

A programmatic per-tick scroll was being fed through a human-flick model.

## Proposed fix

Give the drag autoscroll its own per-line step instead of the wheel queue, and
define its rate by a fixed tick cadence so it does not inherit `redraw_fps`.
Details in the PR.

zipadoodlez:
Closes #1332

## Two separate scroll paths

This PR touches two mechanisms that are easy to conflate. Naming them once so the
rest reads cleanly:

- **Mouse-wheel scrolling** — the user turns a wheel or trackpad. The terminal
  delivers discrete notch events. This is the path that had momentum
  (`enqueue_mouse_scroll`, a velocity multiplier, a queue, a frame drain).
- **Drag-edge autoscroll** — while drag-selecting the transcript, *jcode scrolls
  by itself*, one step per redraw tick, while the cursor sits at the top or bottom
  edge of a pane (`progress_copy_selection_edge_autoscroll`). No wheel is involved,
  and no mode is required: the default drag-to-copy reaches this path, and `Alt+y`
  only pins copy mode on explicitly.

The reported bug is **entirely in the second mechanism**. Its cause is that it
borrowed the first mechanism's model.

## The bug (drag-edge autoscroll)

While drag-selecting the transcript — the default browser-style drag-to-copy, no
mode needed, and also explicit copy mode (`Alt+y`) — holding the cursor at the top
or bottom edge of the chat pane scrolls ~3 lines per frame (~180 lines/s at 60fps)
and keeps drifting after the mouse button is released. The intent was a
browser-like continuous scroll of roughly one line per tick that stops on release.

## Root cause (drag-edge autoscroll used the wheel model)

`progress_copy_selection_edge_autoscroll` (`app/copy_selection.rs`) called
`scroll_copy_selection_pane` -> `enqueue_mouse_scroll` (`app/navigation.rs`), the
wheel path. That path infers "how hard the wheel was flicked" from the gap
between wheel events and then queues an intent to drain over frames.

An autoscroll tick fires every ~16ms at 60fps, so the gap looked like a hard
flick (2x): each tick queued `min(3 * 2, 5) = 5` lines against a 30-line cap,
while each call drained up to 3 lines and `mouse_scroll_drain_amount` drained 3
more per frame. The queue saturated at 30 and drained a flat 3 lines/frame. After
release, the leftover queue kept draining for ~10 frames, which is the drift.

So: a programmatic, tick-driven scroll was being expressed through a model of
human flick physics. A terminal reports discrete notches, not force, so that
model's input was never real; it just happened to be harmless for actual wheel
input and actively wrong for this caller.

## Why this is more than one change

CONTRIBUTING asks for focused changes and to split large work. This is one
coherent change with two separable riders; I am happy to split any of it.

**Core (not separable).** The bug exists *because* the autoscroll shares the wheel
path. Two ways to fix it:

1. Carve the autoscroll out of that path, keep the momentum model for the wheel.
   Smaller diff, but it leaves the trap: the next programmatic scroll that reaches
   for the shared entry point reproduces the bug, and the queue's constants stay
   load-bearing for a caller they were never designed for.
2. Remove the momentum model, make a wheel notch a discrete command.

I chose (2), so the core is "fix the autoscroll" plus "delete the mechanism it was
misusing". Those are one decision, which is why the core is larger than a clamp.

**Rider 1 (separable, feel).** Removing the wheel momentum model also removed the
only thing that made a fast flick cover more ground. With no flick model at all,
`WHEEL_LINES` would be a flat 3. This rider restores "deliberate notch stays
precise, flick covers ground" as a discrete notch-size ladder, and gives the
autoscroll a proximity ladder for the same reason. Pure taste; drop it and keep a
flat `WHEEL_LINES` if you prefer.

**Rider 2 (fully separable, no behavior change).** help / changelog / model-status
each carried their own copy of the same key match, mouse step, and render clamp.
I deduped them because I was already reasoning about those three offsets. Splitable
into its own PR with no effect on the rest.

Commit map for a subset:
- core: `e2aa58b1a`, `71420106b`, `ba522af28`, `d091dc40c`
- rider 1: `d6c22a4a3`, `ea741191a`
- rider 2: `2239560a3`
- cleanup: `1407449d9`

### Drag-edge autoscroll (the fix)

- Steps lines through the plain per-line primitive, and never calls the wheel
  entry point. A programmatic scroll can no longer be mistaken for a flick.
- Rate comes from a fixed tick cadence (`REDRAW_COPY_AUTOSCROLL`, 60ms) instead of
  inheriting `redraw_fps`, so speed is a property of the gesture, not the display.
- With rider 1: 1-3 lines per tick, scaled by distance into the edge hot zone
  (`zone - depth` clamped to 1..=3).

### Mouse-wheel scrolling (model replacement)

- `scroll_wheel` moves `WHEEL_LINES` (3, matching Neovim's `mousescroll ver:3`)
  immediately, and `scroll_rows` applies a native host's exact row delta.
- Deleted: the momentum queue and its fields, the velocity multiplier, the
  ease-out frame drain, the `mouse_scroll_animation` redraw reason,
  `has_pending_mouse_scroll_animation` (trait method plus eight schedule call
  sites), and the session picker's dead write-only timestamp.
- With rider 1: wheel lines scale with the gap between notches — 3 deliberate,
  6 at <=40ms, 9 at <=15ms, capped at 10.
- Wheel input is otherwise unchanged: same targets (chat, side pane, overlays,
  picker preview), same routing.

### Overlays (rider 2, no behavior change)

One `overlay_scroll_key`, one `overlay_scroll_step`, one `overlay_scroll_info`
replace three copies each. `Option<usize>` (`None` = closed, `usize::MAX` = bottom)
was already the model, so no new type.

Net: 20 files, +361/-527.

## Behavior changes, per mechanism

**Drag-edge autoscroll**
- 1-3 lines per 60ms tick by proximity, ~17-50 lines/s, independent of fps.
- Stops immediately on release. No drift.

**Mouse-wheel scrolling**
- A notch is now a fixed number of lines instead of a queue: 3, up to 10 on a fast
  flick.
- No momentum glide.
- Two rapid notches count as a flick (up to 9 lines). Intended, and the one place
  the new ladder is visible. It is why two overlay tests had to clear their
  inter-notch timestamp to measure the base tier.

**Native scrollbar (host-provided deltas)**
- The full delta applies immediately, with no animated remainder.

## Tradeoffs and edge cases

- Wheel momentum glide is gone. It only triggered on very fast flicks before (a
  slow notch drained fully inside the same call), so the perceived loss is small,
  but it is a deliberate removal.
- The wheel notch ladder is jcode's own heuristic, not copied from Neovim. Marked
  with a `ponytail:` note and tunable in one place (`wheel_lines_for_gap`,
  `WHEEL_LINES_MAX`).
- The autoscroll cadence pins the redraw loop to 60ms while a drag autoscroll is
  active, including during streaming. Fine for a transient gesture; flagging it.
- Autoscroll still keeps scrolling while the cursor is held perfectly still. That
  is jcode's existing behavior. Neovim instead scrolls per drag event, by the
  distance the cursor is past the edge; we did not adopt that.
- Untouched: keyboard chat scrolling, side-pane scrolling, tail catch-up, elastic
  overscroll.

## Verification

- `cargo check -p jcode-tui` clean.
- `cargo check -p jcode-tui --all-targets` clean.
- `cargo test -p jcode-tui --lib`: 2302 passed.
  `test_edge_autoscroll_rate_is_proximity_scaled_and_stops_on_release` asserts the
  autoscroll rate, that held ticks do not accelerate, that the proximity tier never
  gets faster deeper in the zone, and that nothing drifts after release.
- Manual, Linux/x86_64, foot 1.28.0, Wayland: drag-select in the transcript and
  hold the cursor at the bottom edge, both in default mode and after `Alt+y`. Slow
  near the inner edge, faster on the boundary, stops on release.
- 18 tests fail in the full suite on this machine (accounts, onboarding, golden
  render, remote reload). They are pre-existing and unrelated: several pass in
  isolation (flaky under parallel execution), one fails deterministically in
  `commands_accounts_02`, and none are in files this PR touches.

## Review notes

- Removed tests only covered the deleted wheel queue (acceleration multiplier,
  drain deceleration, queue saturation). The behavior they asserted does not exist
  anymore.
- One native-scroll test changed: `enqueue_native_scroll` now applies its exact row
  delta immediately, so `state_model_poke_02` asserts immediate deltas rather than
  queued rows.


1jehuang:
Thanks! Closing in favor of #1400, which fixes #1332 with the narrower change (drag-edge autoscroll gets its own path, wheel momentum untouched) and is now merged. Replacing wheel momentum can be proposed separately if still wanted.

Issue #1342, opened 2026-09-20
CI red on master: 7 remaining jcode-tui test failures with no shared cause


MatrixMagician:
## Description

`cargo test -p jcode-tui --lib` on `master` at `e589cbe5a` (v0.86.0) reports `2344 passed; 11 failed`. Three of the eleven are #1340 and one is #1339. This issue tracks the remaining seven, which have no shared cause that I could find.

Each one was re-run individually in its own process and still fails, so none of these is the shared-process-global-state flakiness tracked in #596.

Symptoms are recorded below; I have not traced any of the seven to a specific commit.

## Failing tests

**1. Hidden reasoning reaches the judge-session transcript** — the one worth looking at first, since it is the only containment assertion in the set.

```
tui::app::tests::test_prepare_review_spawned_session_uses_visible_transcript_for_judge_sessions
  crates/jcode-tui/src/tui/app/tests/remote_startup_input_01/part_01.rs:280
  assertion failed: !transcript.contains("hidden reasoning should never leak")
```

**2. Fallback notice never set**

```
tui::app::tests::recent_project_review_falls_back_cleanly_when_no_repo_is_known
  crates/jcode-tui/src/tui/app/tests/onboarding_flow.rs:1676
  assertion failed: app.status_notice.as_ref().is_some_and(|(notice, _)|
      { notice.contains("No recent Git repository found") })
```

**3. Account switch by label picks the wrong account**

```
tui::app::tests::test_account_switch_shorthand_switches_openai_account_by_label
  crates/jcode-tui/src/tui/app/tests/commands_accounts_02/part_01.rs:557
  left:  Some("openai-otter")
  right: Some("openai-1")
```

**4. Logout by label cannot find the account**

```
tui::app::tests::test_logout_clear_anthropic_accounts_removes_all_accounts_once
  crates/jcode-tui/src/tui/app/tests/state_model_poke_02/part_01.rs:1046
  called `Result::unwrap()` on an `Err` value: No account with label 'claude-3' found
```

**5. Gate digest cycle does not finish**

```
tui::app::tests::test_gate_digest_is_delivered_at_turn_end_and_rearms_next_cycle
  crates/jcode-tui/src/tui/app/tests/remote_events_reload_05.rs:945
  with nothing left outstanding the cycle should finish
```

**6. Onboarding banner layout**

```
tui::session_picker::tests::onboarding_banner_renders_prompt_and_both_action_rows
[...]

Issue #1339, opened 2026-09-20
Session::save() skips untouched sessions, so is_debug and canary never persist (4 failing tests)


MatrixMagician:
## Description

`Session::save()` skips writing untouched sessions, so `is_debug` and `canary` — explicit state, set at creation time — are silently dropped. Four tests on `master` fail as a result.

This is the same root cause already identified in #1119 (second comment) and reported again in #1249. Those issues cover the swarm/`--resume` symptom. This one covers a different symptom and, more usefully, names the exact missing entries in the exemption list.

## Root cause

`crates/jcode-base/src/session/persistence.rs:390`:

```rust
if !self.persist_state.snapshot_exists
    && !self.messages.iter().any(super::is_visible_conversation_message)
    && !self.saved
    && self.custom_title.is_none()
    && self.title.is_none()
    && self.parent_id.is_none()
{
    return Ok(());
}
```

`create_headless_session` (`crates/jcode-app-core/src/server/headless.rs:161`) calls `Agent::set_debug(true)`, which sets the flag and immediately saves:

```rust
// crates/jcode-app-core/src/agent/turn_execution.rs:349
pub fn set_debug(&mut self, is_debug: bool) {
    self.session.set_debug(is_debug);
    if let Err(err) = self.session.save() {
        logging::error(&format!("Failed to persist debug session state: {}", err));
    }
}
```

A freshly created headless session has no visible messages, no title, no parent — so `save()` early-returns, writes nothing, and returns `Ok(())`. The error branch never fires, so nothing is logged either. `Session::load()` then fails with `No such file or directory (os error 2)`.

The commit that introduced the early return, `783c979a0 "Avoid persisting untouched sessions"` (2026-08-26), carries a comment explaining that a caller-chosen `title` and a `parent_id` are explicit state and must persist before the first visible message (citing #1144). `is_debug` and `canary` are explicit state by exactly that argument and were not added.

## Impact beyond the tests

A debug or canary session created and not yet used loses its flag on disk. Anything that later reads the session back — resume, takeover, `jcode session` listings — sees a non-debug, non-canary session.

## Reproduction

At `e589cbe5a` (v0.86.0), Linux:

```console
$ cargo test --test e2e
test result: FAILED. 56 passed; 3 failed; 7 ignored

failures:
    session_flow::test_clear_preserves_debug_for_resumed_debug_session
    session_flow::test_debug_create_selfdev_session_marks_canary
    session_flow::test_debug_create_session_marks_debug
```

[...]

PR #1344, opened 2026-09-20
fix: repair copy-selection tests and OrcaRouter provider choice

### LINKED ISSUE #1340: Copy-selection notice gained a "· highlight remains visible" suffix; 3 TUI tests still assert the old string
## Description

`c7afd6620` / `f4bc4db5f "Improve copy selection visibility"` (2026-08-28) changed the copy-selection status notice to add a `· highlight remains visible` suffix. Three tests still assert the old exact string and now fail on `master`.

This is the first of the two failures blocking the Linux `Build & Test` job ([run 35504384040](https://github.com/1jehuang/jcode/actions/runs/35504384040)), so everything after `Run TUI library tests` is skipped.

## Root cause

`crates/jcode-tui/src/tui/app/copy_selection.rs:414` now emits:

```rust
"Copied selection · highlight remains visible"
```

The code change looks intentional; the tests were not updated with it.

## Failing tests

```
tui::app::tests::test_input_composer_drag_selects_and_copies_typed_text
  crates/jcode-tui/src/tui/app/tests/input_copy_selection.rs:102
  left:  Some("Copied selection · highlight remains visible")
  right: Some("Copied selection")

tui::app::tests::test_input_composer_drag_then_release_copies_via_full_mouse_path
  crates/jcode-tui/src/tui/app/tests/input_copy_selection.rs:411
  drag release over the composer must attempt a copy,
  got Some("Copied selection · highlight remains visible")

tui::app::tests::test_changelog_overlay_mouse_drag_release_copies_text
  crates/jcode-tui/src/tui/app/tests/scroll_copy_02/part_01.rs:1481
  assertion failed: matches!(app.status_notice().as_deref(),
      Some("Copied selection") | Some("Failed to copy selection") | Some("Selection is empty"))
```

## Reproduction

At `e589cbe5a` (v0.86.0), Linux:

```console
$ COLORTERM=truecolor cargo test -p jcode-tui --lib -- --test-threads=1 \
    --skip test_prompt_entry_shimmer_color_moves_across_positions \
    --skip right_fact_stack_uses_neutral_gray_except_for_context_usage
```

All three also fail when run individually in their own process, so this is not the shared-global-state flakiness tracked in #596.

## Suggested fix

Update the three assertions to the current notice text. If the suffix is meant to be conditional, the tests should assert the prefix rather than the whole string.

Related: #1339, #1341, #1342, #596.

### LINKED ISSUE #1341: OrcaRouter is in the login catalog but has no ProviderChoice, so it can't be selected explicitly
## Description

`802f1fe8c "feat(provider): add OrcaRouter profile"` (2026-08-16) added OrcaRouter to the login catalog but not to the `ProviderChoice` mapping table. `choice_for_login_provider("orcarouter")` returns `None`, so OrcaRouter cannot be selected as an explicit provider choice, and `provider_matrix` fails.

## Root cause

`ORCAROUTER_LOGIN_PROVIDER` is present in:

- `crates/jcode-provider-metadata/src/catalog.rs:633` (definition)
- `crates/jcode-provider-metadata/src/catalog.rs:1265` (`LOGIN_PROVIDERS`, 54 entries)

but absent from `PROVIDER_CHOICE_LOGIN_PROVIDERS` in `src/cli/provider_init.rs:209`, which is what `choice_for_login_provider` searches:

```rust
pub fn choice_for_login_provider(provider: LoginProviderDescriptor) -> Option<ProviderChoice> {
    PROVIDER_CHOICE_LOGIN_PROVIDERS
        .iter()
        .find(|(choice, candidate)| {
            candidate.id == provider.id && !matches!(choice, ProviderChoice::ClaudeSubprocess)
        })
        .map(|(choice, _)| *choice)
}
```

There is no `ProviderChoice::Orcarouter` variant at all.

Diffing the two tables, exactly two of the 54 login providers have no `ProviderChoice` entry:

- `AUTO_IMPORT_LOGIN_PROVIDER` — intentional, not a real provider
- `ORCAROUTER_LOGIN_PROVIDER` — the gap

## Impact beyond the test

OrcaRouter is reachable through `jcode login` but not as an explicit `--provider` choice, unlike every other OpenAI-compatible profile in the catalog.

## Reproduction

At `e589cbe5a` (v0.86.0):

```console
$ cargo test --test provider_matrix
test result: FAILED. 8 passed; 1 failed; 0 ignored

---- provider_matrix_explicit_compatible_choice_overrides_stale_active_profile_state_space stdout ----
thread '...' panicked at tests/provider_matrix.rs:576:32:
orcarouter should map to a ProviderChoice
```

The test iterates every OpenAI-compatible login provider and requires each to round-trip to a `ProviderChoice`; OrcaRouter is the only one that does not.

Also reproduced upstream on the macOS runner in [run 35504384040](https://github.com/1jehuang/jcode/actions/runs/35504384040) (`Run provider matrix tests`). On the Linux runner the step is skipped because the TUI step fails first (#1340).

## Suggested fix

Add a `ProviderChoice::Orcarouter` variant and the corresponding `PROVIDER_CHOICE_LOGIN_PROVIDERS` entry, matching how the other OpenAI-compatible profiles are wired.

Related: #1339, #1340, #1342.

MatrixMagician:
Two of the failures that have `master` red, fixed independently. Each commit stands alone and can be reviewed or reverted on its own.

Closes #1340. Closes #1341.

## 1. `fix(tui)` — copy-selection tests vs. the preserved-highlight notice

`c7afd6620` / `f4bc4db5f` ("Improve copy selection visibility") changed the copy notice to `Copied selection · highlight remains visible` and made the preserving path keep the selection visible after a copy. One test was updated with it (`scroll_copy_02/part_01.rs:687`); three were not.

- `test_input_composer_drag_selects_and_copies_typed_text` pins the preserving path, so it asserts that notice exactly. It also asserted `copy_selection_anchor.is_none()` / `copy_selection_cursor.is_none()` — the old semantics. `copy_current_selection_preserving_highlight` deliberately leaves both intact (that is the feature), clearing only `copy_selection_mode`, `copy_selection_dragging`, `copy_selection_pending_anchor` and `copy_selection_edge_autoscroll`. The test now asserts the highlight is retained and that copy-selection mode ended.
- The two "a copy was attempted" checks (`input_copy_selection.rs:411`, `scroll_copy_02/part_01.rs:1481`) test that the selection path ran, not the exact wording, so they match on the notice prefix. A later wording change cannot silently turn a successful copy into a test failure.

**Also closes a hole the same change opened.** `test_input_composer_click_still_moves_caret` asserted `status_notice() != Some("Copied selection")` to prove a plain click does not copy. The suffixed notice satisfies that inequality, so the test would have passed even if a plain click had started copying. It now rejects any notice beginning with `Copied selection`.

## 2. `fix(providers)` — OrcaRouter had no `ProviderChoice`

`802f1fe8c` ("feat(provider): add OrcaRouter profile") added OrcaRouter to the login catalog but not to `PROVIDER_CHOICE_LOGIN_PROVIDERS`, so `choice_for_login_provider` returned `None` for it.

This is not only a test failure: OrcaRouter was reachable through `jcode login` but not as an explicit `--provider` choice, unlike every other OpenAI-compatible profile in the catalog.

Diffing the two tables, of the 54 entries in `LOGIN_PROVIDERS` exactly two had no `ProviderChoice` — `AUTO_IMPORT` (intentional, not a real provider) and `ORCAROUTER`. The fix wires the same four sites the other compatible profiles use: the enum variant (carrying the catalog's `orca-router` alias), `as_arg_value`, the login-provider mapping, and the shared OpenAI-compatible init arm, which resolves the profile through `profile_for_choice`. No new match arm logic — OrcaRouter joins the existing `|` chain.

## Verification

Linux, built from `e589cbe5a`.

| Gate | Before | After |
|---|---|---|
| `cargo test -p jcode-tui --lib -- --test-threads=1` | 2344 passed / 11 failed | **2347 passed / 8 failed** |
| `cargo test --test provider_matrix` | 8 passed / 1 failed | **9 passed / 0 failed** |
| `cargo test --test e2e` | 56 passed / 3 failed | 56 passed / 3 failed |
| `cargo build --profile selfdev` | clean | clean |

The 8 remaining TUI failures are #1339 (1) and #1342 (7); the 3 e2e failures are #1339. Neither is touched here.

Adding the enum variant produced no non-exhaustive-match errors, which confirms those four sites are the complete set.

## Notes and tradeoffs

- **Commit 1 is test-only.** It changes no production behaviour: it makes the tests agree with the behaviour `c7afd6620` deliberately introduced. If the preserved highlight was *not* intended, the right fix is in `copy_selection.rs` instead and this commit should be dropped.
- **`cargo fmt --check` still fails on `master`**, in `ui.rs`, `ui_pinned.rs`, `app/tests.rs` and `app/commands_tests.rs`. Pre-existing and untouched here — reformatting them would bury a two-file fix in unrelated churn. The three files this PR touches are fmt-clean.
- **The e2e `ambient` cohort flakes** independently of this change: one full run showed `52 passed / 7 failed`, the extra four all `ambient::*`, which passed in the run immediately before with the identical binary and pass 3/3 when the cohort runs alone. Unrelated to these commits, and not filed.
- The e2e failures in the table are invisible in CI today because the Linux `Build & Test` job stops at the TUI step (#1342), so the e2e step never runs. Fixing the copy tests moves that boundary but does not clear it — #1342's remaining 7 still fail.


1jehuang:
Thanks! Both fixes are already on master (OrcaRouter ProviderChoice and the updated copy-selection notice assertions). Closing as superseded.

Issue #1331, opened 2026-09-20
Cost estimates diverge from real bills when a model has time-of-use, expiring, or per-route pricing


ningfangbin:
## Problem

jcode prices each call from a single flat rate per model, sourced from the
[models.dev](https://models.dev) catalog, and falls back to provider caches and static tables
when that catalog has no entry. That catalog can express a flat price, long-context tiers, and
experimental-mode costs, but it has **no time dimension and no validity window**, and every
figure jcode records is a single flat `ModelCost` with no notion of when it applies or which
route it describes.

As a result there are several common situations where the cost that jcode displays (and the
totals that `/usage` accumulates) drifts substantially away from the user's actual bill, and
the user has **no way to correct it locally**:

- **Time-of-use (peak/off-peak) pricing.** Some vendors bill different rates depending on when a
  call happens (for example, a higher rate during specific UTC windows on weekdays). models.dev
  has no field for a schedule, so the catalog can only hold one flat figure, and jcode's flat
  per-model rate cannot represent a window either.
- **Limited-time promotions and expiring prices.** A discounted rate that reverts on a fixed
  date is expressed today only by a maintainer editing the number back and forth; there is no
  `effective_from`/`effective_until` on the data, and a consumer has no way to know a price is
  time-boxed. jcode's flat rate has no expiry, so a stale promotional figure silently over- or
  under-counts after the promotion ends.
- **Per-route price differences for the same model id.** The same model can be billed at
  different rates depending on which route/aggregator a call goes through (a vendor's own
  endpoint versus a router endpoint), even though the model id is identical. Because jcode keys
  cost on model id alone, it cannot distinguish these.

The practical impact is that the 💰 cost display and `/usage` totals can differ meaningfully
from the real bill for affected models, and the user cannot fix the discrepancy. The cases that
matter most (time-of-use, expiry, per-route) are precisely the ones where the *user* has
authoritative knowledge that a static catalog cannot carry: the user knows their own schedule,
their own negotiated rates, and which route a given model id actually resolves to in their
setup.

## Why existing mechanisms cannot express this

- The models.dev catalog schema has no time, validity, or schedule dimension, so the schedule
  and expiry information has nowhere to live; expecting it to grow such a dimension for every
  vendor is unrealistic.
- jcode's internal cost type is a single flat rate card per model (input/output, plus optional
  cache read/write rates, all in USD per million tokens), so there is no place to attach a
  schedule, an expiry, or a route scope.
[...]

PR #1353, opened 2026-09-21
feat(pricing): add a config-authoritative pricing layer (user rates, tariffs, currency)

ningfangbin:
**Draft — posted so the discussion has something concrete to react to, not as a merge request.** Happy to re-scope, split, or re-spell the surface per your steer; see "Where I'd like your steer" at the bottom.

Tracks #1331.

## What this is

A config-authoritative pricing layer: the rates you write down outrank everything jcode derives itself. Rules match on the `(activity/source key, model id)` pair that already prices and books a call, with an optional route narrowing. Anything the rules do not cover keeps today's chain (curated tables → OpenRouter caches → models.dev → generic fallback) unchanged, so an empty `[pricing]` section is a strict no-op.

## Why — the case this came from

The route I actually pay for is a DeepSeek profile. Today a self-named `[providers.<name>]` route resolves to no catalog entry, so its calls fall back to the generic default rates instead of the numbers on my bill. With this branch:

```toml
# config.toml — the vendor label is my own; the file is plain local JSON, never fetched
[pricing]
fx_base = "USD"

[pricing.fx_rates]
CNY = 7.2

[pricing.providers.deepseek]
file = "deepseek.json"
currency = "CNY"
```

```json
// deepseek.json (excerpt) — re-read when the file changes
{
  "models": {
    "deepseek-v4-flash": {
      "cost": { "input": 1.0, "output": 4.0, "cache_read": 0.02 },
      "tariffs": { "peak": { "multiplier": 2.0 } },
      "schedule": [
        {
          "tariff": "peak",
          "utc_offset_minutes": 480,
          "weekdays": ["Mon", "Tue", "Wed", "Thu", "Fri"],
          "windows": [["09:00", "12:00"], ["14:00", "18:00"]]
        }
      ]
    }
  }
}
```

The numbers are in CNY (the currency the rates are denominated in, not the provider's), the peak multiplier applies to calls **at the instant they ran**, and `/pricing` answers "which layer, which tariff, which currency priced this call". Before: the same route showed a generic $15/$60-per-Mtok fallback estimate.

## Scope of this branch

- `[pricing]` schema and a `Currency` type; local vendor files (never fetched from the network).
- Rules matched by model id with optional route narrowing; per-million-token costs, named tariffs + schedules (fixed UTC offset), `effective_from`/`effective_until`, long-context tiers.
- Multi-currency spend ledger (rollback-safe migration) and currency-aware display.
- `/pricing` explainability; vendor-file parse reports in the config notice.
- Config-save hardening: preserve comments/unmodeled sections on save, replace inference-from-omission with explicit deletion declarations, never overwrite an unparsable `config.toml` (orthogonal — happy to move it to its own issue/PR).
- Docs: `docs/MODEL_PRICING.md` (its quick-start block is kept executable by a test).

Size against current `master`: ~+14.1k/−0.7k lines across 82 files — engine, vendor data, tests, docs. Covered by tests in `jcode-base` and `jcode-tui`.

## Where I'd like your steer

The surface above is the broader design; the **minimal route-scoped override contract** discussed in the issue maps onto the same resolver and is my natural first cut if that is the shape you want first — keyed by source key, exact `(source key, model id)` matching, inline rates, loud validation for partial rules, no-op when the section is empty. Same offer for the `Config::save` hardening: it can become its own issue/PR.

Questions from my earlier comment are still open and would settle the shape:

1. Is the activity/source key the right identity for a rule to name (vs the runtime provider label or the models.dev provider id)?
2. For `[providers.<name>]` routes, should a rule name the profile as the user wrote it, or the derived `openai-compatible:<id>`?
3. Currency: display-only in the first cut, or recorded per currency in the ledger?

I will not ask for review or merge on this draft until those are settled.


ningfangbin:
All four findings are addressed in `73b5ed78d`, `8b6ef9c34`, `93b95097e`, `0ef8c9cbd` (with `a37f44eb0` aligning one older ledger test to the USD-only mirror semantics), with tests for each reproduction:

**1 & 2 — server pricing / call-start instant (`server_events.rs`)**

The server now resolves each call's cost with the same `model_pricing` path the client would — at the call's own start instant, under the provider/model it actually used — and reports it on `ServerEvent::TokenUsage` through two new optional fields, `cost` and `currency` (serde-optional: the previous shape still parses, and an old client is unaffected). The TUI prefers the server value for both display and accrual, so two clients with different cards/currency/vendor files/schedules can no longer bill the same call differently; local pricing remains only as an explicit compatibility fallback (older server, or a call no layer can price). The first remote snapshot is priced at the recorded call start rather than the snapshot's arrival.

Tests: `server_resolved_cost_wins_over_the_clients_local_card` (client-local card $99/$199 would compute $298 for the same tokens; the server's $3.00 is what gets booked — and with no server value the local path still computes $298, which pins the non-empty case), plus protocol round-trip tests for the optional fields, `remote_first_snapshot_is_priced_at_the_call_start_not_arrival`, and `remote_first_snapshot_without_a_recorded_start_falls_back_to_now`.

**3 — currency on rollback (`provider_activity.rs`)**

The legacy `*_usd` mirrors now hold the USD bucket's value only; a cross-currency sum is never written into a USD-named field. A rollback round-trip through an older binary therefore no longer turns `{CNY 30, USD 5}` into "USD 35" — the non-USD amounts drop out of the older binary's view instead of being relabelled as USD, and the trade-off is documented on the type.

Tests: `usd_mirror_tracks_only_the_usd_bucket`, `a_cny_only_window_keeps_a_zero_usd_mirror`, `rollback_roundtrip_keeps_the_usd_bucket_and_never_relabels`.

**4 — failed saves (`config_file.rs`)**

`Config::save()` now snapshots the declared removals for the write and clears them only after the save has actually succeeded; a failed write leaves the declaration in place so a later successful save still applies the deletion.

Tests: `a_failed_save_preserves_the_declared_removal`, `a_successful_save_consumes_the_declaration_exactly_once`.

Two limits are stated in the code rather than silently left: a replayed timeline carries tokens only, so replay prices locally; and the delta branch of a multi-snapshot call still prices locally (the server reports one full-amount event per call).

ningfangbin:
Fixed in `50cb25895`: `Config::save()` now consumes **only the removal declarations its own snapshot applied** — `consume_declared_removals(&removals)` retains the pending entry against that snapshot, so a removal declared after the snapshot (the concurrent revocation case) is not dropped by this save and is applied by the next one. The failed-save path is unchanged: the `?` sits before the consume, so a failed write keeps every declaration for a later successful save.

Tests: `a_save_consumes_only_the_removals_it_snapshotted` (the deterministic interleaving from the finding — snapshot `a` → declare `b` → consume → pending stays `["b"]`) and `a_second_save_does_not_reapply_an_already_consumed_removal`; `config::save_tests` is 17/17 green, and the full `jcode-base` suite keeps the same pre-existing failure set.

Issue #875, opened 2026-08-10
cargo test -p jcode-tui --lib deadlocks silently: lock-order inversion between storage::lock_test_env and ui::render_state_test_lock


potb:
`cargo test -p jcode-tui --lib` can hang forever instead of finishing. The run makes no progress, prints nothing (the test harness buffers until the end), and only ends when a human cancels it. I hit it for 10 minutes before cancelling.

It is not slow compilation and not a slow test: it is a lock-order-inversion deadlock between the two process-global mutexes the test harness uses.

## The two locks

- `jcode_base::storage::lock_test_env()` (`crates/jcode-base/src/storage.rs:34`), guarding env vars + the config cache.
- `jcode_tui::tui::ui::render_state_test_lock()` (`crates/jcode-tui/src/tui/ui.rs:1452`), guarding process-global render state.

Many tests need both. They disagree on the order:

**Order A, env then render** (`crates/jcode-tui/src/tui/ui_tests/prepare.rs:621`, also `:1180`, `:1241`):

```rust
let _guard = crate::storage::lock_test_env();   // env lock
clear_test_render_state_for_tests();            // -> with_render_state_lock -> render lock
```

**Order B, render then env** (`crates/jcode-tui/src/tui/ui_tests/swarm_buffer.rs:763`, also `:794`, `:832`, `:857`, `:883`):

```rust
let _lock = viewport_snapshot_test_lock();      // render lock
let _facts = SessionFactsEnvGuard::set("left"); // -> lock_test_env() at swarm_buffer.rs:717
```

Classic ABBA. Two tests running concurrently, one of each, block forever. Neither lock has a timeout, so the run never terminates. The harness's main thread parks in `recv` waiting for completions that never arrive, which is why there is no output and no failure, just silence.

The same shape exists in `crates/jcode-tui/src/tui/ui_tests/basic/idle_animation_partial.rs` (11 tests take `viewport_snapshot_test_lock()` after `IdleAnimationEnvGuard::enable()`, which takes `lock_test_env()` at `idle_animation_partial.rs:45`), so the ordering there happens to be env-first and is safe, but the guard/lock pairing is the same fragile pattern.

## Reproduction

Two tests, one from each side, with two threads:

```bash
BIN=$(ls target/selfdev/build/jcode-tui/*/out/jcode_tui-* | grep -v '\.d$' | head -1)
for i in $(seq 1 15); do
  "$BIN" --test-threads=2 --exact \
[...]

PR #1356, opened 2026-09-21
fix(test): bound the shared test-env lock wait and name its holder

### LINKED ISSUE #1349: cargo test can wedge forever on the shared test-env lock: unbounded wait, no diagnostics
## Expected and observed

`crates/jcode-base/src/storage.rs`:

```rust
pub fn lock_test_env() -> MutexGuard<'static, ()> {
    test_env_lock()
        .lock()
        .unwrap_or_else(|poisoned| poisoned.into_inner())
}
```

`std::sync::Mutex` is not reentrant, the lock is acquired from a large number of
call sites, and the wait has no bound. A single re-entrant acquisition on a thread
that already holds it therefore does not slow the suite down, it wedges it
permanently, and it takes the whole test binary with it.

Observed on `aarch64-apple-darwin`, 11 cores, running
`cargo test -p jcode-tui --lib` against a tree merged onto v0.86.0:

```
$ sample <pid of the wedged test binary> 3
    2637 Thread_44670464: tui::app::tests::answering_no_on_continue_prompt_shows_...
    2637 Thread_44670511: tui::app::tests::auto_poke_does_not_repeat_until_incomplete_...
    ... eleven test threads, one per core, each:
      11  jcode_base::storage::lock_test_env
          std::sync::poison::mutex::Mutex::lock
          __psynch_mutexwait
    main thread: run_tests_console -> recv CompletedTest -> park
```

Eleven test threads sat in `__psynch_mutexwait` with near-zero CPU; the run had
not terminated after fifty minutes and was killed. Nothing was waiting on the
render-state lock, so this is not the ABBA cycle fixed in #875 and not the Windows
lock-order case in #1201: the env lock alone was the cycle.

Two shapes fit the sample, and they look identical from outside, because a
self-deadlocked owner and an ordinary waiter have the same stack:

1. a thread that called `lock_test_env()` while it already held the lock, or
2. a guard whose thread went away before releasing it.

## Minimal reproduction

A nested acquisition never returns:

```rust
#[test]
fn nested_env_lock_hangs_forever() {
    let _outer = jcode_base::storage::lock_test_env();
    let _inner = jcode_base::storage::lock_test_env(); // never returns
}
```

## Proposed fix

Bound the wait and make the failure name its cause. The uncontended path stays a
single `try_lock`, and the poisoned-lock recovery is unchanged:

```rust
pub fn lock_test_env() -> MutexGuard<'static, ()> {
    let mutex = test_env_lock();
    match mutex.try_lock() {
        Ok(guard) => guard,
        Err(TryLockError::Poisoned(poisoned)) => poisoned.into_inner(),
        Err(TryLockError::WouldBlock) => wait_for_test_env_lock(mutex),
    }
}
```

`wait_for_test_env_lock` polls, warns once after 10s, and after a bound (300s by
default, `JCODE_TEST_ENV_LOCK_TIMEOUT_SECS` to shorten it for diagnosis) panics
with the waiting test's name and the last successful acquirer. Recording the
acquirer on the acquire path is enough to name the culprit: while a holder is
wedged nothing else can acquire, so the record still points at it.

The same record detects the re-entrant case directly in about 200ms instead of at
the full bound: if the recorded holder is this thread and the lock still cannot be
taken, this thread already holds it and the wait cannot end. The short grace
period covers the instructions between taking the lock and storing the record.

The alternative, letting nesting succeed, needs the guard type to change
(`MutexGuard<'static, ()>` becomes a wrapper or a reentrant-mutex guard) at around
forty annotated sites. That is the better end state if nesting is expected; the
bounded wait is the smaller change that removes the unbounded hang.

## Tests

Alongside the change:

- `shared_test_env_lock_timeout_defaults_and_parses_override`
- `a_contended_env_lock_is_acquired_once_the_holder_releases`
- `re_entering_the_env_lock_is_reported_instead_of_hanging` (fails in 0.21s instead of hanging)

ianalitis:
Closes #1349.

## Problem

`lock_test_env()` blocks forever with no bound and no diagnostics. `std::sync::Mutex` is not reentrant, the lock guards process-global test state that many helpers acquire, and the wait has no timeout, so a single re-entrant acquisition on a thread that already holds it wedges the whole test binary.

Observed on `aarch64-apple-darwin` (11 cores) running `cargo test -p jcode-tui --lib`:

```text
$ sample <pid of the wedged test binary> 3
    2637 Thread_44670464: tui::app::tests::answering_no_on_continue_prompt_shows_...
    2637 Thread_44670511: tui::app::tests::auto_poke_does_not_repeat_until_incomplete_...
    ... eleven test threads, one per core, each:
      11  jcode_base::storage::lock_test_env
          std::sync::poison::mutex::Mutex::lock
          __psynch_mutexwait
    main thread: run_tests_console -> recv CompletedTest -> park
```

Eleven test threads sat in `__psynch_mutexwait` with near-zero CPU; the run had not terminated after fifty minutes and was killed. Nothing was waiting on the render-state lock, so this is not the ABBA cycle from #875 or the Windows ordering case in #1201: the env lock alone was the cycle. Two shapes fit the sample and are indistinguishable from outside, because a self-deadlocked owner and an ordinary waiter have the same stack: a thread that already holds the lock calling it again, or a guard whose thread went away.

## Changes

- `lock_test_env` takes the lock with `try_lock` first, so the uncontended path and the poisoned-lock recovery are unchanged.
- A contended wait warns once after 10 seconds and fails after 300 seconds, overridable with `JCODE_TEST_ENV_LOCK_TIMEOUT_SECS`, naming the waiting test and the last successful acquirer. While a holder is wedged nothing else can acquire, so that record still points at the holder.
- A re-entrant acquisition is reported in about 200ms instead of at the bound: if the recorded holder is this thread and the lock still cannot be taken, this thread already holds it and the wait cannot end. The grace period covers the instructions between taking the lock and storing the record.

## Tests

- `storage::tests::shared_test_env_lock_timeout_defaults_and_parses_override`
- `storage::tests::a_contended_env_lock_is_acquired_once_the_holder_releases`
- `storage::tests::re_entering_the_env_lock_is_reported_instead_of_hanging` (0.21s; the same code hangs without the guard)

## Verification

```bash
cargo test -p jcode-base --lib -- storage::tests::        # 9 passed, 0 failed
cargo test -p jcode-base --lib                            # 11 failed, same set as pristine master
```

The full base suite does fail at the default thread count, but with the identical set of eleven pre-existing parallel auth/config races on this branch and on pristine `e589cbe5a` (A/B control run in a separate worktree: 1472 passed / 11 failed there, 1475 passed / 11 failed here, the difference being the three new tests). This change adds no failures.

## Tradeoff

This makes the failure bounded and loud; it does not make nesting legal. Letting nesting succeed needs the return type to change (`MutexGuard<'static, ()>` to a wrapper or a reentrant-mutex guard) at around forty annotated sites, which is a reasonable follow-up if re-entrant use turns out to be common rather than a defect at one call site.


ianalitis:
Both inline findings here are already addressed at the current head, `27231db08`, which is why the review anchored to the pre-fix lines reads as open:

- `TestEnvGuard::drop` now clears the holder record (`holder.held = false`), and `is_held_by_current_thread` requires a held record, so a thread that released the lock and later waits behind another holder is no longer misreported as re-entrant.
- The contended-wait test takes a readiness signal from the spawned waiter instead of sleeping, so the wait path is exercised deterministically.

Your own re-review at this head reports both as done.

PR #1357, opened 2026-09-21
fix(background): publish task status files atomically so bg wait cannot see a torn file

### LINKED ISSUE #1350: bg wait can report a running task as missing while its status file is rewritten
## Expected and observed

`BackgroundTaskManager::status` reads `<task-id>.status.json` and treats an
unreadable file as "task missing", so `wait` returns `None`. The completion path
rewrites that same file with `tokio::fs::write` / `std::fs::write`, which truncates
before writing. A reader that lands in that window sees an empty file and a
running task appears to vanish. Upstream at `e589cbe5a` still writes status files
with a plain truncating write.

Reproduction without a test harness:

```rust
// one manager rewrites the status file while another reads the same path
for _ in 0..400 { writer.write_status_file(&path, &status).await; }
for _ in 0..400 { assert!(reader.status("atomic-status").await.is_some()); }
```

Before the fix this failed 8 of 8 runs with
`a concurrent status write made the task look missing`.

As a real flake it showed up as `jcode-base --lib` failing 3 of 15 consecutive
full-suite runs on
`background::tests::adopted_output_is_readable_while_running_and_preserves_final_result`
(`adopted task is tracked`). A temporary diagnostic in `read_status_file` printed
exactly one failure per failing run:

```text
DIAG read_status_file parse len=0 path=.../035325k0o4.status.json
  err=EOF while parsing a value at line 1 column 0 head=""
```

Expected: a reader observes either the previous complete file or the new complete
file, and `wait` returns a task state rather than `None`.

## Proposed fix

Publish every status-file update through a uniquely named sibling plus rename, so
the rename is the only visible transition:

```rust
fn write_status_file_atomic(path: &Path, contents: &str) -> std::io::Result<()> {
    static WRITE_SEQUENCE: AtomicU64 = AtomicU64::new(0);
    let sequence = WRITE_SEQUENCE.fetch_add(1, Ordering::Relaxed);
    let temp = path.with_extension(format!("tmp-{}-{sequence}", std::process::id()));
    std::fs::write(&temp, contents)?;
    std::fs::rename(&temp, path)
}
```

Apply it to the spawn/adopt initial writes, the completion writes, the cancel
write, and the reconciliation rewrites. With it, the regression test
`background::tests::status_reads_never_observe_a_partially_written_status_file`
fails 8 of 8 runs against the truncating write and passes 10 of 10 after the
change; `jcode-base --lib` then passed 8 of 8 consecutive runs, and 15 of 15 in a
second series.

Notes for whoever implements it: a failed publish can leave one uniquely named
temp file behind, and both the reconcile sweep and `bg cleanup` key off the
`.json` extension, so such a file is inert. Rename-over-existing works on Windows
because Rust opens files with `FILE_SHARE_DELETE`.

ianalitis:
Closes #1350.

## Problem

`BackgroundTaskManager::status` reads `<task-id>.status.json` and treats an unreadable file as "task missing", so `wait` returns `None` while the task is still running. Tasks rewrite that same file on progress and completion with a plain truncating write, so a reader can observe an empty file in the window between truncate and write.

Reproduced at the library level: one manager rewrites the status file 400 times while another reads it 400 times and asserts the task is still tracked. Before this change it failed 8 of 8 runs.

As a suite flake it made `background::tests::adopted_output_is_readable_while_running_and_preserves_final_result` fail 3 of 15 full `jcode-base --lib` runs at the default thread count, and a temporary diagnostic in `read_status_file` printed exactly one failure per failing run:

```text
DIAG read_status_file parse len=0 path=.../035325k0o4.status.json
  err=EOF while parsing a value at line 1 column 0 head=""
```

## Change

Publish status updates through a uniquely named sibling plus a rename, so a reader sees either the previous complete file or the new complete one:

```rust
fn write_status_file_atomic(path: &std::path::Path, contents: &str) -> std::io::Result<()> {
    static WRITE_SEQUENCE: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
    let sequence = WRITE_SEQUENCE.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
    let temp = path.with_extension(format!("tmp-{}-{sequence}", std::process::id()));
    std::fs::write(&temp, contents)?;
    std::fs::rename(&temp, path)
}
```

The spawn, adopt, completion, cancel, and reconciliation writes all route through it, including the async `write_status_file` helper.

Edge cases: a failed publish can leave one uniquely named temp file behind, and both the reconcile sweep and `bg cleanup` key off the `.json` extension, so such a file is inert. Rename-over-existing works on Windows because Rust opens files with `FILE_SHARE_DELETE`.

## Test

`background::tests::status_reads_never_observe_a_partially_written_status_file` runs the concurrent read/write loop.

## Verification

```bash
cargo test -p jcode-base --lib -- background::tests::   # 21 passed, 0 failed
```

A/B control in the same worktree: reverting only `background.rs` to `origin/master` while keeping the new test makes it fail in 0.01s (`0 passed; 1 failed`), and restoring the helper makes `background::tests` pass 21 of 21.


ianalitis:
The finding was right for the commit it reviewed, and I reproduced it: with `13612fd2f`'s write path (no `await` in the loop body, so no suspension point) and the atomic rename replaced by an in-place write, this test passes 8/8 without ever reading during a write.

`dc4651417` is the commit that addresses it. Publication now runs on `spawn_blocking`, so each write suspends, and the same truncating-write control against the current head fails 8/8 on the default current-thread runtime with `a concurrent status write made the task look missing`. That was measured on a tree whose `write_status_file`, `write_status_file_locked`, `publish_status_file` and `write_status_file_atomic` are byte-identical to this branch's head, compared function by function.

The deterministic synchronization this asks for is in this PR already: `background/status_write_tests.rs::staged_atomic_publication_exposes_only_complete_old_then_new_json` installs a pre-rename hook, holds a publication after it has staged its temp file, reads the live path (old complete JSON) and the staged file (new complete JSON) while it is held, then resumes. It cannot pass vacuously either: with the same truncating-plus-inline control it fails closed in 3s with `status write did not reach the pre-rename hook: timed out waiting on channel`. It is one of six passing tests in that module.

I did not add synchronization to this loop test as well. A strengthened version exists on my integration line (multi-thread flavor, a `writes > 1` overlap assertion, and a fails-closed check for an empty file), and under the old inline shape it is not deterministic either: 2/5 runs failed, 3/5 passed. It would be a third overlapping guard without making the property deterministic. Happy to port it if you would rather have the belt and braces.

PR #1360, opened 2026-09-21
test(base): fix the 11 jcode-base lib tests that fail on master (#1358)

### LINKED ISSUE #1358: cargo test -p jcode-base --lib fails 11 tests at the default thread count (process-global test state races)
## Summary

`cargo test -p jcode-base --lib` fails 11 tests at the default thread count on `e589cbe5a` with no local changes. Every failing name passes when run by itself, and the failure text shows one test observing process-global state that another test set (including a developer's real account, when the machine has one). The run is 1472 passed / 11 failed / 5 ignored.

## Reproduction

```bash
git checkout e589cbe5a
cargo test -p jcode-base --lib            # 1472 passed, 11 failed, 5 ignored
cargo test -p jcode-base --lib -- --test-threads=1   # failing names pass
```

Environment: macOS `aarch64-apple-darwin`, 11 cores, `rustc 1.98.1`. It may depend on the machine already having provider auth state under the real `~/.jcode`; the failures below are the observation on a developer machine, not a claim that a clean CI runner reproduces all of them.

## Failing names

```
auth::codex::tests::multi_account_active_switch_works
auth::cursor::tests::vscdb_missing_key_returns_error
auth::oauth::tests::basic::save_claude_tokens_preserves_existing_account_metadata
auth::tests::cursor_status_is_available_for_authenticated_cli_session
auth::tests::full_and_fast_auth_status_document_cursor_cli_exception
config::tests::config_env_fingerprint_tracks_every_apply_env_override_var
config::tests::test_generated_default_config_has_expected_user_defaults
platform::platform_tests::spawn_detached_creates_new_session
prompt::prompt_tests::test_skill_prompt_integration
provider::tests::test_same_provider_account_candidates_include_other_openai_accounts
session::tests::cases::streaming_guard_creates_visible_macos_sleep_assertion
```

Sample failure, which is the shape for most of them:

```text
---- auth::codex::tests::multi_account_active_switch_works stdout ----
thread 'auth::codex::tests::multi_account_active_switch_works' panicked at crates/jcode-base/src/auth/codex_tests.rs:225:5:
  left: Some("openai-otter")
 right: Some("openai-1")
```

`openai-otter` is the account label from the developer's real `~/.jcode`, so that test read the live auth state instead of its own fixture.

## What looks like the cause

`jcode_base::storage::lock_test_env()` is the lane that serializes tests touching `JCODE_HOME`, the config cache, and auth overrides. Tests that take it only through `try_lock`-style helpers, or that set `JCODE_HOME` without it, proceed without exclusion while a sibling is mid-mutation. This is the same class as the two earlier lock-order reports (#875, #1201) and as the unguarded `JCODE_HOME` mutations in #593.

## Notes

- A downstream fork fixed the analogous cases by giving each suite one shared guard, taking the env lock once per test, routing every status-file write through a temp-plus-rename, and giving tests that need both locks a single order. Its copy of this suite passes at the default thread count. Happy to help port any of that if it is useful.
- Separately, the same suite's `background::tests::adopted_output_is_readable_while_running_and_preserves_final_result` races on a torn status file, which is reported with a fix in #1350.

ianalitis:
Fixes #1358.

## Root cause

The issue was filed as a thread-count race, but on re-check all 11 also fail with `--test-threads=1`. They are code-vs-test drift plus two environment leaks:

| Test | Cause |
|---|---|
| `auth::codex::tests::multi_account_active_switch_works` | c71a5986e switched to animal labels; fixture still expected `openai-1`/`openai-2` |
| `auth::oauth::tests::basic::save_claude_tokens_preserves_existing_account_metadata` | same (`claude-1` -> `claude-otter`) |
| `provider::tests::test_same_provider_account_candidates_include_other_openai_accounts` | same |
| `auth::cursor::tests::vscdb_missing_key_returns_error` | asserts the old error text; native rusqlite returns `QueryReturnedNoRows` |
| `auth::tests::cursor_status_is_available_for_authenticated_cli_session`, `full_and_fast_auth_status_document_cursor_cli_exception` | Cursor readiness is native-only now; tests still modelled the CLI path |
| `config::tests::config_env_fingerprint_tracks_every_apply_env_override_var` | `apply_env_overrides` reads `JCODE_MEMORY_JEV_PROVIDER` (2e6e42325) but `CONFIG_ENV_KEYS` never listed it. This is a real bug: config cache fingerprint ignores that var |
| `config::tests::test_generated_default_config_has_expected_user_defaults` | asserts a `reasoning effort "none"` comment the memory section no longer contains |
| `prompt::prompt_tests::test_skill_prompt_integration` | 24838926e reworded the identity line to `You are Jcode.` |
| `platform::platform_tests::spawn_detached_creates_new_session` | child wrote its sid via `ps -o sid=`, which is not portable; read `getsid(child)` from the parent instead |
| `session::tests::cases::streaming_guard_creates_visible_macos_sleep_assertion` | global `pmset` substring match fails whenever any other jcode process is streaming; count this pid's assertions vs a baseline |

Only `CONFIG_ENV_KEYS` is a non-test change (one line).

## Evidence (macOS aarch64, 11 cores)

```
origin/master 2a4edaa02:   1472 passed, 11 failed   (parallel and --test-threads=1)
this branch:               1482 passed,  0 of those 11 fail (3 parallel runs + 1 serial)
```

Unrelated intermittent failures seen on both base and branch: `background::tests::adopted_task_output_exists_while_running_and_is_replaced_on_completion`, `jev::tests::browser_auth_billing_and_redirect_failures_never_retry_or_fallback`, `jev::tests::failed_preflight_does_not_send_decision_request`, `provider::stream_timeout::tests::effort_scaling_never_shrinks_the_base_budget`. Not touched here.

`cargo clippy -p jcode-base --lib --tests`: no new warnings vs master. `cargo fmt --check` clean.


1jehuang:
Thanks for the thorough breakdown! Current master already has these fixed: `cargo test -p jcode-base --lib` passes 1587/0 at default threads, and `JCODE_MEMORY_JEV_PROVIDER` is in CONFIG_ENV_KEYS. Closing as superseded.

PR #1362, opened 2026-09-21
fix(swarm): make stop quiesce queued and active worker turns (#1352)

### LINKED ISSUE #1352: swarm stop acknowledges before queued and active owned work is cancelled
Related: #940 (detached worker commands replayed after stop). This is a different
gap in the same area, observed on `v0.84.229-dev (d027491f6)`.

## Expected and observed

A root-owned scout completed its read-only phase. The coordinator sent a follow-up
implementation message with `delivery=wake`, `await_members` still reported the
previous completed state, and `resume` / `wake` returned
`No swarm plan exists for this swarm` because those actions require a task graph.
`swarm stop` then returned `Stopped agent` for that exact scout id.

The stopped worker nevertheless executed `apply_patch` and selfdev test
invocations afterwards. Its own journal, not a guessed session label, records
those tool calls, and two test-request records identify the same session minutes
after the stop acknowledged. The worker then attempted a completion report that
could not reach the removed membership. Subsequent `stop` and message attempts
returned unknown session / target, and debug `cancel` was unavailable because
debug control is disabled.

Expected: `stop` cancels active and queued owned work and acknowledges only when
the worker cannot start further effects, or it reports cancellation pending.
Observed: membership became unaddressable while work continued.

## Where it looks like it goes wrong

`server/comm_session.rs::handle_comm_stop` removes the session entry and the
interrupt queue and then uses `agent_arc.try_lock()` to mark the session closed.
That path does not request cancellation before removal, so a busy agent can fail
the lock while its running task still owns an `Arc` and keeps executing. Treat
this as a candidate cause, not a proven root cause.

## Fix direction that worked locally

Keep ownership checks, then mark the member `stopping`, atomically close interrupt
delivery, wait for already-admitted queue producers, and repeatedly fire
`SessionControlHandle::request_cancel()` while waiting (bounded, five seconds) to
acquire and retain the terminal agent guard. Only while holding that guard perform
the final verified live/persisted clears, mark the agent closed, remove routing
and membership, complete the interrupt lifecycle, and emit `Done`. A timeout or a
failed final clear leaves membership, the agent, and the closed delivery gate
intact so a later stop can retry safely. Queue registration is the resume
boundary: it cannot reopen delivery while a stop is in flight, but it can reopen a
terminal stopped lifecycle.

Local regression coverage that failed before the change and passed after:

1. A live worker turn emits a text delta and blocks while live and persisted
   follow-up work is queued. Stop must end the turn, empty both queues, leave no
   active-turn signal, avoid a second provider call, and preserve an unrelated
   session/member.
2. `idle_live_agent` reserves an agent before a turn registers its cancellation
   signal; stop begins and the reserved wake is then handed to
   `spawn_tracked_live_turn`. The late wake must be discarded without a provider
   call, the agent closed, and `Done` emitted only after the guard is released.
3. Live and persisted queue producers enter before the stop gate is set and append
   after it. Stop must wait for those producers, clear again, and reject post-stop
   live, persisted, and direct-agent enqueues.
4. An uncooperative agent lock exceeds the quiescence timeout. The target stays
   registered as `stopping`; releasing the lock and retrying the same stop
   succeeds instead of returning `Unknown session` or replaying a cached timeout.

## Notes

- The guarantee this buys is local turn and tool-dispatch quiescence. It does not
  retroactively undo an external effect that began before cancellation reached a
  safe point.
- No session transcripts or journal paths are included here; they stay local.
- Happy to send a PR scoped to stop ownership plus the four regressions if the
  direction looks right.

ianalitis:
Fixes #1352.

## Problem

`swarm stop` acknowledged success while the stopped worker kept working. In the
reproduction recorded on #1352, a completed-then-woken scout executed
`apply_patch` and test invocations *after* `stop` returned `Stopped agent` for
its exact session id, and its subsequent completion report could not reach the
already-removed membership.

`handle_comm_stop` removed the session entry and interrupt queue and then tried
`agent_arc.try_lock()` to mark the session closed. That path never requested
cancellation, so a busy turn kept streaming, and a wake reserved before the stop
could still start provider work afterwards.

## Fix

Stop now, in this order:

1. marks the member `stopping` and closes the per-session interrupt-delivery
   gate, so already-admitted queue producers drain and later live or persisted
   enqueues are refused until an explicit resume;
2. acquires and retains the terminal `Agent` guard while repeatedly firing the
   session's registered cancel signals, bounded at 5s, so a turn that registers
   its cancellation signal *after* stop began cannot miss it;
3. performs verified live and persisted queue clears, `mark_closed`, interrupt
   queue teardown, membership removal and the terminal `Done` with that guard
   still held.

A stop that cannot quiesce leaves membership, Agent and the closed gate intact
and returns a retryable error instead of `Done`, so a later stop resumes rather
than reporting an unknown session. `stopping` members are also excluded from
headless startup recovery, and the gate follows session renames and is reopened
by queue registration for a new or resumed lifecycle.

This reuses existing primitives instead of adding a second cancellation
registry: `SessionControlHandle::request_cancel` (whose turn-cancel registry
already reaches server-initiated turns), the per-session soft-interrupt queues,
and `Agent::mark_closed`. The guarantee is local turn and tool-dispatch
quiescence; it does not retroactively undo an external effect that began before
cancellation reached a safe point.

## Tests

New `comm_session_stop_tests` module (4 tests) covering the concrete races:

- a live worker turn streaming while a live and a persisted follow-up are queued:
  stop must end the turn, empty both queues, leave no active-turn signal, avoid a
  second provider call, and preserve an unrelated session;
- `idle_live_agent` reserving the Agent before a turn registers its cancel
  signal: the late wake is discarded before provider dispatch;
- live and persisted queue producers that enter before stop and append after the
  gate closes: stop waits for them, performs a final verified clear, and rejects
  all post-stop enqueues;
- an uncooperative Agent lock beyond the 5s bound: the target stays registered as
  `stopping` and a later retry succeeds instead of returning `Unknown session`.

Evidence (`cargo test -p jcode-app-core --lib -- --test-threads=1`):

| Run | Result |
| --- | --- |
| `comm_session_stop_tests` with this change | 4 passed, 0 failed |
| The same 4 tests with the handler reverted to its pre-fix shape | 0 passed, 4 failed |

Each negative-control failure is the intended reason: the busy turn keeps
streaming after `Done`, the delivery gate is never closed, and a stop that cannot
quiesce still reports `Done` instead of a retryable state. `cargo fmt -p
jcode-app-core -- --check` is clean.

## Notes and limitations

- Stop error responses go through one helper; the pre-existing `Unknown session`
  and ownership-refusal messages are unchanged.
- Stopped-session delivery-gate tombstones are process-local until an explicit
  resume or a daemon restart.
- The E1 config-cache fixture fix from our own tree is not included: this
  repository has no E1 receipt tests.
- The `#[expect(clippy::too_many_arguments, ...)]` attribute above
  `resolve_swarm_spawn_effort` is moved back onto `spawn_swarm_agent`, which is
  the function that takes the 21 arguments. As written on `master` the
  expectation is unfulfilled, so `cargo clippy -p jcode-app-core --lib -- -D
  warnings` fails on that line before reaching anything this change touches. I
  can split that one-line cleanup out if you prefer it separate.
- `jcode-app-core` currently cannot pass `cargo clippy --all-targets
  --all-features -- -D warnings` on `master` for two unrelated reasons: nine
  never-constructed items in `tool/goal.rs` (the tool was disabled by removing
  only its registry line) and two unused imports in test modules. I left both
  alone; happy to send them as separate focused PRs.


ianalitis:
Addressed the review finding in `2a8c4e8dc`.

**What was wrong:** after a stop times out, the target stays resolvable with an open Agent (that is deliberate, so a retry can finish). Interrupt delivery is already closed by then, but `queue_interrupt:` / `queue_interrupt_urgent:` called `Agent::queue_soft_interrupt` directly, which only rejects a *closed* Agent. So the command reported `queued` while the session was stopping, leaving an interrupt pending across the failed-stop interval.

**Fix:** `DebugInterruptContext` now carries the session map, and both commands queue through the gate-aware `queue_soft_interrupt_for_session` path when a context is present, returning an error instead of `queued` while delivery is closed. The direct-Agent fallback is untouched for callers with no session context.

**Regression:** `stopped_but_unquiesced_worker_rejects_debug_queued_interrupts` holds the Agent lock so the stop times out, then asserts delivery is closed, the debug command errors rather than returning `queued`, and neither the live nor persisted queue holds anything. The debug call is time-bounded so the pre-fix behaviour fails on timeout instead of wedging the suite.

| Run | Result |
| --- | --- |
| `comm_session_stop_tests` with this change | 5 passed, 0 failed |
| The new test with the gate routing reverted | 1 failed in 7s, `a stopping session must not block a debug-queued interrupt on the busy Agent: Elapsed(())` |

Also checked the neighbouring suites: `server::comm_session` 37 passed / 4 failed, `server::debug_*` 30 passed / 1 failed, `server::queue_tests` 2 passed / 1 failed. Every one of those five failures also fails on a checkout without this change, so none is a regression. `cargo fmt -p jcode-app-core -- --check` is clean.

Issue #1133, opened 2026-09-01
Test isolation: the background git probe leaks the host repository's live state into unrelated TUI frame assertions


ianalitis:
## Summary

`gather_git_info` spawns a background thread that shells out to `git` and writes the result into a process-global cache. In a test binary this makes frame-level assertions depend on how many tests ran before them, and on whether the developer's checkout happens to be clean.

This is a test-isolation defect, not a user-visible bug, but it produces order-dependent failures that look like flakiness.

## Root cause

```rust
// crates/jcode-tui/src/tui/app/helpers.rs
static GIT_INFO_CACHE: Mutex<Option<GitInfoCacheEntry>> = Mutex::new(None);

pub(super) fn gather_git_info() -> Option<GitInfo> {
    ...
    *guard = Some((backdated_now(TTL + ...), None, true));
    std::thread::spawn(|| {
        let result = gather_git_info_inner();     // runs `git` in the cwd
        if let Ok(mut guard) = GIT_INFO_CACHE.lock() {
            *guard = Some((Instant::now(), result, false));
        }
    });
    None
}
```

The first caller in the process gets `None` and kicks off the probe. By the time a later test renders, the probe has landed, and the info widget shows the real branch, ahead/behind, and dirty counts of whatever repository `cargo test` was invoked from.

So a test's rendered frame differs depending on:

- whether any earlier test called `gather_git_info` (i.e. test ordering and how many ran),
- whether the probe thread had finished,
- and the live state of the developer's working tree.

## Impact

Frame assertions silently become order-dependent. In my checkout this surfaced as `test_background_task_markdown_is_suppressed_even_if_role_was_lost` failing only when preceded by a few hundred other tests, with the rendered frame containing a `git` info widget that a fresh process would not draw.

There is already a `seed_git_info_cache_for_tests` helper for exactly this problem, but it is opt-in and only one test (the onboarding golden) uses it, so every other rendering test inherits live repository state by default.

## Suggested fix

Do not run the live probe under `cfg(test)`. Read the cache and return whatever is in it, so the default is "no git data" and `seed_git_info_cache_for_tests` remains the explicit way to opt in. That keeps the existing seeded test working and makes every other test independent of the host repository.

## Notes

I have this patched on my fork if a reference is useful.

ianalitis:
A patch is ready on a fork branch, based directly on `master` (`325236119`), one commit:

https://github.com/ianalitis/jcode/tree/fix/test-git-probe-isolation

[...]

PR #1366, opened 2026-09-21
fix(tui): stop the live git probe leaking host repo state into tests

### LINKED ISSUE #1365: TUI test isolation regressed: gather_git_info probes the live repo again, so jcode-tui failures are order-dependent
## Expected and observed

#1133 was closed as completed on 2026-09-03: the background `gather_git_info`
probe no longer ran inside test binaries, so frame assertions stopped depending
on the host checkout. On `master` today the probe runs in tests again:

```console
$ git rev-parse master
2a4edaa02057ac994a601311c4f03ed450e1b3c9

$ cargo test -p jcode-tui --lib -- --test-threads=1
test result: FAILED. 2332 passed; 19 failed; 18 ignored
```

`crates/jcode-tui/src/tui/app/helpers.rs` still contains the `std::thread::spawn`
calls that shell out to `git` in `gather_git_info`, and the test-only read path
that #1133 introduced is gone. `seed_git_info_cache_for_tests` survives (the
onboarding golden test calls it), but it is no longer honored, so seeding does
not prevent the probe from overwriting the seeded value with the host's real
state.

## Why it matters

The failures are order-dependent rather than deterministic: which tests fail
depends on how many tests ran before them and whether the checkout running
`cargo test` is clean. Two of them are the onboarding "recent project" tests:

```
tui::app::tests::recent_project_review_falls_back_cleanly_when_no_repo_is_known
  assertion failed: app.status_notice is "No recent Git repository found"
tui::app::tests::preparing_recent_project_review_finishes_onboarding_and_seeds_the_first_turn
```

That also means the Linux `Build & Test` job's `Run TUI library tests` step
reports a different failure set per runner, which makes the rest of that job's
steps unreachable and the failure list hard to trust.

## Reproduction

```console
$ cargo test -p jcode-tui --lib -- --test-threads=1
test result: FAILED. 2332 passed; 19 failed; 18 ignored

$ cargo test -p jcode-tui --lib -- --test-threads=1 \
    --skip recent_project_review_falls_back_cleanly_when_no_repo_is_known
```

The second command changes the remaining failure set, which is the signal that
these are order-dependent rather than independent defects.

## Suggested fix

Restore the #1133 behavior in the `#[cfg(test)]` branch of `gather_git_info`:
read `GIT_INFO_CACHE` and never spawn the probe, leaving
`seed_git_info_cache_for_tests` as the only way to inject git data. A PR follows.

Related: #1133 (closed), #595, #596, #1342.

ianalitis:
Fixes #1365. Restores the behavior #1133 introduced and that `master` has since lost.

## Problem

`gather_git_info` spawns a background thread that shells out to `git` and writes
the answer into a process-global cache. In a test binary the first caller gets
`None` and later tests inherit whatever repository `cargo test` ran from, so
frame assertions depend on test order and on whether the developer's checkout is
clean.

On `master` (`2a4edaa02`):

```console
$ cargo test -p jcode-tui --lib -- --test-threads=1
test result: FAILED. 2332 passed; 19 failed; 18 ignored
```

Two of those are the onboarding "recent project" tests, which assert on a notice
that the leaked git state overwrites. `seed_git_info_cache_for_tests` still
exists (the onboarding golden test calls it) but is no longer honored, so seeding
does not stop the probe from replacing it.

## Fix

Restore the `#[cfg(test)]` read path in `gather_git_info`: tests read the cache
and never spawn the probe, and `seed_git_info_cache_for_tests` is again the only
way to inject git data. Production behavior is unchanged.

The second commit narrows
`test_background_task_markdown_is_suppressed_even_if_role_was_lost`, which
asserted the rendered frame contains no box-drawing character at all. Info-widget
overlays legitimately draw their own bordered box in the right margin, and whether
they are docked depends on how many frames the process has already rendered, so
that assertion was order-dependent rather than a statement about the card. The
card assertions themselves are untouched.

## Verification

`cargo test -p jcode-tui --lib -- --test-threads=1`:

| | Result |
| --- | --- |
| `master` | 2332 passed, 19 failed |
| this change | 2334 passed, 17 failed |

The remaining failures are tracked in #1340, #1342 and the other open `jcode-tui`
issues; this PR only removes the order dependence that made the set unstable.
`cargo fmt -p jcode-tui -- --check` is clean.


ianalitis:
The P1 points at `commands_accounts_02/part_01.rs` and `state_model_poke_02/part_01.rs`, which are outside this PR's scope.

I split that fix out so each change is independently reviewable: it is now https://github.com/1jehuang/jcode/pull/1368 for issue #1367, and this PR is back to the single isolation commit that its title and description describe.

Two notes from that work, since they affect how you read this PR:

- Your P2 there was right: with one account auto-activated, `/account switch <that label>` proved nothing, and the test passed with the switch removed. It now uses two accounts and fails without the switch.
- `cargo test -p jcode-tui --lib -- --test-threads=1` on this PR alone is 17 failed (from 19 on `master`); with #1368 applied it is 15.

PR #1368, opened 2026-09-21
fix(tui): stop two account tests asserting labels the store never assigns

### LINKED ISSUE #1367: Two jcode-tui account tests assert labels account_store::upsert_account never assigns
## Expected and observed

Two `jcode-tui` tests assert account labels that `account_store::upsert_account`
never assigns, so they fail on `master`:

```
tui::app::tests::test_account_switch_shorthand_switches_openai_account_by_label
    left: Some("openai-otter")
   right: Some("openai-1")

tui::app::tests::test_logout_clear_anthropic_accounts_removes_all_accounts_once
  called `Result::unwrap()` on an `Err` value: No account with label 'claude-3' found
```

Part of the set behind issue #1342.

## Root cause

`upsert_account` (`crates/jcode-base/src/auth/account_store.rs:125`) deliberately
ignores the requested label for a *new* account:

```rust
let requested_label = label_of(&account).to_string();
if let Some(existing) = accounts.iter_mut().find(|e| label_of(e) == requested_label) {
    *existing = account;
    return requested_label;          // update in place keeps the requested label
}
let label = next_account_label(prefix, accounts.len());   // otherwise canonical
```

`next_account_label` calls `canonical_account_label(prefix, count + 1)`, which
builds `prefix-<animal>` from a fixed list (`b28720562` replaced the numeric
scheme). So:

- The OpenAI test inserts one account labelled `openai2`, then asserts the active
  label is `openai-1`. The real label is `openai-<first animal>`, and it is
  derived from the account count, so the assertion can only pass by coincidence.
- The Anthropic test inserts three accounts labelled `requested-{1,2,3}` and then
  calls `set_active_account("claude-3")`, which cannot resolve: no account has
  that label.

## Reproduction

```console
$ cargo test -p jcode-tui --lib -- --test-threads=1 \
    test_account_switch_shorthand_switches_openai_account_by_label \
    test_logout_clear_anthropic_accounts_removes_all_accounts_once
```

Both fail on `master` (`2a4edaa02`) and on `rustc` 1.94 and 1.98 alike.

## Suggested fix

Capture the label `upsert_account` returns and use it, instead of asserting the
requested label or a hard-coded canonical one. That keeps these tests about
switching and clearing accounts rather than about the labelling scheme, which
has already changed once. A PR follows.

Related: #1342, and the copy-selection labels in #1340.

ianalitis:
Fixes #1367.

## Problem

`account_store::upsert_account` ignores the requested label for a **new**
account and assigns its own canonical one, so two `jcode-tui` tests assert a
label the store already replaced:

```
tui::app::tests::test_account_switch_shorthand_switches_openai_account_by_label
  left:  Some("openai-otter")
  right: Some("openai-1")

tui::app::tests::test_logout_clear_anthropic_accounts_removes_all_accounts_once
  No account with label 'claude-3' found
```

The store updates in place when a label already exists (keeping that label) and
otherwise calls `next_account_label(prefix, accounts.len())`, which yields
`prefix-<animal>` from a fixed list. `b28720562` replaced the numeric scheme and
did not update these tests. The OpenAI assertion therefore passes only by
coincidence, and the Anthropic call cannot resolve at all.

## Fix

Both tests capture the label `upsert_account` returns and use it. They still
assert the behaviour under test (switching the active account, and clearing all
accounts exactly once) instead of the labelling scheme, so they survive the next
renaming. No production code changes.

## Verification

```console
$ cargo test -p jcode-tui --lib -- --test-threads=1
```

| | Result |
| --- | --- |
| `master` (`2a4edaa02`) | 2332 passed, 19 failed |
| this change (isolated, without #1366) | 2336 passed, 15 failed |

Both named tests pass; no test that passed on `master` fails here. The remaining
failures are tracked in #1340, #1342 and #1365.


ianalitis:
Addressed the P2 in `9b4f9940e`.

**The finding was correct.** A single inserted account is activated automatically, so `/account switch <that same label>` left the active label unchanged and the test passed even with `submit_input` removed. I verified it: with the switch deleted, the old version still passed.

**Fix:** the test now inserts two accounts, asserts the first is active *before* the switch, and switches to the second.

| Run | Result |
| --- | --- |
| with the switch | passes |
| with the switch removed (negative control) | fails, `right: Some("openai-fox")` |

So the assertion is now discriminating: it detects a no-op `/account switch`. The Anthropic test is unchanged in intent (it asserts labels equal what `upsert_account` returned, then clears all three), and `cargo test -p jcode-tui --lib -- --test-threads=1` is unchanged at 2336 passed / 15 failed.

Issue #1144, opened 2026-09-01
Session::save silently skips the write for a session created with a title, so later lookups by id find nothing


ianalitis:
## Summary

`Session::create(parent, Some(title))` sets `title`, but the untouched-session write skip in `Session::save` only counts `custom_title`. A session whose only caller-chosen state is a `title` is therefore never written to disk, and anything that later loads it by id finds nothing and silently falls back as if the session did not exist.

## Cause

```rust
// crates/jcode-base/src/session/persistence.rs
if !self.persist_state.snapshot_exists
    && !self.messages.iter().any(super::is_visible_conversation_message)
    && !self.saved
    && self.custom_title.is_none()
{
    return Ok(());   // returns Ok, writes nothing
}
```

`save()` returns `Ok(())`, so callers have no signal that the write was skipped.

## Reproduction

```bash
cargo test -p jcode --lib -- --exact --test-threads=1 \
  cli::commands::menubar::tests::session_menu_item_title_loads_persisted_todo_title
```

Fails on `master` (`325236119`). The test creates a session with a title, calls `save()`, writes todos, and expects the menu label to use the todo group.

What actually happens: `format_session_menu_item_title` → `terminal_display_title_for_id` → `Session::load_startup_stub`, which fails with ENOENT because the file was never written. That resolver starts with `load_startup_stub(...).ok().and_then(...)`, so an unreadable session short-circuits the whole chain — including the `crate::todo::load_session_title` branch — and returns `None` before the todos are ever consulted.

Confirmed by probe: `load_todos` returns 1 item and `load_session_title` returns `Some("Meaningful menu labels")`, while `load_startup_stub` returns `Err(No such file or directory)`.

## Scope beyond the test

Any code path that creates a session with a title and later loads it by id is affected. Review/judge sessions are created with one (`"review"`, `"judge"`), so a parent-walk that resolves ancestors by id truncates at the first titled session.

## Fix

Treat `title` like `custom_title` in the skip condition: both are caller-chosen state, as opposed to the hidden session-context message the skip exists to avoid persisting.

Branch: https://github.com/ianalitis/jcode/tree/fix/persist-session-with-title

## Measured

| | before | after |
| --- | --- | --- |
| `cargo test -p jcode-base --lib -- session` | 9 failed | 3 failed |
| `cargo test -p jcode --lib -- menubar` | 1 failed | 0 failed |

[...]

PR #1354, opened 2026-09-21
fix(lint): clear clippy 1.98 and rustfmt drift, repair the dev-bins bench build

ianalitis:
Closes #1348. Closes #1294.

## Problem

On stable `rustc 1.98.1` / `clippy 0.1.98`, the quality job fails on `master` at `e589cbe5a` before any local change:

```bash
cargo clippy --all-targets --all-features -- -D warnings
cargo fmt --all --check
```

Two things also fail before clippy even runs:

- `cargo check --all-targets --all-features` fails: `src/bin/tui_bench.rs` (behind `dev-bins`) no longer builds. `diff_line_wrap` is not a `TuiState` member any more and `SidePanelSnapshot` gained a required `focus_revision` field.
- `cargo fmt --all --check` wants import reordering and rewrapping in four `jcode-tui` files that the lint fixes do not touch.

## What this changes

Portable lint drift, all upstream code, listed by lint:

| Lint | Sites |
| --- | --- |
| `collapsible_if` | `jcode-base/src/voice/streaming_capture.rs`, `jcode-tui-style/src/theme_mode.rs`, `jcode-harness-api/src/edit_stats.rs` (2), `jcode-tui/src/tui/app/terminal_title.rs`, `jcode-sdk/src/{auth,client,launch,ssh}.rs` |
| `needless_return` | `jcode-terminal-launch/src/lib.rs` (the site in #1294), `jcode-sdk/src/{client,launch}.rs`, `jcode-setup-hints/src/lib.rs` (3) |
| `manual_is_multiple_of` | `jcode-base/src/voice.rs`, `jcode-setup-hints/src/lib.rs` |
| `needless_borrow` | `jcode-base/src/memory.rs`, `jcode-app-core/src/tool/{apply_patch,patch}.rs` |
| `match_like_matches_macro` | `jcode-tui/src/tui/app/state_ui.rs` |
| `needless_range_loop` | `jcode-core/src/stdin_detect.rs` (macOS-gated) |
| `while_let_loop` | `jcode-sdk/src/client.rs`, `jcode-sdk/tests/client_behavior.rs` |
| `unnecessary_sort_by`, `double_ended_iterator_last` | `jcode-harness-api-server/src/translate.rs` |
| `unnecessary_get_then_check`, `redundant_closure` | `jcode-tui-mermaid` |
[...]

PR #1373, opened 2026-09-22
fix(session): persist debug, canary and improve-mode state on a blank session

### LINKED ISSUE #1339: Session::save() skips untouched sessions, so is_debug and canary never persist (4 failing tests)

ianalitis:
Fixes #1339.

`Session::save()` returns early when a session has no visible message, no title and no parent, so that a freshly opened panel does not turn its hidden session-context message into a transcript on disk. That guard already carries exemptions for caller-set explicit state — `custom_title`, `title` and `parent_id`, added by #1144 with the argument that a later lookup by id would otherwise find no file and silently treat the session as missing. `is_debug`, `is_canary` and `improve_mode` are explicit state by the same argument and were never added, so setting one and saving drops it silently: `save()` returns `Ok(())`, nothing is logged, and the load-by-id that follows fails with ENOENT.

This adds the three missing exemptions, following the structure and comments already there.

## `improve_mode` is required for the fourth test

The issue's suggested fix names `is_debug` and `canary`. That leaves `tui::app::tests::test_improve_mode_persists_in_session_file`, which the issue also lists, still failing: its session sets only `improve_mode`. I measured this rather than assuming it — with only the `is_debug`/`is_canary` clauses applied, the three e2e tests pass and that TUI test still fails with the identical ENOENT, which is why the third clause is here.

## Verification

Reproduced first at `origin/master` `2a4edaa02` on macOS with the guard reverted (the negative control is the same binary with one hunk removed, so the guard is the only variable):

```
$ cargo test --test e2e session_flow:: -- --test-threads=1
failures:
    session_flow::test_clear_preserves_debug_for_resumed_debug_session
    session_flow::test_debug_create_selfdev_session_marks_canary
    session_flow::test_debug_create_session_marks_debug
test result: FAILED. 3 passed; 3 failed
    each: Error: No such file or directory (os error 2)

$ cargo test -p jcode-tui --lib test_improve_mode_persists_in_session_file -- --test-threads=1
test result: FAILED. 0 passed; 1 failed
    panicked at .../commands_accounts_02/part_02.rs:9:65: load session: No such file or directory (os error 2)
```

With the change:

```
$ cargo test --test e2e session_flow:: -- --test-threads=1
test result: ok. 6 passed; 0 failed

$ cargo test -p jcode-tui --lib test_improve_mode_persists_in_session_file -- --test-threads=1
test result: ok. 1 passed; 0 failed
```

The issue reports this on Linux; the reproduction above is macOS, so the failure is not platform-specific.

## Wider suite

`cargo test -p jcode-base --lib session -- --test-threads=1` reports `126 passed; 3 failed`. All three fail identically with the fix reverted and are unrelated to this change:

| Test | Why |
| --- | --- |
| `auth::tests::cursor_status_is_available_for_authenticated_cli_session` | `left: NotConfigured, right: Available`; needs the `cursor` CLI configured on the machine |
| `platform::platform_tests::spawn_detached_creates_new_session` | fails at `platform_tests.rs:28` before touching sessions |
| `session::tests::cases::streaming_guard_creates_visible_macos_sleep_assertion` | asserts a macOS power assertion is released; reports 0 `BackgroundTask` assertions |

## Scope note

Three other fields are explicit state in the same sense and are still not exempt — `provider_key`, `reasoning_effort`, `subagent_model`, `autoreview_enabled`, `autojudge_enabled`, `working_dir` and `short_name`. I did not add them because no failing test names them, and enumerating every field is treating the symptom; if you would prefer the guard be expressed as "persist once any metadata differs from a fresh session's defaults" so it cannot drift again, I am happy to do that instead, but it is a larger change than the four tests this issue reports.


ianalitis:
For the record, `Quality Guardrails` is red on this head and the cause is not this change. The job log ends with:

```
error[E0407]: method `diff_line_wrap` is not a member of trait `TuiState`
error[E0063]: missing field `focus_revision` in initializer of `SidePanelSnapshot`
error: could not compile `jcode` (bin "tui_bench") due to 2 previous errors
```

That is the `src/bin/tui_bench.rs` build break, which reproduces on unmodified `master` under `cargo check --all-targets --all-features` (GitHub's `action_required` means this PR's upstream run needs maintainer approval, so the failure above is from a fork run of the same commit). It is fixed in #1354, which is still open, so this head will stay red until that lands. Nothing in this PR touches `tui_bench` or `TuiState`; the change is the persistence guard's exemption list.

1jehuang:
Thanks! Master now covers this: the empty-session save guard exempts is_debug/is_canary (and a force path), and test_improve_mode_persists_in_session_file passes on current master. Closing as superseded.

PR #1378, opened 2026-09-22
fix(tui): keep the draft when Ctrl+Up jumps into prompt history

### LINKED ISSUE #1361: Ctrl+Up destroys the current unsubmitted prompt
## Summary

In the TUI, pressing Ctrl+Up (or Alt+Up / Cmd+Up) while the composer contains an
unsubmitted draft overwrites that draft with a prompt from history. The draft is
not stashed, not pushed onto the input undo stack, and not reachable by walking
back down the history. It is simply gone.

This is data loss of user-authored text triggered by a single keystroke on a
navigation binding, and it is recoverable in neither direction.

Version observed: v0.84.23-dev (90c94adb0), macOS, tmux.

## Reproduction

1. Have at least one prompt in history (any previously submitted prompt).
2. Type a multi-line draft into the composer. Do not submit it.
3. Press Ctrl+Up.

Observed: the composer content is replaced by the most recent history entry.
The draft is unrecoverable.

- Ctrl+Z ("undo input edit") does not restore it — no undo snapshot was taken.
- Ctrl+Down walks forward through history and then clears to empty. It never
  returns to the draft, because the draft was never stored as the "position
  after the newest history entry".
- Ctrl+S ("stash or restore the input draft") cannot help retroactively.

Expected: browsing history should be non-destructive. The draft should come back
when the user walks back past the newest entry, and/or Ctrl+Z should restore it.

### Where the draft is dropped

`crates/jcode-tui/src/tui/app/input.rs`, `handle_prompt_history_navigation`.

With a non-empty composer, the function looks up the current input in history.
An in-progress draft is by definition not in history, so `rposition` returns
`None` and control reaches the `explicit_history` branch:

```rust
let Some(current_index) = history.iter().rposition(|prompt| prompt == &app.input) else {
    if explicit_history && matches!(code, KeyCode::Up) {
        return history
            .last()
            .map(|prompt| {
                app.input = prompt.clone();          // draft overwritten here
                app.cursor_pos = app.input.len();
                app.reset_tab_completion();
                app.sync_model_picker_preview_from_input();
            })
            .is_some();
    }
    return false;
};
```

`app.input` is assigned directly. There is no `remember_input_undo_state()` call
before the assignment and no equivalent of the draft-save that the Ctrl+R
overlay performs.

### Why the plain-Up path is safe and the Ctrl+Up path is not

The same function handles bare Up. For bare Up, `explicit_history` is false, so
the unmatched-draft case falls through to `return false` and the draft is left
alone. This is pinned by an existing test,
`test_prompt_history_up_does_not_replace_unmatched_draft`.

Ctrl+Up deliberately opts out of that protection. The dispatcher in
`handle_key_core` normalizes the modifier and comments the intent:

```rust
// Normalize to CONTROL so handle_prompt_history_navigation takes its
// explicit-history path (jump straight into history even mid-draft).
handle_prompt_history_navigation(self, KeyCode::Up, KeyModifiers::CONTROL);
```

Jumping into history mid-draft is a reasonable feature. The defect is that the
jump is implemented as an unrecorded overwrite rather than as a save-then-replace.

### The codebase already has the right pattern, in the same module

`prompt_history.rs` implements the Ctrl+R reverse-search overlay, and it handles
exactly this concern correctly. `PromptHistorySearchState` carries:

```rust
/// Input-line draft captured when the search opened, restored on cancel
/// (Esc) or when the query stops matching anything.
original_input: String,
original_cursor: usize,
```

`open_prompt_history_search` captures the draft, `apply_prompt_history_search_preview`
falls back to it when nothing matches, and `cancel_prompt_history_search` restores
it. So within one module, Ctrl+R treats the draft as precious and Ctrl+Up treats
it as disposable. That inconsistency is the core of this report.

The general-purpose mechanism also already exists and is used by roughly twenty
other input mutations: `remember_input_undo_state()` in
`state_ui_input_helpers.rs`, backing Ctrl+Z via `undo_input_change()`. Word
deletion, line cut, paste, autocomplete and drag-and-drop insertion all call it.
History recall is a conspicuous omission.

### Scope

Both the local and the remote/SSH TUI paths are affected; `key_handling.rs`
calls the same function for Ctrl+Up and Ctrl+Down. The binding is also reached
via Alt+Up and Cmd+Up, since `is_prompt_recall_modifier` accepts Ctrl, Alt and
Super so the gesture works across terminals. That widens the accidental-trigger
surface: Alt+Up and Cmd+Up are ordinary cursor/scroll motions in many editors
and terminals, so a user can hit this without ever intending to open history.

The forward direction has a matching hole. `KeyCode::Down` at the newest entry
does `app.input.clear()`, so even a user who understands the feature cannot walk
back to the draft — there is no slot where it would live.

## Suggested fix

Minimal change, one line, in the `explicit_history` branch:

```rust
app.remember_input_undo_state();
app.input = prompt.clone();
```

I verified this locally against the current tree:

- A repro test asserting draft loss passes on unmodified code, then fails once
  the line is added (bug gone).
- A test asserting Ctrl+Up followed by Ctrl+Z restores the exact draft and
  cursor position passes with the fix.
- The 10 existing `prompt_history` tests and the existing `ctrl_up` tests,
  including `test_ctrl_up_down_always_browses_prompt_history` and
  `test_prompt_history_up_does_not_replace_unmatched_draft`, still pass.

This makes the loss recoverable via the undo affordance users already know, and
does not change the "jump straight into history mid-draft" behavior the comment
says is intended.

A more complete fix would additionally treat the draft as the slot past the
newest history entry, mirroring shell and readline behavior: stash it on entry
into history browsing and restore it when Ctrl+Down walks back off the end,
instead of clearing to empty. That matches what Ctrl+R already does with
`original_input` and would make the whole gesture non-destructive rather than
merely undoable.

## Note on related issues

Issue #264 ("Input history recall") is the closed feature request that
introduced this functionality. Its acceptance criteria covered recall and
forward navigation but did not state that an unsubmitted draft must survive
entering history. This report is a defect follow-up on that gap, not a duplicate.

costajohnt:
Ctrl+Up (and Alt/Cmd+Up) from an unsubmitted draft replaced it with the newest history entry, with no undo snapshot and no way back through history.

This does both things the issue suggests: it takes an undo snapshot before the jump, so Ctrl+Z restores the draft and cursor, and it keeps the draft as the slot past the newest entry, so Down walking back off the end returns it instead of clearing. That matches how the Ctrl+R overlay treats its original input. The saved draft is dropped wherever the undo history is cleared, which includes both submit paths.

Tests: `test_ctrl_up_from_draft_restores_it_walking_back_down` and `test_ctrl_up_from_draft_can_be_undone`. Both fail on master and pass with the change. The existing Ctrl+Up and prompt-history tests still pass. On a full `jcode-tui` run in a `rust:1` container, the only failures are the same environment-dependent ones (clipboard, git, accounts) that fail on master in that container.

Fixes #1361


ianalitis:
Greptile's finding reproduces. I confirmed it independently on `2cf5356` (on top of `ef4c2bd69`) and found a one-line fix that keeps your two tests green.

## Reproduction

```rust
app.display_messages = vec![
    DisplayMessage::user("older prompt"),
    DisplayMessage::assistant("older response"),
    DisplayMessage::user("newer prompt"),
];
app.input = "v1".to_string();
app.cursor_pos = app.input.len();

// Ctrl+Up stashes the draft; Ctrl+Z brings it back.
app.handle_key(KeyCode::Up, KeyModifiers::CONTROL).unwrap();
app.handle_key(KeyCode::Char('z'), KeyModifiers::CONTROL).unwrap();
assert_eq!(app.input, "v1");

// The user keeps editing, then accepts a match through Ctrl+R.
app.input = "v2".to_string();
app.cursor_pos = app.input.len();
app.handle_key(KeyCode::Char('r'), KeyModifiers::CONTROL).unwrap();
for c in "newer".chars() {
    app.handle_key(KeyCode::Char(c), KeyModifiers::empty()).unwrap();
}
app.handle_key(KeyCode::Enter, KeyModifiers::empty()).unwrap();
assert_eq!(app.input, "newer prompt");

// Fails at this head: the composer becomes "v1" and "v2" is gone.
app.handle_key(KeyCode::Down, KeyModifiers::CONTROL).unwrap();
assert_ne!(app.input, "v1");
```

At this head the last assertion fails with `left: "v1"`. Greptile's cause is right: `undo_input_change` restores the composer but leaves `history_draft` holding the pre-edit value, so a later walk off the newest entry re-installs it over the user's current text.

## Fix

`crates/jcode-tui/src/tui/app/state_ui_input_helpers.rs`, in `undo_input_change`:

```rust
if let Some((input, cursor_pos)) = self.input_undo_stack.pop() {
    // The composer now holds a restored draft, so the copy stashed by a
    // history jump is stale: a later Down must not resurrect it.
    self.history_draft = None;
```

I deliberately did **not** add the same clear to the Ctrl+R accept path, which is what Greptile's "or refresh" alternative would suggest: jumping into history from a draft and then accepting a match through Ctrl+R is a case where the draft is still current, so `Ctrl+Down` restoring it is the behaviour this PR wants. Undo is the one route that replaces the composer with an older draft while leaving the user's edits unrepresented in `history_draft`.

## Verification

With the line above, all of these pass on your branch:

```sh
cargo test -p jcode-tui --lib -- --test-threads=1 ctrl_up          # 5 passed
cargo test -p jcode-tui --lib -- --test-threads=1 probe_stale      # the repro above, 1 passed
cargo test -p jcode-tui --lib -- --test-threads=1 prompt_history   # 10 passed
```

Adding the repro from above as a third test in `part_02.rs` would pin it.

One note for the record: I had opened #1379 for the same issue from the undo-snapshot half, about twelve minutes after you opened this one, and my search for existing work did not see this PR because it had not been created yet. This is the fuller fix of the two, so #1379 is closed in its favour. Yours is the one to merge.

costajohnt:
Reproduced it: at 2cf5356, Ctrl+Down brings back "v1" instead of keeping "newer prompt". Pushed your fix in cbc8ca1 (clear `history_draft` in `undo_input_change`) along with your repro as `test_undo_after_ctrl_up_drops_stale_history_draft`. The new test and the existing ctrl_up and prompt_history tests all pass (14).

PR #1379, opened 2026-09-22
fix(tui): keep an unsubmitted draft when Ctrl+Up recalls history

### LINKED ISSUE #1361: Ctrl+Up destroys the current unsubmitted prompt

ianalitis:
Fixes #1361.

Ctrl+Up (and Alt+Up and Cmd+Up, which share the binding) jumps straight into prompt history even mid-draft. When the composer held text that is not in history, that path assigned `app.input` without snapshotting it, so the draft was gone in one keystroke: no undo entry, and Ctrl+Down at the newest entry clears to empty instead of returning to it.

The fix is the mechanism the module already uses for every other input mutation: call `remember_input_undo_state()` before the assignment, so Ctrl+Z restores the exact text and cursor position. Entering history from a draft is deliberate (the dispatcher normalizes to `CONTROL` for that purpose) and is unchanged. The plain-Up path already refuses to replace an unmatched draft and is untouched.

`#1361` reported and diagnosed this, including the one-line shape of the fix; the credit is theirs.

## Verification

- Two regressions in `crates/jcode-tui/src/tui/app/tests/scroll_copy_01/part_02.rs`: the draft comes back via Ctrl+Z with its cursor position, and it still comes back after walking several entries into history.
- A/B on the full `jcode-tui` lib suite, single-threaded, pristine versus patched: **19 failures versus 17**, and the only difference between the two failure sets is these two tests. No pre-existing failure changes state.
- `rustfmt --edition 2024 --check` on both touched files reports no diff inside the changed hunks. The two pre-existing `part_02.rs` diffs it does report are at lines 450 and 484 and are the master drift #1354 fixes.

## Red checks on this branch

`Quality Guardrails` and `Format` are red on any branch based on current `master`, for reasons that are not this change: `src/bin/tui_bench.rs` does not compile (`diff_line_wrap` is no longer a member of `TuiState`, `focus_revision` missing from a `SidePanelSnapshot` initializer) and `cargo fmt --all --check` reports drift in files this branch does not touch. Both are covered by #1354.


ianalitis:
Closing this as a duplicate of #1378 by @costajohnt, which fixes the same issue and came first.

#1378 was opened at 13:37Z and this PR at 13:49Z; my check for existing work on #1361 ran at 13:36Z, one minute before #1378 existed, so it found nothing. That is the whole story: no disagreement about the fix, just a race.

#1378 is also the more complete change. This PR only takes the undo snapshot the report's "and/or" allowed; #1378 does that *and* keeps the draft as the slot past the newest entry, so `Ctrl+Down` walking off the end hands the draft back instead of clearing to empty, matching how the `Ctrl+R` overlay already treats its original input.

I verified #1378 locally on top of `ef4c2bd69` and reviewed the Greptile finding on it (a stale draft can replace the user's later edits after an undo). The reproduction and a one-line fix are in [this comment](#issuecomment-5778084908), and with that fix the two tests here also pass, so nothing from this PR is lost.

The branch is being deleted. Credit for the fix is @costajohnt's, and the diagnosis was @theammir's in #1361.

Issue #1134, opened 2026-09-01
Claude OAuth broken by Anthropic version gate: claude-cli/2.1.123 rejected (claude_code_version_too_old)


charlespers:
Every Claude OAuth turn on `claude-fable-5-1` now fails:

```
Anthropic API error (400 Bad Request): Claude Code 2.1.123 does not support this model;
version 2.1.251 or newer is required. (error_code: claude_code_version_too_old)
```

Anthropic added a server-side gate on the Claude Code version in the User-Agent. jcode presents as `claude-cli/2.1.123` in three places:

- `CLAUDE_CLI_USER_AGENT` in `crates/jcode-base/src/provider/anthropic.rs:39` (the string the gate reads)
- `OAUTH_BILLING_HEADER` `cc_version` in `crates/jcode-provider-anthropic/src/lib.rs:10`
- preflight eval `app_version` in `crates/jcode-provider-anthropic-runtime/src/lib.rs:416`

**Verified against the live API**: same OAuth token, identical request body — UA `claude-cli/2.1.123` → 400 version gate, `claude-cli/2.1.257` → 200. Also verified end to end: binary-patching the string in v0.81.4 makes `jcode run` complete normally again.

Fix branch (bumps all three to 2.1.257, current Claude Code release, `cargo check` clean): https://github.com/charlespers/jcode/tree/fix/claude-code-version-gate — I could not open a PR (`CreatePullRequest` permission refused for forks, both via gh and the API), so please pull from that branch or cherry-pick 	he single commit.

Suggestion: keep the pseudo-version in one shared const, or read it from the installed `claude` binary, so the three spots cannot drift and the next gate bump is a one-line fix.

1jehuang:
Thanks for the precise report and live-API verification. I updated all three OAuth identity sites to Claude Code 2.1.257 in #1135. The affected crates pass `cargo check`, and the source now has no remaining 2.1.123 identity.

--- *— Jcode agent (automated triage), on behalf of @1jehuang*

PR #1388, opened 2026-09-22
fix(anthropic): bump Claude CLI user agent to 2.1.280

arxdeus:
required to support `Opus 5.5` models


arxdeus:
tested with local `jcode` build: `opus 5.5` works fine just with this change

Kenmege:
Independent verification of this PR, built from source rather than byte-patching the shipped binary:

- Checked out `b20c02a` and ran `cargo build --release --bin jcode` (clean build).
- `jcode run -p claude -m claude-opus-5-5 "Reply with exactly: OK"` with Claude OAuth: **OK**. The shipped v0.86.0 fails the same command with `400 claude_code_version_too_old` (requires 2.1.280).
- Regression check on the same build: `-m claude-opus-5` also returns **OK**.
- `grep -rn '2\.1\.257'` on master finds exactly three sites (`CLAUDE_CLI_USER_AGENT`, `OAUTH_BILLING_HEADER`, preflight `app_version`), and this diff updates all three, so nothing is left behind.

This closes #1390 as far as I can tell.

Possible follow-up, not a blocker: this is the third time the same gate has hit (2.1.123 in #1134, 2.1.257 now). The Cursor, Gemini and Antigravity providers already accept a runtime env override for the client version they advertise (`JCODE_CURSOR_CLI_VERSION` and friends). The same pattern here would let users unblock themselves the next time Anthropic raises the minimum, without waiting for a release.

Issue #1393, opened 2026-09-22
Palette re-derives color roles from RGB, so overriding one role recolors unrelated roles and backgrounds


zipadoodlez:
**Priority:** high (correctness: `/colors <role> <hex>` changes roles the user did not name, including backgrounds)
**Area:** `crates/jcode-tui-style/src/palette.rs`, `crates/jcode-tui-style/src/theme_mode.rs`, `crates/jcode-tui-style/src/theme.rs`
**Version:** jcode v0.86.0 (`e589cbe5a`)
**Root cause of:** #1391 (reasoning cannot be recolored without touching every `dim` element)

## TL;DR (plain language)

jcode has no idea which role a pixel came from. It decides by **nearest gray**.

Every semantic color is emitted as a raw RGB value, and later a per-frame pass
guesses which role that RGB "belongs to" by finding the closest role default
within a fixed distance. So `/colors dim red` does not mean "recolor dim". It
means "recolor every color within ~0.16 Oklab of dim's default", which today
includes `tool`, `border`, the raw markdown-dim literal, and even the `user_bg`
and `selection_bg` **backgrounds**.

## Evidence (measured, not inferred)

Running the real functions (`Palette::set` + `remap_literal_with`) on the
default palette, then overriding exactly one role:

```
only dim overridden to red (255,0,0):
    tool         (120,120,120) -> (253,146,130)
    border       (100,100,110) -> (253,105,97)
    user_bg      ( 35, 40, 50) -> (174,  0, 24)   <- background
    selection_bg ( 60, 60, 80) -> (213, 29, 55)   <- background
    raw md_dim   (100,100,100) -> (252,104, 87)

only tool overridden to green (0,255,0):
    dim      (80,80,80) -> (43,199,39)
    pending  (140,140,140) -> (187,254,181)
    border   (100,100,110) -> (50,227,72)

only border overridden to blue (0,0,255):
    tool         (120,120,120) -> (5,81,246)
    dim          (80,80,80) -> (0,46,167)
    pending      (140,140,140) -> (49,112,248)
    selection_bg (60,60,80) -> (4,12,147)
```

Note `user_bg` and `selection_bg` are backgrounds and are graded by separate
harmony rules; a foreground override still drags them.

## Root cause

Role identity is modeled, then **discarded**, then **re-inferred**.

1. The neutral render model tags every span with a semantic role
   (`crates/jcode-render-core/src/model.rs`, `StyleRole`).
[...]

PR #1394, opened 2026-09-22
fix(colors): exact role defaults are owned by their role only

zipadoodlez:
Part of #1393.

## Problem

A `/colors <role> <hex>` override recolored roles the user did not name.

`palette::remap_literal_using` matched the rendered color against only the **overridden** roles, so a single override claimed every color within `FAMILY_RADIUS = 0.16` of it. The roles that actually owned those colors were never candidates, so they could not win. Measured on the default palette:

```
only dim overridden to red (255,0,0):
    tool         (120,120,120) -> (253,146,130)
    border       (100,100,110) -> (253,105,97)
    user_bg      ( 35, 40, 50) -> (174,  0, 24)   <- background
    selection_bg ( 60, 60, 80) -> (213, 29, 55)   <- background

only tool overridden to green (0,255,0):
    dim      (80,80,80) -> (43,199,39)
    pending  (140,140,140) -> (187,254,181)
    border   (100,100,110) -> (50,227,72)
```

## Fix

Test exact-default ownership first: a rendered color that equals a role's default **is** that role's own output, so only that role's override may recolor it (and if the role is not overridden, it is left untouched).

Non-default literal variants keep the historical nearest-overridden-role match, so a "dimmer amber follows Warning" still works.

One function (`palette::remap_literal_using`), no API change.

## Test

`cargo test -p jcode-tui-style` -> 86 passed, including the new
`palette::tests::an_override_does_not_recolor_other_roles_output`.

## Follow-up (not in this PR)

Raw literals that are not any role's default (e.g. `md_dim_color()` = `(100,100,100)`) still bleed among themselves and still need adoption into roles. That is the next step of #1393 and is deliberately out of scope here.


zipadoodlez:
Superseded by #1399, which implements the #1397 end state and includes this fix: exact role defaults are attributed to their role only, so an override cannot recolor another role’s output or the backgrounds. The only behavior this PR kept and #1399 removes is proximity-following for untagged non-default literals (e.g. a “dimmer amber” tracking /colors warning). That narrowing is the explicit trade in #1397 and needs a maintainer ruling there. If the ruling is to keep that behavior, this branch (fix/color-role-attribution-bleed) is a working fallback and can be reopened.

